Issue #18, opened 2019-05-28
Missing ability to include another YAML file in a short way (ie combine load + func call)


mildred:
In ytt, there is the possibility to load values from files, but it does not seem that it is possible to load another YAML file from the main ytt file.

Could this be considered as an addition?

Something like this:

main.yaml
```yaml
---
test: #@ include("hello.yaml")
```

hello.yaml
```yaml
---
#@ def world()
"world"
#@ end

hello: #@ world()
```

the result would be:

```yaml
---
test:
  hello: "world"
```

Having a way to pass variables to the included file would be nice. What this is is basically a `#@ load()` but with the function definition implicit in the included file.

nimakaviani:
hi @mildred ,

ytt in fact allows for loading files with function definitions in other files.

Here is an example from the ytt playground with a very similar usecase to what you explained above: https://get-ytt.io/#example:example-load

mildred:
Yes, I understand, but I wanted a way to avoid having to explicitly declare the functions. I would like to have instead of `funcs.lib.yml`

**`func.lib.yml`:**
```yaml
name: max
cities:
- SF
- LA
```

*without* the `#@ def func1():`. The whole file would be by itself a single function.

Perhaps that can be made into the `load()` builtin. So when there is no `#@ def` in the loaded file, an implicit `_main` function could be defined.

The main template could then be:

**`main.yml`**
```yaml
#@ load("func.lib.yml", func1="_main")

func1_key: #@ func1()
```

which would result in:

```yaml
func1_key:
  name: max
  cities:
  - SF
  - LA
```

It's just that I'm used to have include statements, and having to define functions seems a bit odd when you mean to have the whole file as a single function.

cppforlife:
i was actually considering adding this functionality at some point, but went with a more conservative approach for now.

this was due to awkwardness with parametarization. ie how does one specify which args whole file takes. i guess it might be worth adding this feature without allowing args, but then "upgrade" from no-args to args is a bit weird. wdyt?

mildred:
You're right, what I'm asking is not really necessary, and is more convenience. in any case it's not very important.

Upgrade can happen simply by adding a `#@ def` statement enclosing the whole file that map to the default function name.

mildred:
feel free to close this issue or file it for later if you think it's relevant.

mildred:
What can be useful though, is having a way to perform both the load() and the function call.

A way to shorten the following file:

```yaml
[...]

PR #24, opened 2019-05-31

mildred:
@ytt:lib can be loaded from within starlark and it provides a `lib`
modules with a `load()` function. it takes:

  - A string with the location of the module to load
  - An optional `recursive` keyword argument, defaults to true

It returns a loadable library which accepts the `eval()` method. It
takes a list of keyword arguments that override the loaded library
default values

After evaluation, the returned object has an `output()` method to fetch
the result from the library evaluation. It takes an optional string
argument for the file name to get the results from. If empty, it will
return the concatenated results for all files.

This is quite equivalent as running ytt from within the template and
embedding the output in it.

Implements #18


mildred:
Before really merging, I need to continue testing it and add examples in the playground to explain how it works.

nimakaviani:
yes indeed! I was going to suggest adding some test files here: https://github.com/mildred/ytt/tree/ytt-lib/pkg/yamltemplate/filetests

cppforlife:
few comments at high level:

> A string with the location of the module to load

i saw in this [screenshot](https://github.com/k14s/ytt/issues/18#issuecomment-497922517) that loading path included `_ytt_lib`. this directory is meant to provide a "private" boundary for library dependencies. it shouldnt be something user types in similar to `load("@some-lib:...")` (https://get-ytt.io/#example:example-load-custom-library). for example, with this fs:

```
app1/
app1/app.yml
app1/_ytt_lib/github.com/k14s/k8s-lib/app
app1/_ytt_lib/github.com/k14s/k8s-lib/app/_ytt_lib/private
```

one should be able to use `lib.load("github.com/k14s/k8s-lib/app")` from app1/app.yml (note no _ytt_lib); however, this `lib.load("github.com/k14s/k8s-lib/app/_ytt_lib/private")` from app1/app.yml should not be possible, as private is private to `github.com/k14s/k8s-lib/app` library. i believe current Library object encapsulates this behavior already.

briefly looking at the code, i believe we should be able to pass around Library objects (instead of having to call NewFiles(...)) multiple times => should be able to operate on the in-memory Library tree, and pass around portions of it. Which also leads me to ask, do we actually need distinction between internal and external libraries?

> TODO find a way to make it work on the Web UI where there is no real filesystem

working with Library objects (in memory) will solve this problem as well.

> An optional recursive keyword argument, defaults to true

i dont think we actually need this functionality with above changes.

> After evaluation, the returned object has an output() method to fetch
the result from the library evaluation. It takes an optional string
argument for the file name to get the results from. If empty, it will
return the concatenated results for all files.

i like output vs text distinction. we should also probably consider adding ability to list and get individual files "after" evaluation. something like this:

```yaml
#@ evaled_lib = lib.load("github.com/k14s/k8s-simple-app-example/config-step-2-template").eval({...})
#! files is a sibling to other methods (currently, output and text)
#@ for/end file_name in evaled_lib.files():
- #@ files.read(file_name)
```

mildred:
Good to know about `_ytt_lib`, I didn't understand it as well.

> briefly looking at the code, i believe we should be able to pass around Library objects (instead of having to call NewFiles(...)) multiple times => should be able to operate on the in-memory Library tree, and pass around portions of it. Which also leads me to ask, do we actually need distinction between internal and external libraries?

There is one important use case that requires using `NewFiles()`, it's when you want to load templates from the parent directory. In the first version of this PR, I created a new root library all the time using `NewFiles`. In this version, it's only used when the `Library` object can't be found.

> briefly looking at the code, i believe we should be able to pass around Library objects (instead of having to call NewFiles(...)) multiple times => should be able to operate on the in-memory Library tree, and pass around portions of it. Which also leads me to ask, do we actually need distinction between internal and external libraries?

> > TODO find a way to make it work on the Web UI where there is no real filesystem
> working with Library objects (in memory) will solve this problem as well.

This is what I did in a second version (in this PR) and I forgot to remove the TODO this morning. (That's also why the screenshoots are working in the Web UI)

> > An optional recursive keyword argument, defaults to true
> i dont think we actually need this functionality with above changes.

Well, I reused the global `-R` flag as well. I think it can be useful to have non recursive templating. it can be especially useful if you want to control how sub-templates are loaded and customize their `data.values`

> i like output vs text distinction. we should also probably consider adding ability to list and get individual files "after" evaluation.

I planned for it to work like you said. It just doesn't work yet, but it should be.

mildred:
Note, I cannot use `lib.load()` because `load` is a reserved keyword in Starlark.

I'll make the following changes to this PR:

- have `lib.get()` return libraries in `_ytt_lib` if the path starts with `@`
- have `lib.external()` get a library using `NewFiles()`
- make sure `output(filename)` and `text(filename)` are working
- add tests

Also, I'd like to keep the ability to have non recursive templates. With `-R=false` from the command-line and with `recursive=False` from `lib.load()` and `lib.external()`. Imagine you have a documentation subdirectory containing many files, you don't necessarily want to render them if an occasional `.yml` file is present.

Also, you might want to organize the root folder with the main template, that is itself loading sub-templates from subfolders. Having subfolders rendered all the time is preventing you from customize them using `lib.get()`. That's why I used `_ytt_lib` in the screenshoots, otherwise I would have had the template included twice.

mildred:
As an example, here `lib: a` is included twice in the output, which is confusing. When using `lib.get()`, it's better to have ytt work in non recursive mode:
image

Also, because ytt is recursive, I couldn't have both a value file for the root directory `values.yml` and for the `a` subdirectory `a/values.yml` as a Library can only contain a single values file.

mildred:
image

mildred:
I'm having a bit of a problem creating new tests under `pkg/yamltemplate/filetests`... The templates are not evaluated with a full environment. In particular there is no `Library` object. I think I'll create end to end tests instead.

cppforlife:
> I think I'll create end to end tests instead.

can you try adding tests similar to https://github.com/k14s/ytt/blob/develop/pkg/cmd/template/cmd_data_values_test.go

cppforlife:
> Also, I'd like to keep the ability to have non recursive templates. With -R=false from the command-line and with recursive=False from lib.load() and lib.external().

i still dont think it's a good idea...

> Imagine you have a documentation subdirectory containing many files, you don't necessarily want to render them if an occasional .yml file is present.

im not sure how recursive would help here. if you are providing individual yml files via -f, its not recursive. if you provide entire directory via -f, it's still not recursive but entire directory will be considered. if you were thinking about a directories with subdirectories, then it can be argued that there should be better separation because if you provide multiple -f which are directories, how does one make some of them recursive and some non-recursive? that's why i've removed -R from all k14s CLIs, since it's not specific enough.

https://github.com/k14s/ytt/blob/30988cda7b8a9c128b29a6c3e91427ef4fdfa0da/hack/build.sh#L21-L28 shows how file marking may help with dealing with directories filled with data.

> Also, because ytt is recursive, I couldn't have both a value file for the root directory values.yml and for the a subdirectory a/values.yml as a Library can only contain a single values file.

we do have a problem of infinite loops though -- templating itself from itself. i think we should be able to disallow it upon detection. which means that when we use output("file") we shouldnt try to template other files.

mildred:
Sorry, I was not available the past days, and I'll again not be available until next week.

You're right, recursiveness is not a global setting but a per-directory setting. The thing is that specifying a single directory should be a fairly frequent use-case in which case it makes sense. When not using `@ytt:lib`, it might be useful to have non-recursive as the default because you want to customize the rendering of sub-directories via `@ytt:lib`.

A few ideas are crossing my mind about that:

- have the recursiveness value be a file mark on directories, and have the `-R` or `-R=false` flag override the default

- advertise/mandate the use of `@ytt:lib` using the parent directory (you should, use `lib.get` with a path starting with `../`). This is what [Kustomize](https://kustomize.io/) is doing for example. This will requiring re-parsing the filesystem (not working with Library objects in memory)

- always be recursive but have the ability to mark specific directories as private (like `_ytt_lib` currently). perhaps coupled with the next idea.

- use some other mechanism to customize the data values of child directories that are recursively loaded. Such as having a file in the parent directories defining a template to be used for the values of the child directories, evaluated before the child directory is evaluated with an annotation like `#@data/values-for subdir`

Pick one or many in any order. if only me I'd choose all of the last 3 items (lib.get only working for external paths, and have subdirectories handled another way with a values override instead, or marked private)

cppforlife:
@mildred i havent forgotten about this PR, just thinking a little bit more about different cases...

mildred:
I did not have time until now to follow up on this. Thinking it through, I suggest to change how files are processed and parsed. Currently, libraries and sub-libraries (sub-directories) are walked to find all processable files first, and then all files are processed at the same time. I suggest to change it this way:

- process parent directories before processing child directories
- `data/values` annotated files should only have effect in the directory they are in and their sub-directories
- Only one `data/values` annotated file will be allowed per directory, but multiple `data/values` files would be allowed in a tree hierarchy contrary to now
- `data/values` files are pre-processed using the values from their parent directory (or the command-line), and then their result is applied to the files in the current directory and sub-directories

As files from parent directories would be processed before their child directories, we could introduce a function or annotation to declare some child directories as private and not recurse in them. Removing the need to specify the recursiveness on the CLI.

mildred:
Of course, this would be a backwards compatible change because the `data/values` file present in a sub-directory would no longer have effect on files on parent directories. I would argue that this is currently a bad practice.

mildred:
I changed the implementation to update the evaluation process so it performs in a single step (loading files and `data/values` files and template evaluation in a single step). It should be relatively easy now to load values files in the context of their surrounding directory only. And to have the evaluation of files in the parent directories affect if subdirectories should be loaded or not.

mildred:
This PR has grown too big, and I cannot rebase it easily on latest develop. The main refactoring work needed for this is in #30 which does not change ytt behaviour at all, but provides necessary functions to implement library loading. It should not be long to merge.

Everything related to recursivity is separate from library loading and has been moved in #31.

Closing this PR now, I'll reopoen with improved implementation on top of #30.

PR #30, opened 2019-06-21
Refactoring: move most of initialization in pkg/workspace

mildred:
In order to allow to easily load external libraries in the future, most of the library loading that currently happens in the command-line package must be moved elsewhere so it can later be reused.

This is preliminary work to #24 and should help review later pull requests.

### DIFF
--- a/pkg/files/file.go
+++ b/pkg/files/file.go
@@ -52,7 +52,7 @@ func NewSortedFilesFromPaths(paths []string) ([]*File, error) {
 		default:
 			fileInfo, err := os.Lstat(path)
 			if err != nil {
-				return nil, fmt.Errorf("Checking file '%s'", path)
+				return nil, fmt.Errorf("Checking file '%s': %s", path, err)
 			}
 
 			if fileInfo.IsDir() {
--- a/pkg/workspace/library.go
+++ b/pkg/workspace/library.go
@@ -121,18 +121,38 @@ func (l *Library) FindFile(path string) (*files.File, error) {
 }
 
 type FileInLibrary struct {
-	File    *files.File
-	Library *Library
+	File            *files.File
+	Library         *Library
+	parentLibraries []*Library
+}
+
+func (fileInLib *FileInLibrary) RelativePath() string {
+	var components []string
+	for _, lib := range fileInLib.parentLibraries {
+		components = append(components, lib.name)
+	}
+	_, fileName := files.SplitPath(fileInLib.File.RelativePath())
+	components = append(components, fileName)
+	return strings.Join(components, pathSeparator)
 }
 
 func (l *Library) ListAccessibleFiles() []*FileInLibrary {
+	return l.listAccessibleFiles(nil)
+}
+
+func (l *Library) listAccessibleFiles(parents []*Library) []*FileInLibrary {
 	var result []*FileInLibrary
 	for _, file := range l.files {
-		result = append(result, &FileInLibrary{File: file, Library: l})
+		result = append(result, &FileInLibrary{
+			File:            file,
+			Library:         l,
+			parentLibraries: parents,
+		})
 	}
 	for _, lib := range l.children {
 		if !lib.private {
-			result = append(result, lib.ListAccessibleFiles()...)
+			newParents := append(append([]*Library{}, parents...), lib)
+			result = append(result, lib.listAccessibleFiles(newParents)...)
 		}
 	}
 	return result
--- a/pkg/workspace/library_loader.go
+++ b/pkg/workspace/library_loader.go
@@ -1,0 +1,165 @@
+package workspace
+
+import (
+	"fmt"
+
+	"github.com/k14s/ytt/pkg/files"
+	"github.com/k14s/ytt/pkg/yamlmeta"
+	"github.com/k14s/ytt/pkg/yamltemplate"
+	"github.com/k14s/ytt/pkg/yttlibrary"
+)
+
+// LibraryLoader is an object that is able to evaluate a Library given some
+// options and return an evaluation result.
+type LibraryLoader struct {
+	*Library
+	UI   files.UI
+	Opts TemplateLoaderOpts
+}
+
+// EvalResult represents the output of an evaluation.
+type EvalResult struct {
+	DocSet      *yamlmeta.DocumentSet
+	DocSets     map[*FileInLibrary]*yamlmeta.DocumentSet
+	OutputFiles []files.OutputFile
+}
+
+// EvalValuesAst represents in AST form the data.values to be fed to the
+// evaluation process.
+type EvalValuesAst interface{}
+
+func NewLibraryLoader(lib *Library, ui files.UI, loaderOpts TemplateLoaderOpts) *LibraryLoader {
+	return &LibraryLoader{
+		UI:      ui,
+		Opts:    loaderOpts,
+		Library: lib,
+	}
+}
+
+func (ll *LibraryLoader) LoadValues(parentValues EvalValuesAst) (EvalValuesAst, error) {
+	valuesFiles, err := ll.loadValueFiles(ll.Library)
+	if err != nil {
+		return nil, err
+	}
+
+	noValuesLoader := NewTemplateLoader(nil, ll.UI, ll.Opts)
+
+	vals, err := DataValuesPreProcessing{valuesFiles, parentValues, noValuesLoader, ll.Opts.IgnoreUnknownComments}.Apply()
+	if err != nil {
+		return nil, err
+	}
+
+	return vals, nil
+}
+
+func (ll *LibraryLoader) loadValueFiles(lib *Library) ([]*FileInLibrary, error) {
+	var valuesFiles []*FileInLibrary
+
+	for _, fileInLib := range lib.ListAccessibleFiles() {
+		if fileInLib.File.Type() == files.TypeYAML && fileInLib.File.IsTemplate() {
+			fileBs, err := fileInLib.File.Bytes()
+			if err != nil {
+				return nil, err
+			}
+
+			docSet, err := yamlmeta.NewDocumentSetFromBytes(fileBs, fileInLib.File.RelativePath())
+			if err != nil {
+				return nil, fmt.Errorf("Unmarshaling YAML template '%s': %s", fileInLib.File.RelativePath(), err)
+			}
+
+			tplOpts := yamltemplate.MetasOpts{IgnoreUnknown: ll.Opts.IgnoreUnknownComments}
+
+			valuesDocs, _, err := yttlibrary.DataValues{docSet, tplOpts}.Extract()
+			if err != nil {
+				return nil, err
+			}
+
+			if len(valuesDocs) > 0 {
+				valuesFiles = append(valuesFiles, fileInLib)
+				fileInLib.File.MarkForOutput(false)
+			}
+		}
+	}
+
+	return valuesFiles, nil
+}
+
+func (ll *LibraryLoader) Eval(values EvalValuesAst) (*EvalResult, error) {
+	res := &EvalResult{
+		DocSet:      &yamlmeta.DocumentSet{},
+		OutputFiles: []files.OutputFile{},
+	}
+	overlayProcess := &OverlayPostProcessing{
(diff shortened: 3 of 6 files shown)


mildred:
There is one thing I did not do in this PR but I'll need to work on is find a way to have the `DataValuesPreProcessing.Value()` result in the same format as its input (the values flags). This will allow overriding values when importing libraries.

mildred:
Updated with value type unified

mildred:
(just forgot a file)

pkg/template/core/go_value.go
```diff
@@ -28,6 +28,9 @@ func (e GoValue) AsValueWithCheckedMapKeys(forceMapStringKeys bool) interface{}
 }
 
 func (e GoValue) asStarlarkValue(val interface{}) starlark.Value {
```
cppforlife: why do we need this change in this PR?

pkg/workspace/library_loader.go
```diff
@@ -0,0 +1,190 @@
...
+func NewLibraryLoader(lib *Library, parentValuesAst interface{}, ui files.UI, loaderOpts TemplateLoaderOpts) *LibraryLoader {
+	return &LibraryLoader{
+		UI:      ui,
+		Opts:    loaderOpts,
+		Library: lib,
+		Values:  parentValuesAst,
+	}
+}
+
+func (ll *LibraryLoader) LoadValues() error {
```
cppforlife: let's also combine LoadValues() and Eval()

pkg/workspace/library_loader.go
```diff
@@ -0,0 +1,190 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (ll *LibraryLoader) loadValueFiles(lib *Library, recurse bool) ([]*FileInLibrary, error) {
+	var valuesFiles []*FileInLibrary
+
+	for _, file := range lib.files {
```
cppforlife: we shouldnt be accessing Library private methods. arguably we should probably create a separate package instead of moving this into workspace.

pkg/workspace/library_loader.go
```diff
@@ -0,0 +1,190 @@
...
+			}
+
+			if len(valuesDocs) > 0 {
+				valuesFiles = append(valuesFiles, fileInLib)
+				file.MarkForOutput(false)
+			}
+		}
+	}
+
+	if recurse {
```
cppforlife: let's take out recursion out of this PR.

pkg/workspace/library_loader.go
cppforlife: > arguably we should probably create a separate package instead of moving this into workspace.
i take this back. but we should continue to use ListAccessibleFiles() here.

pkg/template/core/go_value.go
mildred: This is not strictly needed for this PR, but it makes sense to have this conversion possible. It will be needed to convert data value objects in #32.

pkg/workspace/library_loader.go
mildred: I did not combine them because the `--data-values-inspect` flag needs to hook after data values processing but must not evaluate templates.

pkg/workspace/library_loader.go
mildred: I specifically avoided `ListAccessibleFiles()` and implemented recursion here for #31. If I want template evaluation to change the recursion strategy, I need to perform the recursion after template evaluation and not before as `ListAccessibleFiles()` would do.

I can move this code in #31 and keep `ListAccessibleFiles()` here.

pkg/workspace/library_loader.go
mildred: I specifically avoided `ListAccessibleFiles()` because I was thinking of scoping the value files relative to their directory. I mentioned it here: https://github.com/k14s/ytt/pull/24#issuecomment-503953521

The idea was that value files in subdirectories would only affect `data.values` of their directory and subdirectories but not their parent directory.

This is however not implemented in this PR which is just a refactoring, but I prepared the work here. I can remove it from this PR though and move it to a further PR.

pkg/workspace/library_loader.go
mildred: This is not the global `-R` flag we talked about in #24, but rather a preparation for what I mentioned https://github.com/k14s/ytt/pull/30#discussion_r296390561 and https://github.com/k14s/ytt/pull/24#issuecomment-503953521 where value files would be considered for each directory as they are processed and not once for all before.

This however is forward work that might belong to other PRs, I'm ok with removing it from this PR.

pkg/workspace/library_loader.go
cppforlife: > The idea was that value files in subdirectories would only affect data.values of their directory and subdirectories but not their parent directory.

im not really on board with this idea because `data.values` via data library should be returning same values throughout same library. i also dont see a compelling use case for separating data values per directory (since directory isnt a first class concept, library is though).

mildred:
Updated the PR to remove manual recursion for value files and use `ListAccessibleFiles()` instead. Also remove unneeded change in `pkg/template/core/go_value.go`.

pkg/workspace/library_loader.go
cppforlife: i see. we should then change LoadValues() to Values() which returns them and then can be passed into Eval(), which would make this object not mutable (ll.Values fields).

pkg/template/core/go_value.go
cppforlife: i think it should be part of that PR. i would like to make this PR as compact as possible and kick future work into future PRs and discussions.

pkg/files/file.go
```diff
@@ -52,7 +52,7 @@ func NewSortedFilesFromPaths(paths []string) ([]*File, error) {
 		default:
 			fileInfo, err := os.Lstat(path)
 			if err != nil {
-				return nil, fmt.Errorf("Checking file '%s'", path)
+				return nil, fmt.Errorf("Checking file '%s', %v", path, err)
```
cppforlife: good catch. let's change to `fmt.Errorf("Checking file '%s': %s", path, err)` to be more consistent with other error wraps

pkg/workspace/library.go
```diff
@@ -15,6 +15,7 @@ const (
 )
 
 type Library struct {
+	parent   *Library
```
cppforlife: let's also kick Library changes into other PRs, as they are not necessary for this refactor.

pkg/workspace/library_loader.go
mildred: Library objects are split per directory, that's what I meant. Currently, there is no real separation between a Library and its child libraries, but we can introduce value files that would not affect the parent libraries as they do currently.

pkg/workspace/library.go
mildred: I need it to construct the relative path.

The relative path in file.go is not enough, because the relative path at file loading time is not the same as the relative path at evaluation time in case the evaluated library is not the root library.

I can probably avoid this by constructing the relative path at evaluation time though. I'll try that.

pkg/template/core/go_value.go
mildred: This is removed from the PR now.

pkg/workspace/library_loader.go
mildred: Ok, I'm changing the LoadValues function to take parent values override and return the values from the value file with the parent override. Eval() will now take a value parameter.

mildred:
Here are the latest changes from my push: https://github.com/k14s/ytt/compare/9f69e36a324af5a94a72c602132ad8029042b3aa..849de6150f6ce93d56fc23d7fee2ea37aad43f70

I updated:
- LoadValues() dot not change the LibraryLoader any more and Eval() takes a value argument
- Library.parent and Library.RelativePath() removed, replaced by live relative path construction in evaluation routine and stored in FileInLibrary.Path
- I found out that when rebasing my work on top of develop, I was not using the correct relative path and I fixed that. the relative path used for the output files must be relative to the library used to construct the LibraryLoader, not the root library loaded from the filesystem
- Updated files.go error string as you suggested

mildred:
About `LoadValues()`, there is a way to merge it with `Eval()`: provide in LibraryLoader a function pointer to a callback that can be used to inspect the values after they are loaded. I'm not sure this is a good idea, and I already implemented taking values as argument and returning transformed values. but it can be done.

pkg/workspace/library_loader.go
```diff
@@ -0,0 +1,182 @@
...
+		res.OutputFiles = append(res.OutputFiles, outputFile)
+	}
+
+	return res, nil
+}
+
+func (ll *LibraryLoader) evalLibrary(lib *Library, libPath string, values EvalValuesAst, res *EvalResult, overlayProcess *OverlayPostProcessing) error {
+	loader := NewTemplateLoader(values, ll.UI, ll.Opts)
+
+	for _, file := range lib.files {
```
mildred: Use `ListAccessibleFiles` instead

pkg/workspace/library_loader.go
```diff
@@ -0,0 +1,182 @@
...
+
+			ll.UI.Debugf("### %s result\n%s", fileInLib.Path, resultStr)
+			res.OutputFiles = append(res.OutputFiles, files.NewOutputFile(fileInLib.Path, []byte(resultStr)))
+
+		default:
+			return fmt.Errorf("Unknown file type")
+		}
+	}
+
+	for _, subLib := range lib.children {
```
mildred: not needed with `ListAccessibleFiles`

pkg/workspace/library_loader.go
mildred: Except, relative paths needs to be constructed, so unless it's constructed in ListAccessibleFiles, we'll need to keep the current form

pkg/workspace/library.go
```diff
@@ -123,6 +123,7 @@ func (l *Library) FindFile(path string) (*files.File, error) {
 type FileInLibrary struct {
 	File    *files.File
 	Library *Library
+	Path    string
```
mildred: Only add a `parentLibraries []*Library` (private) and make a `fileInLib.RelativePath()` string function to use it. Populate in `ListAccessibleFiles`

pkg/workspace/library_loader.go
```diff
@@ -0,0 +1,182 @@
...
+
+		ll.UI.Debugf("### %s result\n%s", fileInLib.Path, resultDocBytes)
+		outputFile := files.NewOutputFile(fileInLib.Path, resultDocBytes)
+		res.OutputFiles = append(res.OutputFiles, outputFile)
+	}
+
+	return res, nil
+}
+
+func (ll *LibraryLoader) evalLibrary(lib *Library, libPath string, values EvalValuesAst, res *EvalResult, overlayProcess *OverlayPostProcessing) error {
```
cppforlife: i think we can also get rid of libPath since we will be relying on Library objects for relative paths

pkg/workspace/library.go
```diff
@@ -121,18 +121,38 @@ func (l *Library) FindFile(path string) (*files.File, error) {
...
+		result = append(result, &FileInLibrary{
+			File:            file,
+			Library:         l,
+			parentLibraries: parents,
+		})
 	}
 	for _, lib := range l.children {
 		if !lib.private {
-			result = append(result, lib.ListAccessibleFiles()...)
+			newParents := append(parents, lib)
```
cppforlife: this may actually modify parents slice. let's change this to `append(append([]*Library{}, parents...), lib)`

cppforlife:
left one comment, otherwise looks good. would you mind rebasing this on top of develop last time before we merged this in (to include this: https://github.com/k14s/ytt/compare/c6e9032f32614f9a2afa5883ee990356cbd9023c...a744459dca403387fd321db836dbe16219c6c734)?

pkg/workspace/library.go
mildred: I don't think this would happen because `parents` should always be shorter and `append()` would always need to increase the capacity and allocate a new array, but you're right, better be safe.

mildred:
done

cppforlife:
merged 🤘. i did a little light renaming in follow up commit: https://github.com/k14s/ytt/commit/e72d8ad826ea49abde719ac82f63cdb44bb2a992

PR #31, opened 2019-06-21
Recurse annotations

mildred:
Based on #30 this PR adds two annotations that allow customizing the recursive behaviour of ytt. Thanks to the refactoring, we can evaluate template in parent directories before child directories, and the template evaluation can add directives to tell if ytt should recurse to subdirectories, or to tell if a specific subdirectory should be marked as private (in addition to `_ytt_lib` special subdirectory).

- `#@lib/private "subdir"`: tells to mark the specified subdirectory as private
- `#@lib/norecurse`: tells to mark all subdirectories as private


mildred:
Updated on top of latest refactoring

cppforlife:
after thinking more about this proposed change, i still favor _ytt_lib approach to making things private (even in libraries). i am closing this at this point. feel free to reopen if further discussion is to take place.

mildred:
I think it does not appear clearly, and perhaps I should have opened an issue about it, but the idea there was not to replace _ytt_lib by something else, but to control the recursion behaviour.

There is a use case for it : you want to evaliate a ytt file at the top level of your repositories. You have _ytt_lib you want to use, and you have other directories containing files that ytt will attempt to parse and fail (because it contains unrelated things). Having an option to prevent ytt recursing to directories not containing ytt files is useful.

cppforlife:
> directories containing files that ytt will attempt to parse and fail (because it contains unrelated things)
i would argue then either those files shouldnt be present there (removed during vendoring operation or something like that) or ytt invocation should be more explicit what to include. 

it's also a bit weird that file included in a directory would control visibility of its sibling directories (library.children). i could see visibility being controller separately as a feature but i would have to see more concrete examples for why.

PR #32, opened 2019-06-21

mildred:
Based on #30 and replaces #24.

@ytt:lib can be loaded from within starlark and it provides a `lib`
modules with `lib()` and `enternal()` functions. They takes:

  - A string with the location of the module to load
  - An optional value object to override current data/values

After evaluation, the returned object has an `output()` method to fetch
the result from the library evaluation. It takes an optional string
argument for the file name to get the results from. If empty, it will
return the concatenated results for all files. To get results as string,
there is an equivalent functtion called `text()`.

This is quite equivalent as running ytt from within the template and
embedding the output in it.

`external()` loads files from disk (for example if you have to use the
parent directories) and `lib()` from already loaded files in the current
directory.


mildred:
Updated the PR to only allow internal loading, removing the ability to load arbitrary files from disk.

Renamed `@ytt:lib.get()` to `@ytt:lib.eval()` because it evaluated the library. Getting it is not really relevant with internal libraries where every file is already loaded in memory.

mildred:
Replaced by #48

PR #43, opened 2019-07-15
return error when node value evaluation is ambiguous

### LINKED ISSUE #38: [lang] add error message when both existing yaml node value and computed value are present?
Input:
```
x: #@ { "first": "is included"}
  second: is ignored
```
The second line is silently ignored.

Also here:

```
--- #@ { "first": "is included"}
this line: is ignored too
```

If this is not valid, I'd expect an error message instead of silently ignoring part of the input.

nimakaviani:
fixes #38

### DIFF
--- a/pkg/yamltemplate/metas.go
+++ b/pkg/yamltemplate/metas.go
@@ -61,6 +61,12 @@ func NewMetas(node yamlmeta.Node, opts MetasOpts) (Metas, error) {
 
 				if node.GetPosition().IsKnown() {
 					if meta.Position.Line() == node.GetPosition().Line() {
+						if len(node.GetValues()) > 0 && node.GetValues()[0] != nil {
+							return metas, fmt.Errorf(
+								"Expected YAML node at %s to have either computed or YAML value, but found both",
+								meta.Position.AsString())
+						}
+
 						ann.Name = template.AnnotationValue
 					}
 				}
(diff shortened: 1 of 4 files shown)


pkg/yamltemplate/metas.go
```diff
@@ -61,6 +61,13 @@ func NewMetas(node yamlmeta.Node, opts MetasOpts) (Metas, error) {
 
 				if node.GetPosition().IsKnown() {
 					if meta.Position.Line() == node.GetPosition().Line() {
+						// return error when node has both inilne code and children
```
cppforlife: let's remove this comment.

pkg/yamltemplate/metas.go
```diff
@@ -61,6 +61,13 @@ func NewMetas(node yamlmeta.Node, opts MetasOpts) (Metas, error) {
...
+						if len(node.GetValues()) > 0 && node.GetValues()[0] != nil {
+							return metas, fmt.Errorf(
+								"Ambiguous code and value assignment at %s for node with children",
```
cppforlife: Expected YAML node at %s to have either computed or YAML value, but found both

cppforlife:
@nimakaviani left a few comments to fix before merging.

nimakaviani:
fixed.

PR #44, opened 2019-07-22
first pass on allowing keys to be templated

nimakaviani:


nimakaviani:
The implementation is done but I am thinking we want to have some discussions on some of the implementation decisions before we push this PR through. I made the PR so that we can track some fo the conversations.

pkg/template/evaluation_ctx.go
```diff
@@ -88,6 +89,21 @@ func (e *EvaluationCtx) TplSetNode(thread *starlark.Thread, _ *starlark.Builtin,
 	return starlark.None, nil
 }
 
+// args(nodeTag, value Value)
+func (e *EvaluationCtx) TplSetNodeKey(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() > 1 {
+		if _, noop := args.Index(1).(*core.StarlarkNoop); !noop {
+			node := e.parentNodes[len(e.parentNodes)-1]
+			if item, ok := node.(*yamlmeta.MapItem); ok {
```
nimakaviani: Not super stoked that we had to cast back to `yamlmeta.MapItem` in the evaluation context in order to set the key. But it seemed like adding a `SetKey` method to `EvaluationNode` was not also a good idea since key can only be set for a `MapItem` and not for other yaml elements. Hence leaving it the way it is for now.

cppforlife:
not too hot about adding specialized instruction for setting map key. thinking a bit more about this, may be an alternative impl could be turning map item into 2 nodes: map item key and map item value. not sure about the impact, but i think it might be interesting...

nimakaviani:
That's an interesting idea. It will unify the evaluation for nodes but will probably have a larger cascading impact throughout the code.

I am on the fence though on whether or not it is worth the effort. if a full blown templating of keys is in the picture then sure let's do it (something like what was discussed [here](https://github.com/k14s/ytt/issues/29#issue-459087360)), but if it will end up only writing a string value (which I believe will be the primary use case), then the debt it will add to the code may be more than the value we get out of it.

nimakaviani:
the PR is updated based on the conversation with @cppforlife . we decided to remove an instruction for MapItem.Key and do the casting when setting the value in the evaluation context.

pkg/yamltemplate/template.go
```diff
@@ -93,6 +93,14 @@ func (e *Template) build(val interface{}, parentNode yamlmeta.Node, parentTag te
 		SourceLine:  e.newSourceLine(node.GetPosition()),
 	})
 
+	if typedNode, ok := val.(*yamlmeta.MapItem); ok {
+		templateLines, err := e.buildString(typedNode.Key.(string), parentNode, parentTag, e.instructions.NewSetMapItemKey)
```
cppforlife: i think we also need to add `if keyStr, ok := typedNode.Key.(string); ok {` otherwise non-string keys will panic.

pkg/template/evaluation_ctx.go
```diff
@@ -88,6 +89,21 @@ func (e *EvaluationCtx) TplSetNode(thread *starlark.Thread, _ *starlark.Builtin,
 	return starlark.None, nil
 }
 
+// args(nodeTag, value Value)
+func (e *EvaluationCtx) TplSetMapItemKey(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() > 1 {
+		if _, noop := args.Index(1).(*core.StarlarkNoop); !noop {
```
cppforlife: dont think we need to worrry about StarlarkNoop here. we need it for TplSetNode because of `template.replace` usage.

pkg/template/evaluation_ctx.go
cppforlife: we should probably panic here with message "Expected node '%T' to be mapitem" as this is an internal inconsistency error

pkg/yamltemplate/filetests/string-keys.yml
```diff
@@ -0,0 +1,9 @@
+#@ for x in [1,2,3]:
```
cppforlife: let's add one more test that uses ints/etc. as yaml keys.

nimakaviani:
oh huh pushed the first fix too soon. will address the other comments later today.

PR #48, opened 2019-08-05

mildred:
Replaces #32 and implements proposal at https://github.com/k14s/ytt/issues/18#issuecomment-511040240

Just a few remarks to this proposal:

- What's the difference between `.text()` and `.read()` ? The way I see it, they are identical (except read takes a single value and text takes a list). I did not implement `.read()`.

- In a file at `dir/mod0.yaml`, where should `load("@library:module.yaml")` look for its file? Would it be `dir/_ytt_lib/library/module.yaml` (as it is currently) or should that change to `_ytt_lib/library/module.yaml` instead (breaking the current behaviour). I implemented the second option which seemed to be what was implied by the proposal although it will introduce an incompatibility.

Also, I did not refactor the code to account for the library/package/module distinction, I did not want to include too much on this PR that's already quite  big. However, I needed to introduce a new PackageInLibrary object that represents a package and gives access to the root library object. It works quite well with the FileInLibrary object too.

I made many functions package aware, and modified the Starlark built-in `load()` statement to use the same logic as the `lib.get()` function. This way, we only have a single place in the code where we are parsing library paths. Most changes in `template_loader.go` that weren't there in previous PRs are because of that.


mildred:
@cppforlife I just rebased the PR to remove the conflicts, could you take a look at it ?

Hopefully it could be merged soon

docs/lang-ref-ytt.md
```diff
@@ -100,6 +100,66 @@ md5.sum("data") # "8d777f385d3dfec8815d20f7496026dc"
 sha256.sum("data") # "3a6eb0790f39ac87c94f3856b2dd2c5d110e6811602261a9a923d3bb23adc8b7"
 ```
 
+#### Library Loading
```
cppforlife: let's separate this into dedicated file: `See [Library loading specific docs](lang-ref-ytt-lib.md).`

docs/lang-ref-ytt.md
```diff
@@ -100,6 +100,66 @@ md5.sum("data") # "8d777f385d3dfec8815d20f7496026dc"
...
+
+    @ytt:lib can be loaded from within starlark and it provides a `lib` module
+    with a `lib.get()` function that returns a library object.
+
+    The library name `LIBNAME` must match one of the forms:
+
+    - `@LIBNAME:[PACKAGE/MODULE]`: loads a library name. The library is found in
+      the `_ytt_lib` directory at the root of the current library.
+
+    - `LIBPATH:[PACKAGE/MODULE]`: loads a library path relative to current
```
cppforlife: dont think we should support this variation, as all "child" libraries should be under _ytt_lib (above case)

docs/lang-ref-ytt.md
```diff
@@ -100,6 +100,66 @@ md5.sum("data") # "8d777f385d3dfec8815d20f7496026dc"
...
+       file location.
+
+    - `@:[PACKAGE/MODULE]`: loads the current library. The package specified is
+      loaded relative to the current library root.
+
+    - `PACKAGE/MODULE`: loads a package in the current library relative to the
+       current executing file location.
+
+    If a package and module is specified, only this package and module will be
+    evaluated. Package evaluation is always recursive.
```
cppforlife: > If a package and module is specified, only this package and module will be
    evaluated. Package evaluation is always recursive.

we should move this into the next section as this section just talks about format of load paths.

pkg/yttlibrary/overlay/map.go
```diff
@@ -19,7 +21,7 @@ func (o OverlayOp) mergeMapItem(leftMap *yamlmeta.Map, newItem *yamlmeta.MapItem
 
 	leftIdx, found, err := ann.Index(leftMap)
 	if err != nil {
-		return err
+		return fmt.Errorf("Searching in map for item: %s", err)
```
cppforlife: are changes to overlay library just general improvements? if so let's split that into a separate PR.

pkg/yttlibrary/overlay/op.go
```diff
@@ -62,22 +62,26 @@ func (o OverlayOp) apply(left, right interface{}, parentMatchChildDefaults Match
 		for _, item := range typedRight.Items {
 			item := item.DeepCopy()
 
+			var errPrefix string = "Overlaying"
```
cppforlife: errPrefix := "Overlaying"

pkg/yttlibrary/overlay/op.go
```diff
@@ -62,22 +62,26 @@ func (o OverlayOp) apply(left, right interface{}, parentMatchChildDefaults Match
...
 					err = o.replaceMapItem(typedLeft, item, parentMatchChildDefaults)
+					errPrefix = "Replacing"
 				default:
 					err = fmt.Errorf("Overlay op %s is not supported on map item", op)
 				}
 			}
 			if err != nil {
-				return false, fmt.Errorf("Map item (key '%s') on %s: %s",
-					item.Key, item.Position.AsString(), err)
+				return false, fmt.Errorf("%s map item (key '%s') on %s: %s",
```
cppforlife: if we are changing for map items, we should probably change it for arrays, docs, etc.

pkg/yttlibrary/overlay/map_item_match_annotation.go
```diff
@@ -50,7 +50,15 @@ func (a MapItemMatchAnnotation) Index(leftMap *yamlmeta.Map) (int, bool, error)
 		count = 1
 	}
 
-	return idx, found, a.expects.Check(count)
+	err := a.expects.Check(count)
+	if err != nil {
+		if found {
+			err = fmt.Errorf("Found item: %s", err)
```
cppforlife: check errors shouldnt be augmented here as only Check() inside knows what expected (found item vs not found).

pkg/yttlibrary/overlay/map.go
cppforlife: we should also move err augmentation into Index().

pkg/yttlibrary/lib/api.go
```diff
@@ -0,0 +1,283 @@
...
+
+type libraryToLoad struct {
+	loader  eval.Loader
+	libPath string
+	values  []eval.ValuesAst
+	result  *eval.Result
+	evalErr error
+}
+
+func (l *libraryToLoad) starlark() starlark.Value {
```
cppforlife: starlark() -> AsStarlarkValue(); also should be public (even though object itself is private).

pkg/yttlibrary/lib/api.go
```diff
@@ -0,0 +1,283 @@
...
+
+func (l *libraryToLoad) DataValues(
+	thread *starlark.Thread, f *starlark.Builtin,
+	args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+
+	if args.Len() != 1 {
+		return starlark.None, fmt.Errorf("expected one argument")
+	}
+
+	argsvalues := yamlmeta.NewASTFromInterface(core.NewStarlarkValue(args.Index(0)).AsInterface())
```
cppforlife: argsvalues -> values

pkg/yttlibrary/lib/api.go
```diff
@@ -0,0 +1,283 @@
...
+func (l *libraryToLoad) Yaml(
+	thread *starlark.Thread, f *starlark.Builtin,
+	args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+
+	err := l.ensureLoaded()
+	if err != nil {
+		return starlark.None, err
+	}
+
+	if args.Len() == 0 {
```
cppforlife: use switch instead of if-elif-else

pkg/yttlibrary/lib/api.go
```diff
@@ -0,0 +1,283 @@
...
+
+	} else if args.Len() == 1 {
+		arg0 := core.NewStarlarkValue(args.Index(0)).AsInterface()
+
+		list, ok := arg0.([]interface{})
+		if !ok {
+			return starlark.None, fmt.Errorf("expected argument to be a list of strings")
+		}
+
+		var resDocSet *yamlmeta.DocumentSet
```
cppforlife: let's always return DocumentSet, even if it's empty. it's good to be consistent with return value types from a single function.

pkg/yttlibrary/lib/api.go
```diff
@@ -0,0 +1,283 @@
...
+	err := l.ensureLoaded()
+	if err != nil {
+		return starlark.None, err
+	}
+
+	if args.Len() == 0 {
+		return yamltemplate.NewStarlarkFragment(l.result.DocSet), nil
+
+	} else if args.Len() == 1 {
+		arg0 := core.NewStarlarkValue(args.Index(0)).AsInterface()
```
cppforlife: arg0 -> moduleNames (or pkgModuleNames or somethign like that).

pkg/yttlibrary/lib/api.go
```diff
@@ -0,0 +1,283 @@
...
+
+		list, ok := arg0.([]interface{})
```
cppforlife: list -> typedModuleNames

pkg/yttlibrary/lib/api.go
```diff
@@ -0,0 +1,283 @@
...
+		if !ok {
+			return starlark.None, fmt.Errorf("expected argument to be a list of strings")
+		}
+
+		var resDocSet *yamlmeta.DocumentSet
+
+		for _, listitem := range list {
```
cppforlife: listitem -> typedModuleName

pkg/yttlibrary/lib/api.go
```diff
@@ -0,0 +1,283 @@
...
+	thread *starlark.Thread, f *starlark.Builtin,
+	args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+
+	err := l.ensureLoaded()
+	if err != nil {
+		return starlark.None, err
+	}
+
+	if args.Len() == 0 {
+		return yamltemplate.NewStarlarkFragment(l.result.DocSet), nil
```
cppforlife: is `l.result.DocSet` order deterministic?

pkg/template/core/go_value.go
```diff
@@ -25,6 +25,9 @@ func (e GoValue) AsStarlarkValue() starlark.Value {
 }
 
 func (e GoValue) asStarlarkValue(val interface{}) starlark.Value {
+	if obj, ok := val.(starlark.Value); ok {
```
cppforlife: why do we need this?

pkg/yttlibrary/lib/api.go
```diff
@@ -0,0 +1,283 @@
...
+	l := &libraryToLoad{
+		loader:  m.loader,
+		libPath: libPath,
+		values:  []eval.ValuesAst{},
+	}
+
+	return l.starlark(), nil
+}
+
+type libraryToLoad struct {
```
cppforlife: libraryToLoad -> libModuleLibrary

pkg/template/core/go_value.go
```diff
@@ -66,6 +69,9 @@ func (e GoValue) asStarlarkValue(val interface{}) starlark.Value {
 	case []interface{}:
 		return e.listAsStarlarkValue(typedVal)
 
+	case *starlark.Builtin:
```
cppforlife: same as above.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -94,17 +95,21 @@ func (p DataValuesPreProcessing) overlay(valuesDoc, newValuesDoc *yamlmeta.Docum
...
+	for i, valuesFlagsAst := range p.valuesFlagsAst {
+		astFlagValues := &yamlmeta.Document{Value: valuesFlagsAst}
 
-	result, err := p.overlay(valuesDoc, astFlagValues)
-	if err != nil {
-		return nil, fmt.Errorf("Overlaying data values from flags (provided via --data-value-*) on top of data values from files (marked as @data/values): %s", err)
+		valuesDoc, err = p.overlay(valuesDoc, astFlagValues)
+
+		if err != nil {
+			return nil, fmt.Errorf("Overlaying (%d) data values from flags (provided via --data-value-* or library loading) on top of data values from files (marked as @data/values): %s", i, err)
```
cppforlife: im not sure `i` help here to the end user. is there something better we can do?

pkg/workspace/data_values_pre_processing.go
```diff
@@ -94,17 +95,21 @@ func (p DataValuesPreProcessing) overlay(valuesDoc, newValuesDoc *yamlmeta.Docum
 }
 
 func (p DataValuesPreProcessing) overlayFlags(valuesDoc *yamlmeta.Document) (*yamlmeta.Document, error) {
+	var err error
```
cppforlife: can move `var err error` above p.overlay

pkg/workspace/library.go
```diff
@@ -120,33 +98,96 @@ func (l *Library) FindFile(path string) (*files.File, error) {
...
-func (fileInLib *FileInLibrary) RelativePath() string {
+func (packageInLib *PackageInLibrary) Package() *Library {
+	return packageInLib.pkgPath[len(packageInLib.pkgPath)-1]
+}
+
+func (packageInLib *PackageInLibrary) Library() *Library {
+	return packageInLib.pkgPath[0]
+}
+
+var LoaderRegexp = regexp.MustCompile("^((@?)([^:]*):)?(.*)$")
```
cppforlife: should use \A...\z instead of ^...$

cppforlife: lets also make a dedicated class LoadPath that has Library(), Package(), Module() accessors. reading code below that uses matches[x] is quite tedious. we should also add some unit tests that test out different variations.

pkg/eval/eval.go
```diff
@@ -0,0 +1,28 @@
+package eval
+
+import (
+	"github.com/k14s/ytt/pkg/files"
+	"github.com/k14s/ytt/pkg/yamlmeta"
+	"go.starlark.net/starlark"
+)
+
+type Loader interface {
```
cppforlife: since TemplateLoader sits within workspace for now, are we able to combine pkg/eval with pkg/workspace? i might separate them eventually but it seems to be a scope creep for this PR.

pkg/workspace/library.go
```diff
@@ -3,14 +3,16 @@ package workspace
...
 	"strings"
 
 	"github.com/k14s/ytt/pkg/files"
 )
 
 const (
-	privateName   = "_ytt_lib"
+	libLocation   = "_ytt_lib"
+	privateName   = libLocation
 	pathSeparator = "/"
```
cppforlife: i think we can get rid of pathSeparator and use files.{SplitPath|JoinPath}

cppforlife:
left several comments to start it off.

cppforlife:
@mildred ive spent some quality time thinking about this feature. ive also unfortunately had a hard time reviewing this pr (tried to split it up into smaller sections), so ended up rewriting from scratch and separating into multiple commits: https://github.com/k14s/ytt/compare/4ec3e574d0b4bf909231926670267c291c85a2fa...820fc899efdac460117c06ae958fb9da43190822.

current impl does not deal with accessing "self" library and only allows files to access _ytt_lib in the same folder (ie does not keep track of "root" of library). ive documented exposed api here: https://github.com/k14s/ytt/blob/develop/docs/lang-ref-ytt-library.md. 

i'd like to say thank you for pushing this work and thinking thru different cases. our brainstorming in slack have been a useful foundation for getting this feature out. this feature is available in the latest release.

mildred:
Hi, Thank you for getting this through and I'm sorry for not following up this PR since late september. I got many things to around that time and couldn't find time for this. It was in the top spot of my todolist, but looked over :)

I believed the code I had in this PR was quite understandable, and I could not really make smaller functional increments. I'm sorry to hear that the review was difficult.

I did not looked at all the details but from what you added to the documentation, it seems like you did a great job of porting this over and there is now something nice to use.

cppforlife:
> I believed the code I had in this PR was quite understandable, and I could not really make smaller functional increments. I'm sorry to hear that the review was difficult.

i think it was mostly related to one large commit vs series of smaller ones. typically it easier to follow line of impl thought thru smaller commits. but that may be just my own thinking.

if you do end up using this feature do let us know how it goes. thanks again for contributing.

Issue #76, opened 2019-12-24
[overlay] include list of matched locations when overlay match matches unexpected number of nodes


cppforlife:
currently error message just states that x matched and it was expecting y. it might be inconvenient to find what is being matched (potentially accidently). (example: https://kubernetes.slack.com/archives/CH8KCCKA5/p1577132571199400)

let's add list of locations that were matched for better introspecting.

cppforlife:
included in https://github.com/k14s/ytt/releases/tag/v0.26.0

PR #91, opened 2020-02-08
When overlay fails expects, report position of matches

pivotaljohn:
_(this is a sketch, intended to spark dialogue and critique)_

aimed at #76

### DIFF
--- a/pkg/cmd/template/cmd_overlays_test.go
+++ b/pkg/cmd/template/cmd_overlays_test.go
@@ -228,9 +228,7 @@ overlayed: true
 
 	expectedErr := "Overlaying (in following order: overlay.yml): " +
 		"Document on line overlay.yml:4: " +
-		"Expected number of matched nodes to be 1, but was 2\n" +
-		"line tpl.yml:2\n" +
-		"line tpl.yml:6"
+		"Expected number of matched nodes to be 1, but was 2 (lines: tpl.yml:2, tpl.yml:6)"
 
 	if out.Err.Error() != expectedErr {
 		t.Fatalf("Expected error to match '%s' but was '%s'", expectedErr, out.Err.Error())
--- a/pkg/cmd/template/cmd_overlays_test.go
+++ b/pkg/cmd/template/cmd_overlays_test.go
@@ -194,3 +194,45 @@ map: {}
 		t.Fatalf("Expected error to match string but was '%s'", out.Err.Error())
 	}
 }
+
+func TestDocumentOverlayMultipleMatchesDescriptiveError(t *testing.T) {
+	yamlTplData := []byte(`
+---
+id: 1
+name: foo
+
+---
+id: 2
+name: foo
+`)
+
+	yamlOverlayTplData := []byte(`
+#@ load("@ytt:overlay", "overlay")
+#@overlay/match by=overlay.all
+---
+overlayed: true
+`)
+
+	filesToProcess := files.NewSortedFiles([]*files.File{
+		files.MustNewFileFromSource(files.NewBytesSource("tpl.yml", yamlTplData)),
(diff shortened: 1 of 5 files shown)


pkg/yttlibrary/overlay/array_item_match_annotation.go
```diff
@@ -66,7 +66,12 @@ func (a ArrayItemMatchAnnotation) Indexes(leftArray *yamlmeta.Array) ([]int, err
 		return nil, err
 	}
 
-	return idxs, a.expects.Check(len(idxs))
+	nodes := MatchingNodes{}
```
pivotaljohn: @cppforlife one "draft" aspect of the PR is that — all things being equal — would expect `MatchNodes` to return the collection of matching nodes.  However, there is downstream logic that expects a collection of indexes.

A question I'm exploring: what made passing around indexes of matches a suitable tactic, to date? ... does this PR/feature now make passing around references _more_ suitable (or not)?

cppforlife: indexes are used to reference to a node within its parent (eg array item within array). since things can be removed you need stable ref to what to remove.

jtigger:
@cppforlife this PR is ready for a proper review / merge (I don't have permissions to mark it as ready for review).

pkg/cmd/template/cmd_overlays_test.go
```diff
@@ -194,3 +194,45 @@ map: {}
...
+	opts := cmdtpl.NewOptions()
+
+	out := opts.RunWithFiles(cmdtpl.TemplateInput{Files: filesToProcess}, ui)
+	if out.Err == nil {
+		t.Fatalf("Expected RunWithFiles to error")
+	}
+
+	expectedErr := "Overlaying (in following order: overlay.yml): " +
+		"Document on line overlay.yml:4: " +
+		"Expected number of matched nodes to be 1, but was 2\n" +
```
cppforlife: we should probably just use parenthesis with commas (e.g. `but was 2 (line tpl.yml:2, line tpl.yml:6)`). multiline errors for now arent formatted properly in nested errors.

pkg/yttlibrary/overlay/match_annotation_expects_kwarg.go
```diff
@@ -115,18 +119,29 @@ func (a MatchAnnotationExpectsKwarg) checkString(typedVal starlark.String, num i
...
-		lastErr = a.checkValue(val, num)
+		lastErr = a.checkValue(val, matches)
 		if lastErr == nil {
 			return nil
 		}
 	}
 	return lastErr
 }
+
+func formatPositions(pos []*filepos.Position) string {
```
cppforlife: lets place formatPositions on MatchAnnotationExpectsKwarg. you can use `func (MatchAnnotationExpectsKwarg) formatPositions(...` sig with naming `MatchAnnotationExpectsKwarg`.

cppforlife:
left a few comments. otherwise look good.

pkg/cmd/template/cmd_overlays_test.go
jtigger: Yeah, I invested the least attention to this aspect having no real idea of what would work.  The format you suggest reads better than one-per-line.

pkg/yttlibrary/overlay/match_annotation_expects_kwarg.go
jtigger: Yeah, not sure why I didn't do that: it's an implementation detail of `#Check()`.

cppforlife:
merged. minor fixes added: https://github.com/k14s/ytt/compare/a3c0f3df6338d184157f6e473f54664923335976...1227be5b00e1d114492e03087b911808fbe0863c

Issue #90, opened 2020-02-07
[overlay] support matching multiple map items via overlay.all


cppforlife:
would like for following to support

```yaml
#@ load("@ytt:overlay", "overlay")
#@overlay/match by=overlay.subset({"uaa": true})
---
clients:
  #@overlay/match by=overlay.all,expects="1+"
  _:
    secret: foo
```

to support updating following:

```yaml
---
uaa: true
clients:
  client1:
    secret: blah1
  client2:
    secret: blah2
```

to result in

```yaml
---
uaa: true
clients:
  client1:
    secret: foo
  client2:
    secret: foo
```

jtigger:
I assume that we'd technically want an `#@overlay/match` on the parent (i.e. `clients`) to indicate that map is an overlay... yeah?  Or is it in scope for this work to infer that if any child node is annotated with `#@overlay...` that the whole graph is an overlay?

cppforlife:
@jtigger snippets were implying that they were inside a regular overlay yaml doc. updated that to make that more clear.

cppforlife:
to be clear this should be generic behaviour so something like this should also work:

```yaml
#@ load("@ytt:overlay", "overlay")
#@overlay/match by=overlay.subset({"uaa": true})
---
clients:
  #@overlay/match by=overlay.subset({}),expects="1+"
  _:
    secret: foo
```

jtigger:
Yeah, I read this as, "implement the `by` keyword for map items match."

cppforlife:
included in https://github.com/k14s/ytt/releases/tag/v0.26.0

PR #93, opened 2020-02-13
Support matching multiple map items

pivotaljohn:
... so that operations on maps can be applied to multiple (matching) items.

(addresses #90)


pkg/cmd/template/cmd_overlays_test.go
```diff
@@ -146,6 +146,73 @@ yamlfunc: yamlfunc`)
...
+    secret: ree
+  client2:
+    secret: ree
+  client3:
+    secret: ree
+`
+
+	filesToProcess := files.NewSortedFiles([]*files.File{
+		files.MustNewFileFromSource(files.NewBytesSource("tpl.yml", yamlTplData)),
+		// Note that overlay1 is alphanumerically before overlay2
```
cppforlife: dont think we need this comment here

pkg/yttlibrary/overlay/map_item_match_annotation.go
```diff
@@ -2,15 +2,19 @@ package overlay
...
 import (
 	"fmt"
-	"reflect"
-
 	"github.com/k14s/ytt/pkg/template"
+	tplcore "github.com/k14s/ytt/pkg/template/core"
 	"github.com/k14s/ytt/pkg/yamlmeta"
+	"github.com/k14s/ytt/pkg/yamltemplate"
 	"go.starlark.net/starlark"
+	"reflect"
```
cppforlife: std imports first, line break, all other imports.

pkg/yttlibrary/overlay/map_item_match_annotation.go
```diff
@@ -42,22 +49,55 @@ func NewMapItemMatchAnnotation(newItem *yamlmeta.MapItem,
...
 
-	return idx, found, a.expects.Check(count)
+	return idxs, a.expects.Check(len(idxs))
 }
 
-func (a MapItemMatchAnnotation) MatchNode(leftMap *yamlmeta.Map) (int, bool) {
-	for i, item := range leftMap.Items {
-		if reflect.DeepEqual(item.Key, a.newItem.Key) {
-			return i, true
+func (a MapItemMatchAnnotation) MatchNode(leftMap *yamlmeta.Map) ([]int, error) {
```
cppforlife: -> MatchNodes

pkg/cmd/template/cmd_overlays_test.go
jtigger: Nope: copy-pasta.

jtigger:
@cppforlife heads-up: starting to run into a number of (rougher-than-typical) merge conflicts.

docs/lang-ref-ytt-overlay.md
```diff
@@ -67,30 +67,55 @@ Annotations on the "right-side" nodes:
...
+   this example will successfully match array items that have a map with a key `name` of value `item2`;
+    ```yaml
+    #@overlay/match by=overlay.map_key("name")
+    - name: item2
+    ```
+   likewise, this example matches items in a map that have a key `name` of value `item2` (note: the key name `_` is arbitrary and ignored);
+    ```yaml
+    #@overlay/match by=overlay.map_key("name")
+    _:
+      name: item2
```
cppforlife: may be rename inner name key to avoid confusion with what's being matched here?

pkg/cmd/template/cmd_overlays_test.go
```diff
@@ -146,6 +146,195 @@ yamlfunc: yamlfunc`)
 	}
 }
 
+func TestDocumentOverlaysMapItems(t *testing.T) {
```
cppforlife: any reason why do these tests here vs in https://github.com/k14s/ytt/tree/develop/pkg/yamltemplate/filetests/ytt-library as individual files?

docs/lang-ref-ytt-overlay.md
cppforlife: doh, i misunderstood what was the going on. ignore this...

pkg/cmd/template/cmd_overlays_test.go
```diff
@@ -146,6 +146,195 @@ yamlfunc: yamlfunc`)
...
+  client3:
+    needsSecret: true
+`)
+
+	yamlOverlayTplData := []byte(`
+#@ load("@ytt:overlay", "overlay")
+#@overlay/match by=overlay.all
+---
+clients:
+  #@overlay/match by=overlay.subset({"client3": {"needsSecret": True}})
```
cppforlife: should this be `by=overlay.subset({"needsSecret": True})` because we are matching on value part of mapitem (ie client3 is key part of mapitem)?

pkg/yttlibrary/overlay/api.go
```diff
@@ -198,6 +206,9 @@ func (b overlayModule) Subset(
 		if _, ok := actualObj.(*yamlmeta.Document); ok {
 			expectedObj = &yamlmeta.Document{Value: expectedObj}
 		}
+		if _, ok := actualObj.(*yamlmeta.MapItem); ok {
```
cppforlife: not sure i get why we need this part. shouldnt we be doing something similar to what we do for ArrayItem above?

pkg/cmd/template/cmd_overlays_test.go
jtigger: No.  Somehow I missed that whole sub-tree in the project.  I'll move these tests over there.

Hmmm... what factors decide which form each of these tests should take?  (e.g. why even have `cmd_overlays_test.go` in the first place?)

pkg/yttlibrary/overlay/api.go
jtigger: I hoped so, too.

However, when the `Tuple` is exchanged for its node from the AST, it comes back as a `Map` wrapping the `MapItem`, not the `MapItem`, itself.  So, to make the comparison, we need to reach into that `Map`.

In the case where there are no items _in_ the map (e.g. `{}`), we need a zero-value.  I read that as a lower-level detail and hence `itemFromMap()`.

pkg/cmd/template/cmd_overlays_test.go
jtigger: Yes.... yes, it should. ... and makes me suspicious of `MapItem` implementation of `pullOutMapValue()` below (see [comments](https://github.com/k14s/ytt/pull/93/files/14e537d3f63ad991af1517e1c19b3eab5188f4a3#r382657540), below).

Digging into this...

pkg/cmd/template/cmd_overlays_test.go
cppforlife: cmd_overlays_test.go tests e2e behaviour of overlays as applied to other documents (and some other plumbing surrounding that). no way to do that within single filetest.

pkg/yttlibrary/overlay/api.go
jtigger: Fascinatingly, by passing in the value of menu items, no reshaping was required within the matcher.

Issue #105, opened 2020-03-11
[data/values] support applying data values to pulled in libraries (via library module)


cppforlife:
it's necessary sometimes to customize data values for libraries that were pulled in. currently `@data/values` allows configuration of the "current" library but not its child libraries. of course one can always use overlays to change any aspect of configuration; however, it's better to provide a way to configure child libraries explicitly. 

let's introduce additional flag on `@data/values` annotation to indicate that values are for a particular library, like so (cannot use `for` kwarg since it's a keyword):

```yaml
#@data/values library="github.com/cloudfoundry/capi-k8s-release"
---
replicaCount: 2
```

in some cases one needs to configure nested library (db-config is included in capi-k8s-release), so we should be able to do something like this:

```yaml
#@data/values library=("github.com/cloudfoundry/capi-k8s-release", "github.com/cloudfoundry/db-config")
#! or 
#@data/values library="github.com/cloudfoundry/capi-k8s-release:github.com/cloudfoundry/db-config"
---
foo: blah
```

above data values should be overlayed before library module provided data values are merged. we should also support a way to overlay values after library module provided data values via `after_library_module=True` kwarg.

when same library is included multiple times, ytt should somehow identify a particular library instance (ideally by an additional given name).

we should also figure out how to pass in data value for a particular library over command line (may be something like that: `ytt -f config --data-value-yaml github.com/cloudfoundry/capi-k8s-release:replicaCount=2 --github.com/cloudfoundry/capi-k8s-release:github.com/cloudfoundry/db-config:foo=blah`)

TBD:
- should we use `:` for separating libraries
- better name for `after_library_module`?

cppforlife:
for multiple copies of the same library, i'm thinking of 
```bash
--data-value-yaml github.com/cloudfoundry/capi-k8s-release~inst1:replicaCount=2 \
--data-value-yaml github.com/cloudfoundry/capi-k8s-release~inst2:replicaCount=2
```

and for data values file...

```yaml
#@data/values library="github.com/cloudfoundry/capi-k8s-release~inst1"
---
...
```

may be some other char to indicate named library?

ewrenn8:
## Updated direction
We will introduce a new annotation `#@library "github.com/cloudfoundry/capi-k8s-release"` which will apply to yaml documents. The yaml document will then only be included if/when the library is evaluated through the library module.

The order of data value overlaying will be: 
1. library data value files 
[...]

PR #110, opened 2020-03-24
Add library and after_library_module kwargs

ewrenn8:
Implements part of #105 (doesnt include support for cmd line, env var, or specific instance)

### DIFF
--- a/pkg/cmd/template/cmd_library_module_test.go
+++ b/pkg/cmd/template/cmd_library_module_test.go
@@ -392,6 +392,243 @@ lib: |
 	runAndCompare(t, filesToProcess, expectedYAMLTplData)
 }
 
+func TestLibraryAttachedDataValues(t *testing.T) {
+	configBytes := []byte(`
+#@ load("@ytt:template", "template")
+#@ load("@ytt:library", "library")
+#@ load("@ytt:data", "data")
+
+curr: #@ data.values.current_val
+--- #@ template.replace(library.get("lib").eval())`)
+
+	dataValueBytes := []byte(`
+#@data/values
+---
+current_val: val1
+
+#@library/name "@lib"
+#@data/values
+---
+lib_val: val2
+
+#@library/name "@lib@github.com/dir/nested-lib"
+#@data/values
+---
+nested_lib_val: nested_val2`)
+
+	libDVBytes := []byte(`
+#@data/values
+---
+lib_val: val1`)
+
+	libConfigBytes := []byte(`
+#@ load("@ytt:template", "template")
+#@ load("@ytt:library", "library")
+#@ load("@ytt:data", "data")
+
+lib_val: #@ data.values.lib_val
+--- #@ template.replace(library.get("github.com/dir/nested-lib").eval())`)
+
+	nestedLibConfigBytes := []byte(`
+#@ load("@ytt:data", "data")
+
+nested_lib_val: #@ data.values.nested_lib_val`)
+
+	nestedLibDVBytes := []byte(`
+#@data/values
+---
+nested_lib_val: nested_val1`)
+
+	expectedYAMLTplData := `curr: val1
+---
+lib_val: val2
+---
+nested_lib_val: nested_val2
+`
+
+	filesToProcess := files.NewSortedFiles([]*files.File{
+		files.MustNewFileFromSource(files.NewBytesSource("values.yml", dataValueBytes)),
+		files.MustNewFileFromSource(files.NewBytesSource("config.yml", configBytes)),
+		files.MustNewFileFromSource(files.NewBytesSource("_ytt_lib/lib/values.yml", libDVBytes)),
+		files.MustNewFileFromSource(files.NewBytesSource("_ytt_lib/lib/config.yml", libConfigBytes)),
+		files.MustNewFileFromSource(files.NewBytesSource("_ytt_lib/lib/_ytt_lib/github.com/dir/nested-lib/values.yml", nestedLibDVBytes)),
+		files.MustNewFileFromSource(files.NewBytesSource("_ytt_lib/lib/_ytt_lib/github.com/dir/nested-lib/config.yml", nestedLibConfigBytes)),
+	})
+
+	runAndCompare(t, filesToProcess, expectedYAMLTplData)
+}
+
+func TestLibraryAfterLibModuleDataValues(t *testing.T) {
+	configBytes := []byte(`
+#@ load("@ytt:template", "template")
+#@ load("@ytt:library", "library")
+#@ load("@ytt:data", "data")
+
+#@ def dv1():
+lib_val1: "foo"
+#@ end
+
+#@ def dv2():
+lib_val2: "bar"
+#@ end
+
+--- #@ template.replace(library.get("lib").with_data_values(dv1()).with_data_values(dv2()).eval())`)
+
+	dataValueBytes := []byte(`
+#@library/name "@lib"
+#@data/values
+---
+lib_val1: val1
+
+#@library/name "@lib"
+#@data/values after_library_module=True
+---
+lib_val2: val2`)
+
+	libDVBytes := []byte(`
+#@data/values
+---
+lib_val1: "unused"
+lib_val2: "unused"`)
+
+	libConfigBytes := []byte(`
+#@ load("@ytt:data", "data")
+
+lib_val1: #@ data.values.lib_val1
+lib_val2: #@ data.values.lib_val2`)
+
+	expectedYAMLTplData := `lib_val1: foo
+lib_val2: val2
+`
+
+	filesToProcess := files.NewSortedFiles([]*files.File{
+		files.MustNewFileFromSource(files.NewBytesSource("values.yml", dataValueBytes)),
+		files.MustNewFileFromSource(files.NewBytesSource("config.yml", configBytes)),
+		files.MustNewFileFromSource(files.NewBytesSource("_ytt_lib/lib/values.yml", libDVBytes)),
+		files.MustNewFileFromSource(files.NewBytesSource("_ytt_lib/lib/config.yml", libConfigBytes)),
+	})
+
+	runAndCompare(t, filesToProcess, expectedYAMLTplData)
+}
+
+func TestTaggedLibraryDataValues(t *testing.T) {
+	configBytes := []byte(`
+#@ load("@ytt:template", "template")
+#@ load("@ytt:library", "library")
+#@ load("@ytt:data", "data")
+
+--- #@ template.replace(library.get("lib", tag="inst1").eval())
+--- #@ template.replace(library.get("lib", tag="inst2").eval())
+--- #@ template.replace(library.get("with-nested-lib", tag="inst1").eval())`)
+
+	dataValueBytes := []byte(`
+#@library/name "@lib~inst1"
+#@data/values
+---
+lib_val1: val1
+
+#@library/name "@lib~inst2"
+#@data/values
+---
+lib_val2: val2
+
+#@library/name "@with-nested-lib~inst1@lib~inst1"
+#@data/values
+---
+nested_lib_val1: new-val1`)
+
+	libDVBytes := []byte(`
+#@data/values
+---
+lib_val1: "library-defined"
+lib_val2: "library-defined"`)
+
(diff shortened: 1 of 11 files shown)


pivotal-issuemaster:
@ewrenn8 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/ytt&pullRequestId=110)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=k14s/ytt&pullRequestId=110) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@ewrenn8 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/ytt&pullRequestId=110)!

pkg/cmd/template/cmd_library_module_test.go
```diff
@@ -392,6 +392,120 @@ lib: |
...
+
+curr: #@ data.values.current_val
+--- #@ template.replace(library.get("lib").eval())`)
+
+	dataValueBytes := []byte(`
+#@data/values
+---
+current_val: val1
+
+#@data/values library="lib"
```
cppforlife: i thought we ended up with '@lib' and '@lib:@child-lib' syntax

pkg/workspace/library_module.go
```diff
@@ -92,12 +104,13 @@ func (l *libraryValue) WithDataValues(thread *starlark.Thread, f *starlark.Built
...
 	dataValues := core.NewStarlarkValue(args.Index(0)).AsGoValue()
 
-	libVal := &libraryValue{l.desc, l.libraryCtx, nil, l.libraryExecutionFactory}
+	libVal := &libraryValue{l.desc, l.libraryCtx, l.dataValuess, l.afterValuess, l.libraryDataValues, l.libraryExecutionFactory}
 	libVal.dataValuess = append([]*yamlmeta.Document{}, l.dataValuess...)
 	libVal.dataValuess = append(libVal.dataValuess, &yamlmeta.Document{
 		Value:    yamlmeta.NewASTFromInterface(dataValues),
 		Position: filepos.NewUnknownPosition(),
 	})
+	libVal.dataValuess = append(libVal.dataValuess, l.afterValuess...)
```
cppforlife: dont think afterValuess should be appended here. they should be done inside eval() since one can write `library.get("foo").with_data_values(vals1).with_data_values(vals2)`

pkg/yttlibrary/data_values.go
```diff
@@ -95,3 +110,37 @@ func (v DataValues) checkNonDocs(val interface{}) error {
 
 	return nil
 }
+
+func processDVAnnotationArgs(annContent string) (map[string]string, error) {
```
cppforlife: i think we should remove knowledge of library kwargs out of this place, which means DataValues.Extract() would continue to return single array of data values. lets move annotation checking (and splitting into two arrays) after templating, before overlaying over here: https://github.com/k14s/ytt/blob/dab244464dbdf009e59e81cabd9b6d9bf3b3747e/pkg/workspace/data_values_pre_processing.go#L53. then you can use template.HasAnnotation etc.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -149,3 +158,26 @@ func (p DataValuesPreProcessing) overlayValuesOverlays(valuesDoc *yamlmeta.Docum
 
 	return result, nil
 }
+
+func isLibDataDoc(doc *yamlmeta.Document) (bool, error) {
```
cppforlife: lets throw this function on DataValuesPreProcessing. if it doesnt need access to anything on it, then you can declare it like so: `func (DataValuesPreProcessing) isLibDataDoc...`

cppforlife: we should also figure out how to consolidate logic in this function with GetValuesForLibraryAndChildren. they both are doing same parsing and arguably should be doing same validation. it could be with a new class eg DataValuesDoc{doc}, that has appropriate methods. we definitely want to ensure that validation of kwards and library value format is done as part of .apply() to throw invalid data/values declaration errors early on (not conditioned on calling to library module .eval())

pkg/cmd/template/data_values_flags.go
```diff
@@ -195,3 +212,28 @@ func (s *DataValuesFlags) buildOverlay(keyPieces []string, value interface{}, de
 
 	return &yamlmeta.Document{Value: resultMap, Position: pos}
 }
+
+func (s DataValuesFlags) CreateDataValues(keyStr string, val interface{}, descStr string) (*workspace.DataValues, error) {
```
cppforlife, line 216: private?

pkg/cmd/template/data_values_flags.go
```diff
@@ -195,3 +212,28 @@ func (s *DataValuesFlags) buildOverlay(keyPieces []string, value interface{}, de
...
+	var lib string
+
+	hasLib := strings.Contains(keyStr, ":")
+	if hasLib {
+		keyParts := strings.Split(keyStr, ":")
```
cppforlife, line 221: what if there are multiple `:`?

pkg/cmd/template/data_values_flags.go
```diff
@@ -195,3 +212,28 @@ func (s *DataValuesFlags) buildOverlay(keyPieces []string, value interface{}, de
...
+		keyStr = keyParts[1]
+	}
+
+	overlayValueDoc, err := workspace.NewDataValues(s.buildOverlay(strings.Split(keyStr, "."), val, descStr))
+	if err != nil {
+		return nil, err
+	}
+
+	if hasLib {
+		err = overlayValueDoc.SetLibAndTag(lib)
```
cppforlife, line 232: would be better to have a constructor that takes doc and lib instead of having public `SetLibAndTag` that we set later == makes DataValues more immutable.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -149,3 +164,101 @@ func (p DataValuesPreProcessing) overlayValuesOverlays(valuesDoc *yamlmeta.Docum
 
 	return result, nil
 }
+
+type DataValues struct {
```
cppforlife, line 168: lets move this to data_values.go

pkg/workspace/data_values_pre_processing.go
```diff
@@ -149,3 +164,101 @@ func (p DataValuesPreProcessing) overlayValuesOverlays(valuesDoc *yamlmeta.Docum
...
+			}
+			piece.Tag = libAndTag[1]
+		}
+		dvd.LibPath = append(dvd.LibPath, piece)
+	}
+
+	return nil
+}
+
+func (dvd *DataValues) HasLib() bool { return dvd.LibPath != nil }
```
cppforlife, line 252: return len(dvd.LibPath) > 0

pkg/workspace/template_loader.go
```diff
@@ -27,7 +28,7 @@ type TemplateLoaderOpts struct {
 	StrictYAML            bool
 }
 
-func NewTemplateLoader(values *yamlmeta.Document, ui files.UI, opts TemplateLoaderOpts,
+func NewTemplateLoader(values *DataValues, libraryValueDocs []*DataValues, ui files.UI, opts TemplateLoaderOpts,
```
cppforlife, line 31: libraryValueDocs -> libraryValuess

pkg/workspace/library_module.go
```diff
@@ -92,12 +177,26 @@ func (l *libraryValue) WithDataValues(thread *starlark.Thread, f *starlark.Built
 
 	dataValues := core.NewStarlarkValue(args.Index(0)).AsGoValue()
 
-	libVal := &libraryValue{l.desc, l.libraryCtx, nil, l.libraryExecutionFactory}
-	libVal.dataValuess = append([]*yamlmeta.Document{}, l.dataValuess...)
-	libVal.dataValuess = append(libVal.dataValuess, &yamlmeta.Document{
+	libVal := &libraryValue{
```
cppforlife, line 180: formatting:

```
libVal := &libraryValue{l.desc, l.tag, l.libraryCtx, l.valuess,
  l.afterModValuess, l.libraryDataValues, l.libraryExecutionFactory}
```

pkg/workspace/library_module.go
```diff
@@ -46,6 +48,26 @@ func (l LibraryModule) Get(thread *starlark.Thread, f *starlark.Builtin,
 		return starlark.None, err
 	}
 
+	var libTag string
```
cppforlife, line 51: may be pull this out into libTag(kwargs) (string, error)

pkg/workspace/library_module.go
```diff
@@ -58,13 +80,76 @@ func (l LibraryModule) Get(thread *starlark.Thread, f *starlark.Builtin,
 
 	libraryCtx := LibraryExecutionContext{Current: foundLib, Root: foundLib}
 
-	return (&libraryValue{libPath, libraryCtx, nil, l.libraryExecutionFactory}).AsStarlarkValue(), nil
+	beforeLibModValuess, afterLibModValuess, childLibValueDocs, err := l.getValuesForLibraryAndChildren(
```
cppforlife, line 83: not Docs

pkg/workspace/template_loader.go
```diff
@@ -201,7 +203,7 @@ func (l *TemplateLoader) EvalYAML(libraryCtx LibraryExecutionContext, file *file
 	l.ui.Debugf("### template\n%s", compiledTemplate.DebugCodeAsString())
 
 	yttLibrary := yttlibrary.NewAPI(compiledTemplate.TplReplaceNode,
-		l.values, l, NewLibraryModule(libraryCtx, l.libraryExecFactory).AsModule())
+		l.values.Doc, l, NewLibraryModule(libraryCtx, l.libraryExecFactory, l.libraryValueDocs).AsModule())
```
cppforlife, line 206: `l.values.Doc` should probably pass in DataValues type

pkg/workspace/library_module.go
```diff
@@ -58,13 +80,76 @@ func (l LibraryModule) Get(thread *starlark.Thread, f *starlark.Builtin,
...
+
+func (l LibraryModule) getValuesForLibraryAndChildren(valueDocs []*DataValues,
+	currentLibName, libTag string) ([]*DataValues, []*DataValues, []*DataValues, error) {
+
+	var currentBeforeModValues []*DataValues
+	var currentAfterModValues []*DataValues
+	var childLibValues []*DataValues
+
+	for _, doc := range valueDocs {
+		copiedDataValues := &DataValues{Doc: doc.Doc, LibPath: doc.LibPath, AfterLibMod: doc.AfterLibMod}
```
cppforlife, line 117: doc.Doc could be deep copied (may be just in case?), but doc.LibPath seems to be somewhat dangerous to pass around like that since it is just pointing to original slice. probably best to make a DataValues.DeepCopy() method and do some work there.

pkg/workspace/library_loader.go
```diff
@@ -40,13 +40,14 @@ func NewLibraryLoader(libraryCtx LibraryExecutionContext,
 	}
 }
 
-func (ll *LibraryLoader) Values(valuesOverlays []*yamlmeta.Document) (*yamlmeta.Document, error) {
-	loader := NewTemplateLoader(&yamlmeta.Document{},
+func (ll *LibraryLoader) Values(valuesOverlays []*DataValues) (*DataValues, []*DataValues, error) {
+	emptyDataDoc, _ := NewDataValues(&yamlmeta.Document{})
```
cppforlife, line 44: should we have NewDataValuesEmpty() constructor?

pkg/workspace/data_values.go
```diff
@@ -0,0 +1,144 @@
...
+	return hasLibAnn, libPath, afterLibMod, nil
+}
+
+func parseLibStr(libStr string) ([]LibPathPiece, error) {
+	if libStr == "" {
+		return nil, fmt.Errorf("library name cannot be empty")
+	}
+
+	if !strings.HasPrefix(libStr, "@") {
+		return nil, fmt.Errorf("library string must being with '@'")
```
cppforlife, line 127: "begin"

Issue #112, opened 2020-03-31
[overlay] overlay module improvements


cppforlife:
TBD:

- [ ] rename `expects` to `expect` kwarg?
- [ ] add `when` kwarg to support conditional actions?
- [ ] traverse value for merge/replace/append/insert (since they may include overlay directives) when short circuiting
- [ ] change default array match matcher to not_op(all) which would result in appending?
- [ ] match child defaults needs to better merge if node specific rules (node specific rules should take precedence)
- [ ] docs: add table to docs describing overlay actions behaviors
- [ ] docs: how to add overlay annotation on a first map item within an array item

notes from the plane: https://gist.github.com/cppforlife/f5eddb3f7dd73239144fb0c148d8f9f2

cppforlife:
when= included in https://github.com/k14s/ytt/releases/tag/v0.28.0

PR #116, opened 2020-04-07
overlay when= kwarg

aaronshurley:
This PR is for the first item in #112.

### DIFF
--- a/pkg/yamltemplate/filetests/ytt-library/overlay/overlay.yml
+++ b/pkg/yamltemplate/filetests/ytt-library/overlay/overlay.yml
@@ -1,4 +1,5 @@
 #@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:template", "template")
 
 #@ def test1_left():
 key1: val1
@@ -82,7 +83,6 @@ key1:
 #@ end
 
 test5: #@ overlay.apply(test5_left(), test5_right())
-
 +++
 
 test1:
--- a/pkg/yamltemplate/filetests/ytt-library/overlay/when.yaml
+++ b/pkg/yamltemplate/filetests/ytt-library/overlay/when.yaml
@@ -1,0 +1,117 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:template", "template")
+
+#@ def test1_when_left():
+map:
+  input_string: val1
+  input_int: val1
+  input_func: val1
+  input_list: val1
+array:
+- val1
+array2:
+- val1
+array3:
+- val1
+#@ end
+
+#@ def true_func(num):
+#@ if num == 1:
+#@   return True
+#@ end
+#@ return False
+#@ end
+
+#@ def test1_when_right():
+map:
+  #@overlay/match when="1+"
+  input_string: string
+  #@overlay/match when="1+"
+  negative_string_case: no-match
+  #@overlay/match when=1
+  input_int: int
+  #@overlay/match when=1
+  negative_int_case: no-match
+  #@overlay/match when=true_func
+  input_func: func
+  #@overlay/match when=true_func
+  negative_func_case: no-match
+  #@overlay/match when=[3,1]
+  input_list: list
+  #@overlay/match when=[2,5]
+  negative_list_case: no-match
+array:
+#@overlay/match by=overlay.all,when="1+"
+- input_string
+#@overlay/match by=overlay.all,when="10+"
+- negative_string_case
+array2:
+#@overlay/match by=overlay.all,when=1
+- input_int
+#@overlay/match by=overlay.all,when=3
+- negative_int_case
+array3:
+#@overlay/match by=overlay.all,when=true_func
+- input_func
+#@overlay/match by=overlay.subset({"blah": "blah"}),when=true_func
+- negative_func_case
+#@ end
+
+test1: #@ overlay.apply(test1_when_left(), test1_when_right())
+
+#@ def test2_when_doc_left():
+---
+input_string: val
+input_int: val
+input_func: val
+input_list: val
+#@ end
+
+#@ def test2_when_doc_right():
+#@overlay/match by=overlay.subset({"input_string": "val"}),when="1+"
+---
+input_string: string
+#@overlay/match by=overlay.subset({"blah": "blah"}),when="1+"
+---
+negative_string_case: no-match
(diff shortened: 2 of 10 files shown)


pkg/yttlibrary/overlay/match_annotation_expects_kwarg.go
```diff
@@ -73,33 +97,33 @@ func (a MatchAnnotationExpectsKwarg) checkValue(val interface{}, matches []*file
 		}
 		if typedResult, ok := result.(starlark.Bool); ok {
 			if !bool(typedResult) {
-				return fmt.Errorf("Expectation of number of matched nodes failed")
+				return MatchAnnotationNumMatchError{"Expectation of number of matched nodes failed"}
```
cppforlife, line 100: i think this should be plain error, not MatchAnnotationNumMatchError since called function returned gibberish.

pkg/yttlibrary/overlay/match_annotation_expects_kwarg.go
```diff
@@ -73,33 +97,33 @@ func (a MatchAnnotationExpectsKwarg) checkValue(val interface{}, matches []*file
...
 			return nil
 		}
-		return fmt.Errorf("Expected keyword argument 'expects' to have a function that returns a boolean")
+		return fmt.Errorf("Expected keyword argument '%s' to have a function that returns a boolean", kw)
 
 	default:
-		return fmt.Errorf("Expected '%s' annotation keyword argument 'expects'"+
-			" to be either int, string or function, but was %T", AnnotationMatch, typedVal)
+		return fmt.Errorf("Expected '%s' annotation keyword argument '%s'"+
+			" to be either int, string or function, but was %T", AnnotationMatch, kw, typedVal)
```
cppforlife, line 108: i typically keep empty space at the end of the line, instead of the beginning.

pkg/yttlibrary/overlay/match_annotation_expects_kwarg.go
```diff
@@ -36,26 +49,37 @@ func (a MatchAnnotationExpectsKwarg) Check(matches []*filepos.Position) error {
...
+		return a.checkValue(*a.expects, MatchAnnotationKwargExpects, matches)
 
 	default:
-		return a.checkValue(starlark.MakeInt(1), matches)
+		return a.checkValue(starlark.MakeInt(1), "", matches)
 	}
 }
 
-func (a MatchAnnotationExpectsKwarg) checkValue(val interface{}, matches []*filepos.Position) error {
+func (a MatchAnnotationExpectsKwarg) checkValue(val interface{}, kw string, matches []*filepos.Position) error {
```
cppforlife, line 82: let's call it kwarg instead of kw.

pkg/yttlibrary/overlay/array.go
```diff
@@ -20,6 +20,9 @@ func (o OverlayOp) mergeArrayItem(
 
 	leftIdxs, err := ann.Indexes(leftArray)
 	if err != nil {
+		if _, ok := err.(MatchAnnotationNumMatchError); ok && ann.expects.when != nil {
```
cppforlife, line 23: given that we repeat `if _, ok := err.(MatchAnnotationNumMatchError); ok && ann.expects.when != nil {` and also dig into privates of ArrayItemMatchAnnotation (eg expects), i think we should probably find a better way to distinguish if we should early exit.

may be set a field on MatchAnnotationNumMatchError about it being "conditional" (ie was used with `when`) and have a little method isMatchAnnotationNumMatchErrorConditional(err)

pkg/yttlibrary/overlay/match_annotation_expects_kwarg.go
```diff
@@ -127,15 +151,15 @@ func (a MatchAnnotationExpectsKwarg) checkString(typedVal starlark.String, match
 	return fmt.Errorf("Expected '%s' to be in format 'i+' where i is an integer", typedValStr)
 }
 
-func (a MatchAnnotationExpectsKwarg) checkList(typedVal *starlark.List, matches []*filepos.Position) error {
+func (a MatchAnnotationExpectsKwarg) checkList(typedVal *starlark.List, kw string, matches []*filepos.Position) error {
```
cppforlife, line 154: kw -> kwarg

pkg/yttlibrary/overlay/match_annotation_expects_kwarg.go
cppforlife: oops. never mind about this.

pkg/yamltemplate/filetests/ytt-library/overlay/overlay.yml
```diff
@@ -83,6 +84,123 @@ key1:
 
 test5: #@ overlay.apply(test5_left(), test5_right())
 
+#@ def test6_when_left():
```
cppforlife, line 87: actually lets move when tests into a separate file.

aaronshurley:
Thanks for the feedback! We made the recommended changes. Please let us know if there's anything else.

pkg/yamltemplate/filetests/ytt-library/overlay/when.yaml
```diff
@@ -0,0 +1,139 @@
...
+#@ if num == 1:
+#@   return True
+#@ end
+#@ return False
+#@ end
+
+#@ def test1_when_right():
+map:
+  #@overlay/match when="1+"
+  #@overlay/replace
```
cppforlife, line 28: we dont need all these replace btw.

aaronshurley: Wonderful. Removed them in the latest commit. Nice catch!

pkg/yamltemplate/filetests/ytt-library/overlay/when.yaml
```diff
@@ -0,0 +1,125 @@
...
+---
+input_string: val
+input_int: val
+input_func: val
+input_list: val
+#@ end
+
+#@ def test2_when_doc_right():
+#@overlay/match by=overlay.subset({"input_string": "val"}),when="1+"
+#@overaly/replace
```
cppforlife, line 72: i imagine these overlay/replaces are not meant to be here either?

aaronshurley: Yup... I'll squash these changes with the previous commit. Sorry about that

PR #118, opened 2020-04-14
Review schema proposal

aaronshurley:
We're creating this PR as a way to provide feedback on the schema proposal. We're hoping this allows us to more collaboratively discuss certain sections of the proposal.


aaronshurley:
We hoped that we could provide comments on lines that we did not change but this wasn't possible. Closing.

PR #128, opened 2020-04-25
Trim whitespace from templated source

jtigger:
Working example of whitespace-trimming source from YAML templates just before parsing.

### DIFF
--- a/pkg/yamltemplate/filetests/code-whitespace.yml
+++ b/pkg/yamltemplate/filetests/code-whitespace.yml
@@ -1,3 +1,4 @@
+#! blocks are determined by `:` and `end`, not whitespace
 #@ if True:
 #@ a = 123
 #@ else:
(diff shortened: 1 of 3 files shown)


pkg/yamltemplate/filetests/code-whitespace.yml
```diff
@@ -12,15 +13,32 @@ test1: #@ a
...
+#@   for b in [20]:
+#@     for c in [300]:
+test3: #@ a+b+c
+#@     end
+#@   end
+#@ end
+
+#! trims space from multi-line strings
 #@ a = "line1\n\
 #@    line2"
```
cppforlife, line 26: that seems a bit weird tbh. there is clearly space between line1 and line2 since it's a multiline string, but it gets removed. it wouldnt be removed from `foo.star`:

```
foo = "blah\
   blah"
```

jtigger: I hear ya. I'm not 100% convinced myself.

Trimming is one way of enforcing that the blocks are delineated by the start (`:`) and end (`end`) tokens (rather than indentation).

Skip trimming if current line is continuation of previous?

jtigger: Trimming leading whitespace on a continuing line is tantamount to deleting whitespace in the middle of a line. That's wrong.

jtigger: For inquiring minds, it's an error in Starlark to follow the line continuation character (i.e. `\`) with anything but a newline: [scan.go#L607](https://github.com/google/starlark-go/blob/be5394c419b6941c14d194a98925a71512f8f979/syntax/scan.go#L607).

PR #134, opened 2020-05-06
Switch to using library ref

ewrenn8:
- A library ref is either the path to lib or the alias
  assigned to the lib during library.get(..., alias=...)
- No uniqueness contraints on alias yet

### DIFF
--- a/pkg/workspace/data_values.go
+++ b/pkg/workspace/data_values.go
@@ -11,43 +11,62 @@ import (
 )
 
 const (
-	dvsLibrarySep    = "@"
-	dvsLibraryTagSep = "~"
+	AnnotationLibraryRef = "library/ref"
+
+	dvsLibrarySep            = "@"
+	dvsLibraryAliasIndicator = "~"
 )
+
+type LibRefPiece struct {
+	Path  string
+	Alias string
+}
+
+func (p LibRefPiece) Matches(lpp LibRefPiece) bool {
+	pathMatch := p.Path == lpp.Path
+	if p.Alias == "" {
+		return pathMatch
+	}
+
+	aliasMatch := p.Alias == lpp.Alias
+	if p.Path == "" {
+		return aliasMatch
+	}
+
+	return aliasMatch && pathMatch
+}
+
+func (p LibRefPiece) AsString() string {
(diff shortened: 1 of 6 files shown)


pkg/workspace/data_values.go
```diff
@@ -11,43 +11,55 @@ import (
...
 )
 
+type LibRefPiece struct {
+	Path  string
+	Alias string
+}
+
+func (p LibRefPiece) Matches(lpp LibRefPiece) bool {
+	if p.Alias != "" {
+		return lpp.Alias == p.Alias
```
cppforlife, line 27: should we check if path here if it's > 0?

pkg/workspace/data_values.go
```diff
@@ -11,43 +11,55 @@ import (
...
+	}
+	return lpp.Path == p.Path && p.Path != ""
+}
+
+func (p LibRefPiece) AsString() string {
+	if p.Path != "" {
+		return p.Path
```
cppforlife, line 34: probably want to return path + alias if alias is non empty?

pkg/workspace/data_values.go
```diff
@@ -138,45 +151,47 @@ func parseDVAnnotations(doc *yamlmeta.Document) (bool, []LibPathPiece, bool, err
...
-		piece := LibPathPiece{LibName: libAndTag[0]}
-		if len(libAndTag) == 2 {
-			if libAndTag[1] == "" {
-				return nil, fmt.Errorf("Expected library tag to not be empty")
+	var result []LibRefPiece
+	for _, refPiece := range strings.Split(libRefStr, dvsLibrarySep)[1:] {
+		pathAndAlias := strings.Split(refPiece, dvsLibraryAliasIndicator)
+		switch l := len(pathAndAlias); {
+		case l > 2:
+			return nil, fmt.Errorf("Expected library alias not to contain more than one '%s'", dvsLibraryAliasIndicator)
```
cppforlife, line 179: it's probably worth just stating expected format here eg `Expected library ref piece 'blah' to be in format 'library', '~alias' or 'library~alias'`, instead of trying to describe it in english.

PR #143, opened 2020-05-15
Implement the ability to load own starlark modules.

wonderix:
I'm using ytt inside my application. I would like to provide own starlark modules that could be loaded inside templates.

I tried 2 variants:
1. Add a `Load` function to `TemplateOptions` of type `func(thread *starlark.Thread, module string) (starlark.StringDict, error)`
2. Add a `Extender` function of type `func(load ModuleLoader) ModuleLoader`

I took the second variant, because it
* requires less code changes
* allows to define a precedence  of own modules and ytt modules.
* allows nesting

### DIFF
--- a/pkg/cmd/template/cmd.go
+++ b/pkg/cmd/template/cmd.go
@@ -21,6 +21,7 @@ type TemplateOptions struct {
 	RegularFilesSourceOpts RegularFilesSourceOpts
 	FileMarksOpts          FileMarksOpts
 	DataValuesFlags        DataValuesFlags
+	Extender               func(load workspace.ModuleLoader) workspace.ModuleLoader
 }
 
 type TemplateInput struct {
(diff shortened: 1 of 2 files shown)


cppforlife:
can you describe a bit more what you using this functionality for?

wonderix:
I'm using this functionality in [shalm](https://github.com/wonderix/shalm).
I'm using [ytt to render templates](https://github.com/wonderix/shalm/blob/ae4c330d561c2c74cbfbc0d968d3fbacb8f38ed7/pkg/shalm/chart_template.go#L167-L250) there. I'm creating my own starlark objects within this project and I would like to pass these objects to ytt.
One solution would be to serialize those objects into YAML and add them as additional file to ytt. But in that case I will lose the ability to also call starlark functions, which are located in these object.
Therefore I would prefer having a way to directly pass these objects into ytt. 
The only way I found was to use these kind of extension.

cppforlife:
i think we would want to provide something more constrained for "side-loading". for example allowing specifically to register for `@shalm` library.

wonderix:
OK

cppforlife:
closing per my last comment above. feel free to PR something that allows registering extension as a library.

PR #148, opened 2020-05-21
Set filename to Location when inserting doc starts

### LINKED ISSUE #147: Error message missing filename when match has no document start
For example:

`template-with-a-document.yml`
```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deploymentB
```

`template-without-a-document.yml`
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: deploymentA
```

`overlay-on-all-documents.yml`
```
#@overlay/match by=lambda i,l,r: True
---
metadata:
  namespace: foos
```

yields:

```
Overlaying (in following order: overlay-on-all-documents.yml): Document on line
overlay-on-all-documents.yml:2: Expected number of matched nodes to be 1, but
was 2 (lines: template-with-a-document.yml:1, 1)
```
[gist in ytt Playground](https://get-ytt.io/#gist:https://gist.github.com/jtigger/bbf6f6816d62173827a04cc8b448a55a)

Note the missing filename in the last match.

jtigger:
Fixes #147

### DIFF
--- a/pkg/yamlmeta/parser.go
+++ b/pkg/yamlmeta/parser.go
@@ -65,6 +65,7 @@ func (p *Parser) ParseBytes(data []byte, associatedName string) (*DocumentSet, e
 	// (note that first doc marker may be several lines down)
 	if !startsWithDocMarker && !docSet.Items[0].Position.IsKnown() {
 		docSet.Items[0].Position = filepos.NewPosition(1)
+		docSet.Items[0].Position.SetFile(associatedName)
 	}
 
 	return docSet, nil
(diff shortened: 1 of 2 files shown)


pkg/cmd/template/cmd_overlays_test.go
```diff
@@ -228,7 +227,7 @@ overlayed: true
 
 	expectedErr := "Overlaying (in following order: overlay.yml): " +
 		"Document on line overlay.yml:4: " +
-		"Expected number of matched nodes to be 1, but was 2 (lines: tpl.yml:2, tpl.yml:6)"
+		"Expected number of matched nodes to be 1, but was 2 (lines: tpl.yml:1, tpl.yml:5)"
```
cppforlife, line 230: instead of changing this test. let's add one to yamlmeta/parser_test.go that explicitly pulls document's filepos and checks AsString() result to include filename.

jtigger: Done. Deliberately did not "pin-down" other constructions of `Position` to ensure the `associatedName` is properly set; seems excessive.

Issue #169, opened 2020-07-13
add regexp.replace 


cppforlife:
currently we have regexp.match (https://github.com/k14s/ytt/blob/develop/docs/lang-ref-ytt.md). jacob barrett (https://kubernetes.slack.com/archives/CH8KCCKA5/p1594261022023900) asked about ability to replace matched parts.

ytt's regexp is backed by Go's regexp (https://golang.org/pkg/regexp/). so we would want to "shim" it in some friendly way. so far i like https://golang.org/pkg/regexp/#Regexp.ReplaceAllStringFunc as a direction.

aaronshurley:
Looks like this work was merged in as part of https://github.com/vmware-tanzu/carvel-ytt/pull/191. Closing.

PR #191, opened 2020-08-22
Implement regexp.replace

jbrunton:
Possible implementation for #169, subject to a few questions/notes about the API:

1. In the original issue, @cppforlife suggested implementing using [Regexp.ReplaceAllStringFunc](https://golang.org/pkg/regexp/#Regexp.ReplaceAllStringFunc), but [Regexp.ReplaceAllLiteralString](https://golang.org/pkg/regexp/#Regexp.ReplaceAllLiteralString) seems simpler for the use cases discussed in the issue and the referenced Slack discussion. Am I missing a use case that would require implementation with ReplaceAllStringFunc?

2. Depending on the answer to question 1 above, there's also a question about whether or not to implement this using [Regexp.ReplaceAllString](https://golang.org/pkg/regexp/#Regexp.ReplaceAllString). This expands `$` signs to submatches (e.g. `$1` to the first submatch), but I wasn't sure how useful this would be, so applying YAGNI for now. (EDIT: As @cppforlife points out below, I mistakenly pushed an implementation using ReplaceAllString, but following subsequent discussion that has been kept.)

3. There was some discussion about whether or not to have separate `replace` and `replaceAll` functions. Given the suggested approach, I'm assuming we are happy to simply replace all (certainly, that tends to be what I most often need when I'm doing a replace), but just want to confirm.

4. There was also discussion about passing flags (e.g. for case-insensitive matches) as a separate param. However, Go uses [RE2 syntax](https://github.com/google/re2/wiki/Syntax) which embeds flags into the regex (example in the tests), so that may not be necessary. One use case for flags might be to select between different underlying implementations (e.g. to choose between ReplaceAllString and ReplaceAllLiteralString depending on whether or not the client wishes to expand submatches), but again, applying YAGNI for now.

### DIFF
--- a/docs/lang-ref-ytt.md
+++ b/docs/lang-ref-ytt.md
@@ -66,8 +66,13 @@ data.list("/data/data.txt") # read file
 load("@ytt:regexp", "regexp")
 
 regexp.match("[a-z]+[0-9]+", "__hello123__") # True
+
 regexp.replace("[a-z]+[0-9]+", "__hello123__", "foo") # __foo__
+regexp.replace("(?i)[a-z]+[0-9]+", "__hello123__HI456__") # __bye__bye__
+regexp.replace("[a-z]+[0-9]+", "__hello123__", lambda s: str(len(s))) # __8__
 ```
+
+See the [RE2 docs](https://github.com/google/re2/wiki/Syntax) for more on the regex syntax supported.
 
 ### url
 
--- a/docs/lang-ref-ytt.md
+++ b/docs/lang-ref-ytt.md
@@ -67,12 +67,15 @@ load("@ytt:regexp", "regexp")
 
 regexp.match("[a-z]+[0-9]+", "__hello123__") # True
 
-regexp.replace("[a-z]+[0-9]+", "__hello123__", "foo") # __foo__
-regexp.replace("(?i)[a-z]+[0-9]+", "__hello123__HI456__") # __bye__bye__
+regexp.replace("[a-z]+[0-9]+", "__hello123__", "foo")                 # __foo__
+regexp.replace("(?i)[a-z]+[0-9]+", "__hello123__HI456__")             # __bye__bye__
+regexp.replace("([a-z]+)[0-9]+", "__hello123__bye123__", "$1")        # __hello__bye__
 regexp.replace("[a-z]+[0-9]+", "__hello123__", lambda s: str(len(s))) # __8__
 ```
 
 See the [RE2 docs](https://github.com/google/re2/wiki/Syntax) for more on the regex syntax supported.
+
+Note that you can pass either a string or a lambda function as the third parameter. When given a string, `$` symbols are expanded, so that `$1` expands to the first submatch. When given a lambda function, the match is directly replaced by the result of the function.
 
 ### url
 
--- a/docs/lang-ref-ytt.md
+++ b/docs/lang-ref-ytt.md
@@ -66,6 +66,7 @@ data.list("/data/data.txt") # read file
 load("@ytt:regexp", "regexp")
 
 regexp.match("[a-z]+[0-9]+", "__hello123__") # True
+regexp.replace("[a-z]+[0-9]+", "__hello123__", "foo") # __foo__
 ```
 
 ### url
--- a/pkg/yamltemplate/filetests/ytt-library/regexp.yml
+++ b/pkg/yamltemplate/filetests/ytt-library/regexp.yml
@@ -5,7 +5,7 @@ test2: #@ regexp.match("^[a-z]+[0-9]+$", "__hello123__")
 test3: #@ regexp.replace("[a-z]+[0-9]+", "__hello123__HI456__", "bye")
 test4: #@ regexp.replace("(?i)[a-z]+[0-9]+", "__hello123__HI456__", "bye")
 test5: #@ regexp.replace("(?i)([a-z]+)[0-9]+", "__hello123__HI456__", "$1")
-test6: #@ regexp.replace("[a-z]+[0-9]+", "__hello123__", lambda a: str(len(a)))
+test6: #@ regexp.replace("(?i)[a-z]+[0-9]+", "__hello123__HI456__", lambda a: str(len(a)))
 
 +++
 
@@ -14,4 +14,4 @@ test2: false
 test3: __bye__HI456__
 test4: __bye__bye__
 test5: __hello__HI__
-test6: __8__
+test6: __8__5__
--- a/pkg/yamltemplate/filetests/ytt-library/regexp.yml
+++ b/pkg/yamltemplate/filetests/ytt-library/regexp.yml
@@ -2,8 +2,12 @@
 
 test1: #@ regexp.match("[a-z]+[0-9]+", "__hello123__")
 test2: #@ regexp.match("^[a-z]+[0-9]+$", "__hello123__")
+test3: #@ regexp.replace("[a-z]+[0-9]+", "__foo123__BAR456__", "baz")
+test4: #@ regexp.replace("(?i)[a-z]+[0-9]+", "__foo123__BAR456__", "baz")
 
 +++
 
 test1: true
 test2: false
+test3: __baz__BAR456__
+test4: __baz__baz__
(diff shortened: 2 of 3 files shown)


gcheadle-vmware:
Hi @jbrunton! Thank you for the contribution!

After reviewing the PR with @pivotaljohn, we think that your decision to use ReplaceAllString is a well formed decision. We couldn't think of a use case that required a function; we could always add ReplaceAllStringFunc in the future. 

Regarding the replace vs replaceAll, your assumption is correct; in the application configuration domain, usually you'd want to replace all.

@cppforlife looks good to me, how does this look to you as the original requestor on this issue?

cppforlife:
> ...so applying YAGNI for now.

just to be clear, it seems that this PR does use ReplaceAllString which automatically expands $s and flags that configure case sensitivity etc. so we are exposing this features.

> Am I missing a use case that would require implementation with ReplaceAllStringFunc?

it's useful when you want to transform matched piece and that transformation might be conditional based on the piece. i m thinking since we are adding replace functionality we should consider adding this as well. im also interested in seeing how api for that would look like e.g. (a) should that be a separate function e.g. `regexp.replace_func` or (b) should we make replacement argument allow to be a function e.g. `regexp.replace("foo", "blah", lambda a: str(len(a)))` (c) something else?

jbrunton:
> it's useful when you want to transform matched piece and that transformation might be conditional based on the piece. i m thinking since we are adding replace functionality we should consider adding this as well. im also interested in seeing how api for that would look like e.g. (a) should that be a separate function e.g. regexp.replace_func or (b) should we make replacement argument allow to be a function e.g. regexp.replace("foo", "blah", lambda a: str(len(a))) (c) something else?

Ah, I was under the impression that Starlark didn't support lambdas, which was why I didn't understand the use case, but now I see the Go implementation _does_ so your suggestion in the original issue now makes a lot more sense to me! I'll have a think about these questions.

>it seems that this PR does use ReplaceAllString which automatically expands $s

Oops! That's unintentional (probably crept in when I was trying out both options for comparison purposes). I'll revert to ReplaceAllLiteralString by default, but if anyone thinks of any use cases where this would be helpful I can keep it (or add the option via flags or another replace function, depending on what seems appropriate).

cppforlife:
> I'll revert to ReplaceAllLiteralString by default

i think keeping it as is should be fine.

pkg/yttlibrary/regexp.go
```diff
@@ -47,3 +48,67 @@ func (b regexpModule) Match(thread *starlark.Thread, f *starlark.Builtin, args s
...
+		return starlark.None, err
+	}
+
+	source, err := core.NewStarlarkValue(args.Index(1)).AsString()
+	if err != nil {
+		return starlark.None, err
+	}
+
+	repl := args.Index(2)
+	switch repl.(type) {
```
jbrunton: This implementation keeps a single `regexp.replace` method which can take either a string or lambda as the third param. To client users, I think this makes sense:

1. In both cases the third param is the "repl" argument, i.e. the thing to replace matches with, just expressed differently.
2. It's hard to mistake a string for a lambda function, so I don't think this will be confusing.
3. Starlark, being based on Python, is a dynamic language, so no need to provide methods with different names when they accept similar arguments with just slightly different signatures.

However:

4. From skimming through the API, I think this may be the first time that the ytt library adds a method which requires a dynamic type check like this? If so: are we ok with that?
5. This check works but I'm not 100% certain there isn't a more canonical way to use the starlark-go API. I'll do some further digging when I have time, but figured I'd push what I have so far to get feedback.

pkg/yttlibrary/regexp.go
```diff
@@ -47,3 +48,67 @@ func (b regexpModule) Match(thread *starlark.Thread, f *starlark.Builtin, args s
...
+		}
+	}()
+
+	newString := re.ReplaceAllStringFunc(source, func(match string) string {
+		args := starlark.Tuple{starlark.String(match)}
+		result, err := starlark.Call(thread, repl, args, []starlark.Tuple{})
+		if err != nil {
+			panic(err)
+		}
+		newString, err := core.NewStarlarkValue(result).AsString()
```
jbrunton: I think this is the correct way to extract the Go string value from the starlark.Value result? We can't simply do `result.String()`, as that adds quotes to the string representation.

pkg/yttlibrary/regexp.go
```diff
@@ -47,3 +48,67 @@ func (b regexpModule) Match(thread *starlark.Thread, f *starlark.Builtin, args s
...
+
+	newString := re.ReplaceAllString(source, replStr)
+
+	return starlark.String(newString), nil
+}
+
+func (b regexpModule) replaceLambda(thread *starlark.Thread, re *regexp.Regexp, source string, repl *starlark.Function) (result starlark.Value, err error) {
+	defer func() {
+		if r, ok := recover().(error); ok {
+			err = r
```
jbrunton: "Catching" the error here to avoid an ugly stack trace if there's an error evaluating the applied lambda function or casting the result to a string.

Example:

```yaml
#@ load("@ytt:regexp", "regexp")
test: #@ regexp.replace("[a-z]+[0-9]+", "__hello123__", lambda s: len(s))
```

```
ytt: Error:
- regexp.replace: expected starlark.String, but was starlark.Int
    in <toplevel>
      test.yml:2 | test: #@ regexp.replace("[a-z]+[0-9]+", "__hello123__", lambda s: len(s))
```

pkg/yttlibrary/regexp.go
cppforlife: > From skimming through the API, I think this may be the first time that the ytt library adds a method which requires a dynamic type check like this? If so: are we ok with that?

we do something similar for some overlay keyword args e.g. https://github.com/k14s/ytt/blob/develop/pkg/yttlibrary/overlay/match_annotation_expects_kwarg.go#L91. using Callable interface is better for type check.

pkg/yttlibrary/regexp.go
cppforlife: hmm, i wonder if there is a better way to propagate errors that happen instead of using panic/recover... what if we have `var lastErr error` and we set it to error that happens (if value isnt string), and then for all subsequent calls to replace function if err is non-nil return empty string. then at the end return lastErr.

docs/lang-ref-ytt.md
```diff
@@ -66,8 +66,14 @@ data.list("/data/data.txt") # read file
 load("@ytt:regexp", "regexp")
 
 regexp.match("[a-z]+[0-9]+", "__hello123__") # True
+
+regexp.replace("[a-z]+[0-9]+", "__hello123__", "foo") # __foo__
+regexp.replace("(?i)[a-z]+[0-9]+", "__hello123__HI456__") # __bye__bye__
+regexp.replace("[a-z]+[0-9]+", "__hello123__", lambda s: str(len(s))) # __8__
```
cppforlife, line 73: should we add an example with $1?

pkg/yamltemplate/filetests/ytt-library/regexp.yml
```diff
@@ -2,8 +2,16 @@
 
 test1: #@ regexp.match("[a-z]+[0-9]+", "__hello123__")
 test2: #@ regexp.match("^[a-z]+[0-9]+$", "__hello123__")
+test3: #@ regexp.replace("[a-z]+[0-9]+", "__hello123__HI456__", "bye")
+test4: #@ regexp.replace("(?i)[a-z]+[0-9]+", "__hello123__HI456__", "bye")
+test5: #@ regexp.replace("(?i)([a-z]+)[0-9]+", "__hello123__HI456__", "$1")
+test6: #@ regexp.replace("[a-z]+[0-9]+", "__hello123__", lambda a: str(len(a)))
```
cppforlife: lets make this test have multiple matches.

jbrunton:
Thanks for the feedback, @cppforlife! I'll take another look at this later this week.

pkg/yttlibrary/regexp.go
jbrunton: ~~That approach works, although my implementation (using named return params in the enclosing function) doesn't quite adopt the separate `lastErr` variable. Lmk what you think.~~

On a second reading, I decided it was clearer with that separate variable after all. Updated.

cppforlife:
looks great!

Issue #208, opened 2020-09-25
enhance structs to support iteration, bracket accessor, ...


cppforlife:
currently we use struct package provided by starlark. it seems that it lacks few things that would make it friendlier to use (especially for case of data.values).

- [ ] support iteration (e.g. `for k in st`)
- [ ] support [] accessor (e.g. st["foo"] in addition to st.foo)
- [ ] `in` check (e.g. `"foo" in st`)
- [ ] anything else that we have done for yaml fragments https://github.com/k14s/ytt/blob/develop/pkg/yamltemplate/starlark_fragment.go?

jtigger:
Started working on this.
Thought I was done when `struct.encode()` was working as described in this issue.
However, while writing up corresponding docs, discovered that `struct.make()` is an entirely different code-path.
Will address that bit too.

stealthybox:
Just noting here that iterating over maps for this use-case does appear to work properly with ytt 0.31.0.
```yaml
#@data/values
---
env:
    KUBE_SECRET: my-registry-token  #! does not yet exist -- will be created in the same Namespace
    SYNC_PERIOD: 3600  #! tokens expire; refresh faster than that
```
```yaml
---
#! ... abbreviated

        env:
        #@ for k in data.values.env:
        - name: #@ str(k)
          value: #@ str(data.values.env[k])
        #@ end
```
This is common with config for k8s Deployments.

I'm not knowledgeable whether maps and structs are similar or related in Starlark, and there may be other technicalities that mean they are not fully iterable yet.

Either way, this wasn't clear from these examples:
- https://carvel.dev/ytt/#example:example-for
- https://carvel.dev/ytt/#example:example-datatypes

PR #210, opened 2020-10-05
Make struct an iterable mapping and a sequence

jtigger:
Addresses #208 .

### DIFF
--- a/docs/lang-ref-structs.md
+++ b/docs/lang-ref-structs.md
@@ -52,8 +52,8 @@ Attributes can be referenced:
 
 Referencing an attribute that is not defined on the `struct` results in an error:
 ```yaml
-db_url: data.values.bd_conn              # <== struct has no .bd_conn field or method
-db_host: data.values["db_conn"]["hast"]  # <== key "hast" not in struct
+db_url: #@ data.values.bd_conn              # <== struct has no .bd_conn field or method
+db_host: #@ data.values["db_conn"]["hast"]  # <== key "hast" not in struct
 ```
 
 __
@@ -88,7 +88,7 @@ The following built-ins can be useful with `struct` values:
     # `value` is a struct that _might_ have a field, "additional_ports"
     
     def ports(value):
-      if(hasattr(value, "additional_ports")):
+      if hasattr(value, "additional_ports"):
       ... 
     end
     ```
--- a/docs/lang-ref-structs.md
+++ b/docs/lang-ref-structs.md
@@ -1,0 +1,103 @@
+### Structs
+
+Structs are well-defined data objects, comprised of key/value pairs known as "attributes". They are a way to store and refer to data of a known structure.
+
+The most commonly used `struct` is `data.values`, supplied by the [`@ytt:data`](ytt-data-values.md) module.
+
+For example, a data values defined by:
+
+```yaml
+#@data/values
+---
+db_conn:
+  host: acme.example.com
+```
+
+is automatically processed into a `struct` (named `values`): the keys in the `@data/values` file are defined one-for-one as attributes on the `struct`.
+ 
+Those attribues can be referenced by name:
+
+```yaml
+#@ load("@ytt:data", "data")
+---
+persistence:
+  db_url: #@ data.values.db_conn.host
+```
+
+_([ytt @data/values](ytt-data-values.md) describes this process in more detail.)_
+
+__
+
+Described below:
+- [Attributes](#attributes) — how to refer to a datum within a `struct`
+- [Build-in Functions](#built-in-functions) — functions that support using a `struct`
+- [Creating `struct`s](#creating-structs) — support for creating your own `struct`s
+
+---
+
+#### Attributes
+
+Attributes are a key/value pair, where the key is a `string` and the value can be of any type.
+
+Attributes can be referenced:
+- by name (using "[dot notation](https://github.com/google/starlark-go/blob/master/doc/spec.md#dot-expressions)")
+    ```yaml
+    db_url: #@ data.values.db_conn.host
+    ```
+- by string (using "[index notation](https://github.com/google/starlark-go/blob/master/doc/spec.md#index-expressions)") _(as of v0.31.0)_:
+    ```yaml
+    db_url: #@ data.values["db_conn"]["host"]
+    ```
+  useful when the name of the attribute is not known statically.
+
+Referencing an attribute that is not defined on the `struct` results in an error:
+```yaml
+db_url: data.values.bd_conn              # <== struct has no .bd_conn field or method
+db_host: data.values["db_conn"]["hast"]  # <== key "hast" not in struct
+```
+
+__
+
+#### Built-in Functions
+
+The following built-ins can be useful with `struct` values:
+
+- [`dir()`](https://github.com/google/starlark-go/blob/master/doc/spec.md#dir) enumerates all its attributes.
+    ```
+    load("@ytt:struct", "struct")
+    
+    foo = struct.encode({"a": 1, "b": 2, "c": 3})
+    keys = dir(foo)  # <== ["a", "b", "c"]
+    ```
+  
+- [`getattr()`](https://github.com/google/starlark-go/blob/master/doc/spec.md#getattr) can be used to select an attribute.
+    ```yaml
+    #@ for vpc_config in getattr(getattr(data.values.accounts, foundation_name), vpc_name):
+       ...
+    #@ end
+    ```
+  - as of v0.31.0, `struct`s support [index notation](https://github.com/google/starlark-go/blob/master/doc/spec.md#index-expressions) behaving identically, more succinctly/readably:
+    ```yaml
+    #@ for vpc_config in data.values.accounts[foundation_name][vpc_name]:
+       ...
+    #@ end
+    ```
+
+- [`hasattr()`](https://github.com/google/starlark-go/blob/master/doc/spec.md#hasattr) reports whether a value has a specific attribute
+    ```python
+    # `value` is a struct that _might_ have a field, "additional_ports"
+    
+    def ports(value):
+      if(hasattr(value, "additional_ports")):
+      ... 
+    end
+    ```
+
+__
+
+#### Creating `struct`s
+
+`struct` instances can be made using the `@ytt:struct` module.
+
+See [ytt Library: struct module](lang-ref-ytt-struct.md) for details.
+
--- a/docs/lang-ref-structs.md
+++ b/docs/lang-ref-structs.md
@@ -40,7 +40,7 @@ Described below:
 Attributes are a key/value pair, where the key is a `string` and the value can be of any type.
 
 Attributes can be referenced:
-- by name (using "[dot notation](https://github.com/google/starlark-go/blob/master/doc/spec.md#dot-expressions)")
+- by field (using "[dot notation](https://github.com/google/starlark-go/blob/master/doc/spec.md#dot-expressions)")
     ```yaml
     db_url: #@ data.values.db_conn.host
     ```
--- a/docs/lang-ref-ytt-struct.md
+++ b/docs/lang-ref-ytt-struct.md
@@ -67,8 +67,9 @@ _Example 1: Data structure from a dictionary_
 load("@ytt:struct", "struct")
 
 d = struct.encode({"a": [1,2,3,{"c":456}], "b": "str"})
-d.a     # ==> [1, 2, 3, c: (struct...)]
-d.a[3].c   # ==> 456
+d.a        # <== [1, 2, 3, c: (struct...)]
+d.a[3].c   # <== 456
+bar["b"]   # <== "str"
 ```
 
 __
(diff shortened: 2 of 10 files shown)


jtigger:
As I was writing up corresponding docs, I discovered that `struct.make()` yields `starlarkstruct.Struct` instances (provided by Starlark) where `struct.encode()` yields instances of `core.StarlarkStruct`.

Looks like we'll need to port over the `Make()` functionality (i.e. building struct from kwargs).

Also, docs update needs to be finished before merging.

Creating draft PR, in the meantime, to signal someone has started work on this issue.

jtigger:
Now all structs created in `ytt` are of the custom type: `core.StarlarkStruct`.

docs/lang-ref-structs.md
```diff
@@ -0,0 +1,103 @@
+### Structs
+
+Structs are well-defined data objects, comprised of key/value pairs known as "attributes". They are a way to store and refer to data of a known structure.
```
cppforlife, line 3: should add version availability for this feature.

docs/lang-ref-ytt-struct.md
```diff
@@ -0,0 +1,221 @@
...
+**Examples:** 
+
+_Example 1: Binding a function value_
+
+```python
+load("@ytt:struct", "struct")
+
+conn = struct.make(hostname="svc.example.com", default_port=1022, protocol="https")
+
+def _url(this, port=None):
```
cppforlife: self might be more pythonic.

pkg/yttlibrary/struct.go
```diff
@@ -29,6 +29,25 @@ var (
 
 type structModule struct{}
 
+func (b structModule) Make(_ *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if len(args) > 0 {
+		return nil, fmt.Errorf("struct: unexpected positional arguments; structs are made from keyword arguments only")
```
cppforlife: core.ErrWrapper will add struct.make to the beginning of error, so `struct: ` is redundant.

cppforlife: also instead of ; lets use (...)

pkg/template/core/starlark_struct.go
```diff
@@ -40,3 +45,50 @@ func (s *StarlarkStruct) AttrNames() []string {
 	})
 	return keys
 }
+
+func (s *StarlarkStruct) Get(key starlark.Value) (val starlark.Value, found bool, err error) {
+	if attr, ok := NewStarlarkValue(key).AsGoValue().(string); ok {
+		val, err = s.Attr(attr)
+		return val, val != nil, err
+	} else {
```
cppforlife: could kill else and just do return.

pkg/template/core/starlark_struct.go
```diff
@@ -40,3 +45,50 @@ func (s *StarlarkStruct) AttrNames() []string {
 	})
 	return keys
 }
+
+func (s *StarlarkStruct) Get(key starlark.Value) (val starlark.Value, found bool, err error) {
+	if attr, ok := NewStarlarkValue(key).AsGoValue().(string); ok {
+		val, err = s.Attr(attr)
+		return val, val != nil, err
```
cppforlife: seems like it would better to call s.data.Get for an explicit found boolean instead of checking on val!=nil which is "non-coded" contract from Attr() method.

pkg/template/core/starlark_struct.go
```diff
@@ -40,3 +45,50 @@ func (s *StarlarkStruct) AttrNames() []string {
...
+	} else {
+		return nil, false, fmt.Errorf("expected key `%s` to be a string but is a %T", key, key)
+	}
+}
+
+func (s *StarlarkStruct) Iterate() starlark.Iterator {
+	return &StarlarkStructIterator{
```
cppforlife, line 62: one liner?

pkg/template/core/starlark_struct.go
```diff
@@ -5,17 +5,22 @@ package core
...
 
-var _ starlark.Value = &StarlarkStruct{}
-var _ starlark.HasAttrs = &StarlarkStruct{}
+var _ starlark.Value = (*StarlarkStruct)(nil)
+var _ starlark.HasAttrs = (*StarlarkStruct)(nil)
+var _ starlark.IterableMapping = (*StarlarkStruct)(nil)
+var _ starlark.Sequence = (*StarlarkStruct)(nil)
 
 func (s *StarlarkStruct) String() string        { return "struct(...)" }
 func (s *StarlarkStruct) Type() string          { return "struct" }
```
cppforlife, line 26: it's interesting that Truth() method always returns true in starlark's official struct: https://github.com/google/starlark-go/blob/master/starlarkstruct/struct.go#L159. what other difference are note worthy? hashable?

cppforlife: another interesting difference is return of NoSuchAttrError (https://github.com/google/starlark-go/blob/master/starlarkstruct/struct.go#L231) which makes runtime augment err with a hint of misspellings.

pkg/yttlibrary/struct.go
jtigger: > also instead of ; lets use (...)

What's the motivation for `...` ?

Argument for `;` is that it's grammatically correct. Semicolon ties two sentences together into a single notion. Ellipsis indicate that intermediate text has been elided. In conversation, it conveys spoken word, trailing off... Granted colloquially, `...` has become interface compatible with `;` ... AFAIK.

docs/lang-ref-structs.md
jtigger: Do you mean retroactively? I _did_ note in places where the index notation is referenced that this is available "as of v0.31.0"

pkg/template/core/starlark_struct.go
jtigger: > it's interesting that Truth() method always returns true in Starlark's official struct

seriously.  Yeah, and _with a comment_ that reads, `// even when empty`. Gonna chew on what "truth" means, here. 🤔

jtigger: With respect to _additional_ functionality...

I _am_ interested in pulling some of that over (love that hinting bit, yes).

That said, given that we were basically using the `ytt` version of `struct` for most of our own cases, I feel less compelled to tackle that in _this_ PR.

I _did_ review `yamlfragment` as #208 requested.

Seems "good enough" for now.

pkg/template/core/starlark_struct.go
jtigger: eh. When I tried it, it was _less_ readable.

pkg/template/core/starlark_struct.go
jtigger: Truth:

So... all these data structures are truthy if only if they are non-empty.
```
func (d *Dict) Truth() Bool { return d.Len() > 0 }
func (l *List) Truth() Bool { return l.Len() > 0 }
func (s *Set) Truth() Bool { return s.Len() > 0 }
func (t Tuple) Truth() Bool { return len(t) > 0 }
func (r rangeValue) Truth() Bool { return r.len > 0 }
```

why would `struct` be different? 🤔

pkg/yttlibrary/struct.go
cppforlife: what i meant was `unexpected positional arguments (structs are made from keyword arguments only)`

docs/lang-ref-structs.md
```diff
@@ -0,0 +1,103 @@
...
+    ```
+- by string (using "[index notation](https://github.com/google/starlark-go/blob/master/doc/spec.md#index-expressions)") _(as of v0.31.0)_:
+    ```yaml
+    db_url: #@ data.values["db_conn"]["host"]
+    ```
+  useful when the name of the attribute is not known statically.
+
+Referencing an attribute that is not defined on the `struct` results in an error:
+```yaml
+db_url: data.values.bd_conn              # <== struct has no .bd_conn field or method
```
cppforlife: missing #@

docs/lang-ref-structs.md
```diff
@@ -0,0 +1,103 @@
...
+       ...
+    #@ end
+    ```
+
+- [`hasattr()`](https://github.com/google/starlark-go/blob/master/doc/spec.md#hasattr) reports whether a value has a specific attribute
+    ```python
+    # `value` is a struct that _might_ have a field, "additional_ports"
+    
+    def ports(value):
+      if(hasattr(value, "additional_ports")):
```
cppforlife: probably better to drop ( around hasattr.

docs/lang-ref-structs.md
cppforlife: nm, didnt see index notation has a version notice at the end of the line.

pkg/template/core/starlark_struct.go
cppforlife: not sure why they have done that. im ok keeping it inline especially that yamlfrag also does len() > 0.

pkg/yttlibrary/struct.go
jtigger: Oh, geez. Yeah, I see the established pattern for using the parenthetical. 👍

docs/lang-ref-structs.md
jtigger: Good catch; thanks.

pkg/template/core/starlark_struct.go
jtigger: And, in fact, `Get()` should return `starlark.None`, not `nil` when no error occurred but the item was not found.

docs/lang-ref-structs.md
```diff
@@ -0,0 +1,103 @@
...
+#### Attributes
+
+Attributes are a key/value pair, where the key is a `string` and the value can be of any type.
+
+Attributes can be referenced:
+- by name (using "[dot notation](https://github.com/google/starlark-go/blob/master/doc/spec.md#dot-expressions)")
+    ```yaml
+    db_url: #@ data.values.db_conn.host
+    ```
+- by string (using "[index notation](https://github.com/google/starlark-go/blob/master/doc/spec.md#index-expressions)") _(as of v0.31.0)_:
```
StevenLocke, line 47: This is also accessing by name, it's changing the notation not the access method.

docs/lang-ref-ytt-struct.md
```diff
@@ -0,0 +1,221 @@
...
+
+_Example 1: _
+
+```python
+load("@ytt:struct", "struct")
+
+foo = struct.encode({"a": [1,2,3,{"c":456}], "b": "str"})
+bar = struct.decode(foo)
+
+bar["a"]  # <== [1, 2, 3, {"c": 456}]
```
StevenLocke, line 43: I would also add
bar["b"]  # <== "str"

docs/ytt-data-values.md
```diff
@@ -1,27 +1,32 @@
 ## ytt @data/values
 
-- [Defining data values](#defining-data-values)
+The standard way to externalize configuration values is to declare them as "Data Values"
+and then reference those values in templates.
+
+- [Defining and Using Data Values](#declaring-and-using-data-values)
```
StevenLocke: `[Declaring and Using Data Values](#declaring-and-using-data-values)`

docs/ytt-data-values.md
```diff
@@ -1,27 +1,32 @@
...
   nested: val2
-key3:
+key3: val3
 key4:
 ```
 
-Subsequently these values can be accessed via `@ytt:data` library:
+`ytt` processes all Data Values files prior to rendering templates.
+
+Templates access those processed values via the `@ytt:data` module:
```
StevenLocke: Templates "can" access...
Templates "are able to" access...

docs/ytt-data-values.md
```diff
@@ -37,14 +42,25 @@ Resulting in
...
 fourth: null
 ```
 
-We typically recommend to use snake case (e.g. `some_key.nested_one`) for naming data values.
-
-Note that if data value contains a `-` (dash) in its name, you will have to use `getattr` function like so `getattr(data.values.key2, "nested-with-dash")` to access its value. This is to avoid parsing ambiguity with substraction operation.
-
+Note:
+- Data Values keys must be strings.
+- use snake-case for keys (e.g. `db_conn.secure`) — keys that contain a `-` (dash)
```
StevenLocke: I'd still mark this as a recommendation. As in,
We recommend using snake-case...,
to avoid the idea that this is a requirement.

StevenLocke (commented):
Incomplete review, only looked at the markdown files.

docs/lang-ref-structs.md
pivotaljohn: nice catch. Will use "field" there.

docs/ytt-data-values.md
pivotaljohn: Thank you! using the tersest form.

docs/ytt-data-values.md
pivotaljohn: whoops! Thank you!

docs/ytt-data-values.md
pivotaljohn: Fair enough. Even more relevant with the support for index notation.

PR #214, opened 2020-10-07
Add experimental Schemas feature

gcheadle-vmware:
- Schema features are disabled by default, to use provide the '--enable-experiment-schema' flag.
- Documents marked with the Schema annotation are not included in the output.


pkg/yamlmeta/schema.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+	}
+}
+
+// TODO
+/// set the types
+// check the types
+//
+
+//Document{
```
cppforlife: commented out lines probably shouldnt be include here.

pkg/yamlmeta/parser.go
```diff
@@ -123,6 +123,14 @@ func (p *Parser) parseBytes(data []byte, lineCorrection int) (*DocumentSet, erro
 	return docSet, nil
 }
 
+//type Type struct {
```
cppforlife: commented out lines probably shouldnt be include here.

pkg/workspace/library_loader.go
```diff
@@ -44,9 +44,36 @@ func NewLibraryLoader(libraryCtx LibraryExecutionContext,
...
+
+	if len(schemaFiles) > 0 {
+		libraryCtx := LibraryExecutionContext{Current: schemaFiles[0].Library, Root: NewRootLibrary(nil)}
+
+		_, resultDocSet, err := loader.EvalYAML(libraryCtx, schemaFiles[0].File)
+		if err != nil {
+			return nil, err
+		}
+
+		tplOpts := yamltemplate.MetasOpts{IgnoreUnknown: true}
```
cppforlife: shouldnt be hard coded.

pkg/workspace/template_loader.go
```diff
@@ -185,7 +186,17 @@ func (l *TemplateLoader) EvalYAML(libraryCtx LibraryExecutionContext, file *file
 	if err != nil {
 		return nil, nil, err
 	}
-
+	if _, ok := l.schema.(yamlmeta.AnySchema); !ok {
```
cppforlife, line 189: checking schema in EvalYAML seems to be too low level if we are trying to just check data values schema. should we just check it in Values()?

pkg/workspace/doc_extractor.go
```diff
@@ -12,31 +12,36 @@ import (
 )
 
 const (
-	AnnotationDataValues structmeta.AnnotationName = "data/values"
+	AnnotationDataValues  structmeta.AnnotationName = "data/values"
```
cppforlife, line 15: dont think these two constants should live here. probably better move them back into yttlibrary.

pkg/cmd/template/cmd.go
```diff
@@ -123,7 +127,29 @@ func (o *TemplateOptions) RunWithFiles(in TemplateInput, ui cmdcore.PlainUI) Tem
...
-	values, libraryValues, err := libraryLoader.Values(valuesOverlays)
+	schemaDocs, err := libraryLoader.Schemas()
+	if err != nil {
+		return TemplateOutput{Err: err}
+	}
+	var schema yamlmeta.Schema = yamlmeta.AnySchema{}
+	if len(schemaDocs) > 0 {
+		if o.SchemaEnabled {
+			schema = yamlmeta.NewDocumentSchema(schemaDocs[0])
+			if err != nil {
```
cppforlife, line 138: which err != nil is this checking?

pkg/workspace/template_loader.go
pivotaljohn: Great catch: there's no need to place this code in paths that we don't intend the use for some time. It's just risk with no (obvious) reward.

pkg/workspace/doc_extractor.go
pivotaljohn: Could you share more about what leads you to that idea?

I notice that...
- currently, those constants are only used in the `workspace` package
- each group of annotations live within their own package
- data values currently lives in `workspace` package

pkg/cmd/template/cmd.go
pivotaljohn: Thank you. For the immediate, teammates are encouraged to do some local linting (e.g. GoLand comes with such inspects, automatically). Will discuss use of a linter in an upcoming Tech Forum.

Issue #14, opened 2019-04-21
[lang] Support for standard YAML comments


dpb587-pivotal:
Is there a reason standard YAML comments are not supported? Since `ytt` seems to try and sort of follow YAML syntax, I expected it to be ignored by `ytt`.

```
$ cat > example.yml <<EOF
earlier: former
# later: latter # TODO someday
EOF

$ ytt template -f example.yml
Error: Unknown comment syntax at line example.yml:2: ' later: latter # TODO someday': Unknown metadata format (use '#@' or '#!')
```

I could imagine `#something` without a following space to be something `ytt` doesn't support due to theoretical ambiguity, but `#spacesomething` seems reasonable for `ytt` to ignore as standard YAML convention and non-`ytt` directive-style.

I understand `#!` is an alternative, but that's just more of a migration/difference and don't think I understand why it should be needed.

(meant to submit as @dpb587 from personal side-project learnings)

cppforlife:
our current thinking was to be more cautious and disallow plain comments (`# blah`) since users may forget to add `@` after `#` (`#@ blah`) and their code would not run. i've definitely been a bit annoyed by it as well. i've added `file-mark` to be able to specify that particular file is plain (non-templated yaml)

```
ytt t -R -f . --file-mark file1.yml:type=yaml-plain
```

we may at some point relax this restriction, and just ignore `#`. @nimakaviani wdyt about that?

nimakaviani:
yeah even though it is annoying to have the interpreter yell at you because of using standard yaml comments, that appeared to be the safer choice compared to rendering some of the templating code useless.

As for the `file-mark` flag, i like the idea however not totally fond of the syntax. A less verbose option with dedicated flags may be slightly easier from a ui perspective.

cppforlife:
@dpb587-pivotal wdyt?

dpb587:
Thanks, I understand the concerns. From my perspective, `ytt` seems additive to YAML so I'd hoped standard YAML would not need additional work. In terms of users forgetting, I'd expect them to quickly notice if their `ytt` operations are not taking effect as they're updating and testing, so with `ytt` requiring `#!`/`#@`-type directives, I'd assume it to be fairly discoverable. Alternatively a less verbose form of `--debug` could emit warnings for syntax hints if `ytt` were to accept standard comments.

Regarding `file-mark`, I hadn't realized its use yet, so it's good to know. Although I, too, think it's a little difficult to realize or confusing.

[...]

PR #226, opened 2020-11-01
Report "unrecognized meta" errors in users' terms and hint at remedy

jtigger:
Discussed in detail in k14s/ytt#14

### DIFF
--- a/pkg/cmd/template/cmd_library_module_test.go
+++ b/pkg/cmd/template/cmd_library_module_test.go
@@ -1120,7 +1120,7 @@ lib_int: #@ data.values.int`)
 		t.Fatalf("Expected RunWithFiles to error but it did not")
 	}
 
-	if !strings.Contains(out.Err.Error(), "' ignored!': Unknown metadata format") {
+	if !strings.Contains(out.Err.Error(), "'# ignored!': Unrecognized comment type") {
 		t.Fatalf("Expected ref error but got '%s'", out.Err)
 	}
 }
--- a/pkg/cmd/template/cmd_test.go
+++ b/pkg/cmd/template/cmd_test.go
@@ -507,7 +507,7 @@ yamlfunc: yamlfunc`)
 		t.Fatalf("Expected RunWithFiles to fail")
 	}
 
-	if out.Err.Error() != "Unknown comment syntax at line tpl.yml:2: ' plain YAML comment': Unknown metadata format (use '#@' or '#!')" {
+	if !strings.Contains(out.Err.Error(), "Non-ytt comment at line tpl.yml:2: '# plain YAML comment': Unrecognized comment type (expected '#@' or '#!')") {
 		t.Fatalf("Expected RunWithFiles to fail with error, but was '%s'", out.Err.Error())
 	}
 }
--- a/pkg/structmeta/meta.go
+++ b/pkg/structmeta/meta.go
@@ -65,7 +65,7 @@ func NewMetaFromString(data string, opts MetaOpts) (Meta, error) {
 				Content: data,
 			}}
 		} else {
-			return Meta{}, fmt.Errorf("Unknown metadata format (use '#@' or '#!')")
+			return Meta{}, fmt.Errorf("Unrecognized comment type (expected '#@' or '#!')")
 		}
 	}
 
--- a/pkg/yamltemplate/metas.go
+++ b/pkg/yamltemplate/metas.go
@@ -43,7 +43,8 @@ func NewStructMetaFromMeta(meta *yamlmeta.Meta, opts MetasOpts) (structmeta.Meta
 	structMeta, err := structmeta.NewMetaFromString(meta.Data, structmeta.MetaOpts{IgnoreUnknown: opts.IgnoreUnknown})
 	if err != nil {
 		return structmeta.Meta{}, fmt.Errorf(
-			"Unknown comment syntax at %s: '%s': %s", meta.Position.AsString(), meta.Data, err)
+			"Non-ytt comment at %s: '#%s': %s. (hint: if this is plain YAML — not a template — consider `--file-mark '<filename>:type=yaml-plain'`)",
+			meta.Position.AsString(), meta.Data, err)
 	}
 	return structMeta, nil
 }


jtigger:
Intended as a (reasonable?) compromise:
- tool remains "safe" by default
- not onerous — processing YAML that the user did not author (which often includes "regular" comments) is eased with a single character.

cppforlife:
one reservation i have is that by making flag single char it's impossible to tell what it does without referring to documentation (i can already see this line https://github.com/cloudfoundry/cf-for-k8s/blob/develop/build/postgres/build.sh#L9 would not be very friendly to understand). ... and because it's only a single char it's not going to raise any "flags" whether user should really use it or not.

so far i'm -1 on this shorthand addition, though improvements to error messages are good.

jtigger:
> one reservation i have is that by making flag single char it's impossible to tell what it does without referring to documentation (i can already see this line https://github.com/cloudfoundry/cf-for-k8s/blob/develop/build/postgres/build.sh#L9 would not be very friendly to understand). ... and because it's only a single char it's not going to raise any "flags" whether user should really use it or not.

This PR does not seek to remove the long-form of the flag. That can still be used to properly document what's happening, such as the case you've linked.

> though improvements to error messages are good.
👍 

> so far i'm -1 on this shorthand addition

For the collection of use-cases centered around the command-line development, I find it disproportionately irritating to have to type such a long flag name just to configure the tool to do the thing I expected it to do in the first place.

What information would you need to see in order to be convinced that _requiring_ users (especially Configuration Authors) to include `--ignore-unknown-comments` in every single `ytt` invocation is an undesirable experience?

jessehu:
Thanks @jtigger & @cppforlife. We can add shorthand name for long name which is useful for devs who do not like typing :D  One minor suggestion is changing the PR title (or description) to mention something like ‘provide descriptive error message for using standard yaml comment’ because the error message improvement in this PR is also valuable.

cppforlife:
> This PR does not seek to remove the long-form of the flag. 

yup, understood.

> That can still be used to properly document what's happening, such as the case you've linked.

that's i guess my current thinking: when i see `ytt -i -f ... -f ...` (or even `ytt -if ...`) in a script it's very easy to "skip over" it because its so tiny vs a full descriptive flag name. because the barrier for using -i is so low it would be way to easier to unintentionally copy it into other places that do not require it. full flag name imho makes users pause about its purpose and whether it's needed. (it's also aligned with `library.get("...", ignore_unknown_comments=True)` usage.

> I find it disproportionately irritating to have to type such a long flag name just to configure the tool to do the thing I expected it to do in the first place ... snip... include --ignore-unknown-comments in every single ytt 

addition of -i signals to me that something has changed and there are suddenly more workflows where it would be appropriate. what are those use cases? can we address them in better, safer, ways?

jtigger:
> that's i guess my current thinking: when i see ytt -i -f ... -f ... (or even ytt -if ...) in a script it's very easy to "skip over" it because its so tiny vs a full descriptive flag name. because the barrier for using -i is so low it would be way to easier to unintentionally copy it into other places that do not require it. full flag name imho makes users pause about its purpose and whether it's needed. (it's also aligned with library.get("...", ignore_unknown_comments=True) usage.

I wonder...
- how probable is this risk? 
- what is exactly the impact of that risk?
- is that cost worth the friction to adoption that this brings?

> addition of -i signals to me that something has changed and there are suddenly more workflows where it would be appropriate. what are those use cases? can we address them in better, safer, ways?

Is there only _one_ interpretation of this signal? I mean, everything is relative. 😉 

---

**Use Case** (with example)
The use case I have personally experienced, matches that of what others have shared with me, and is an example of what was brought up in #14 is this:

As a Configuration Author, I want to integrate Kubernetes manifests that others have written into my platform.

For example, I want to install Contour on my cluster
```console
$ kapp deploy -a contour -f https://projectcontour.io/quickstart/contour.yaml
```

But, wait, I  forgot to mention that I installed CF for K8s, first. Today, that product is using Istio and establishes an ingress on 80 and 443, already. My deploy fails because those ports are already taken. That's fine, I'll just write an overlay to slide Contour over to an unused port.

```
$ ytt -f https://projectcontour.io/quickstart/contour.yaml
ytt: Error: Unknown metadata format
...
```

This invokes confusion:
- given that one of YAML's selling points is that authors can comment, it is unsurprising that this YAML _does_ contain comments.
- given that `ytt` presents itself as a YAML templating tool, is quite surprising that this valid YAML is rejected by the tool.

Today, that's it. With the comment portion of the PR, we at least mention the presence of the flag. Even then, let's game it out...

```console
$ ytt --ignore-unknown-comments -f https://projectcontour.io/quickstart/contour.yaml
(success)
```

But now, I write my overlay and I want to check my work: take the diff of before and after. In order to get a good diff output (formatting, etc), I need to invoke `ytt` twice: once on the unedited pile of manifests, then again on the overlayed set.

```console
$ diff <(ytt --ignore-unknown-comments -f https://projectcontour.io/quickstart/contour.yaml) <(ytt --ignore-unknown-comments -f https://projectcontour.io/quickstart/contour.yaml -f reassign-ingress-ports.yml) 
```

This experience broadly (albeit not universally) invokes frustration: the tool is making the productive thing an irritating thing.

The probably of this use case is high. It's hard to imagine any non-trivial use of Kubernetes without using someone else's YAML.
The impact of this use case is not insignificant. The intensity of responses to it range from mild surprise through to "this tool makes me feel dumb." This does not sound like the user satisfaction that Fred Brooks pointed out as the true goal of software products.

cppforlife:
👍, lets work with this example...

> ytt --ignore-unknown-comments -f https://projectcontour.io/quickstart/contour.yaml -f reassign-ingress-ports.yml

by adding --ignore-unknown-comments we are applying to all files including reassign-ingress-ports.yml.

existing options:

- keep it as is (isnt pleasant for use case above)
  - not convenient for examples above
- use `--file-mark contour.yaml:type=plain-yaml`
  - very verbose, too powerful for day to day use

other options:

- drop --ignore-unknown-comments feature entirely
- add -i as an alias to --ignore-unknown-comments
- introduce a new way (hopefully a lot shorter) to mark file as plain-yaml
  - `-f` today supports this form `-f name=src-path`. may be there is something there?
- introduce smarter detection in ytt for plain vs templated file
  - e.g. if we see one `#@` or `#!` comment then consider file as template, otherwise plain yaml
    - there is a chance that somebody uses `#@` or `#!` in plain yaml, but ytt would try to execute that anyway so it's a question how do you "downgrade" template to plain-yaml
- ...what else can we do?

if we just extend this example a little further, when multiple people are involved:

```
wget -O- https://projectcontour.io/quickstart/contour.yaml > contour.yaml
wget -O- https://something.yaml > something.yaml
# add bunch of other ytt templates
ytt -i -f .
# push out to shared repo
```

now somebody else who wants to get final config _needs_ to know that -i is necessary instead of just executing `ytt -f .` as one would hope. of course we can solve this via _ytt_lib use and library.get("...", ignore_unknown_comments=True) but arguably may be we should have something more explicit.

jessehu:
IMHO, +1 for `drop --ignore-unknown-comments feature entirely` or defaulting `--ignore-unknown-comments` to true, as discusssed in [`Support for standard YAML comments · Issue #14 · k14s/ytt`](https://github.com/k14s/ytt/issues/14):
1) In our use cases, --ignore-unknown-comments is used almost in every ytt commands, because we get the yamls from upstream open source projects or generated from helm charts. 
2) It follows the ytt principle that `ytt (pronounced spelled out) is a templating tool that understands YAML structure`.

jessehu:
We can also rename --ignore-unknown-comments to --use-strict-comments with default value false. Similar to "use strict" in Javascript.

jessehu:
LGTM 👍

Issue #223, opened 2020-10-23
Can I add a map to an array of maps?


amalagaura:
I am not sure if this is supported and I am trying different overlay syntax. I am not really able to modify the source yamls so trying to do this with YTT.

```
---
id: 1
type: book
awards:
  books:
  - id: 1
    title: International Botev
    reviewers:
      - id: 2
        name: PersonB
  - id: 2
    title: Dayton Literary Peace Prize
    reviewers:
      - id: 3
        name: PersonC
        
#! How do I add a map to an array of maps?
```

```
#@ load("@ytt:overlay", "overlay")

#@overlay/match by=overlay.all
---
awards:
  books:
    #@overlay/match by=overlay.all, expects="1+"
    #@overlay/match missing_ok=True
    reviewers:
      #@overlay/append
      - id: 1
        name: PersonA
 ```

So trying to add a map to the reviewers array in both books..

I created a StackOverflow ticket for it.
https://stackoverflow.com/questions/64504795/how-do-i-add-a-map-to-an-array-of-maps-in-ytt

gcheadle-vmware:
Hi @amalagaura!

Overlays are a great way to patch yaml and can certainly work for your case. 

> So trying to add a map to the reviewers array in both books.

You are really close and only two moves away, all you have to do is remove the `overlay/match missing_ok=True`, and make `reviewers:` an array item. Right now, having multiple `overlay/match` annotations cancel out all but the last one, so your  `#@overlay/match missing_ok=True` annotation inadvertently canceled out the one above.

If you want the behavior of an ` overlay/match missing_ok=True`, so that if a book does not have a reviewer key, one is added, you can achieve this with overlays as well. Here is a gist demonstrating that behavior: https://get-ytt.io/#gist:https://gist.github.com/gcheadle-vmware/a6243ee73fa5cc139dba870690eb15c5

amalagaura:
Awesome thanks!

amalagaura:
I think this example would be great in your overlay playground. Can I just add this in there with a PR or are you planning to do that?

cari-lynn:
Hey @amalagaura! We welcome PR's for documentation improvements. So thank you. 

I took a look at the overlay playground and it looks like there is an [append an array item example](https://get-ytt.io/#example:example-append-array-item). The main difference being that the item we are matching in this case is an array item, and in the playground example it's a map.
In the [match by key example](https://get-ytt.io/#example:example-match-by-key) there is an example on how to match on an array item.

[...]

PR #236, opened 2020-11-14
Add example of adding a map to an array in a complex object

amalagaura:
Follow up from issue #223 to document the example/answer given

### DIFF
--- a/examples/playground/overlays/example-append-map-to-array/_data.yml
+++ b/examples/playground/overlays/example-append-map-to-array/_data.yml
@@ -1,0 +1,17 @@
+---
+id: 1
+type: book
+awards:
+  books:
+    - id: 1
+      title: International Botev
+      reviewers:
+        - id: 2
+          name: PersonB
+    - id: 2
+      title: Dayton Literary Peace Prize
+      reviewers:
+        - id: 3
+          name: PersonC
+    - id: 3
+      title: Example Book
--- a/examples/playground/overlays/example-append-map-to-array/_data.yml
+++ b/examples/playground/overlays/example-append-map-to-array/_data.yml
@@ -3,15 +3,10 @@ id: 1
 type: book
 awards:
   books:
-    - id: 1
+    - id: book-botev
       title: International Botev
       reviewers:
-        - id: 2
-          name: PersonB
-    - id: 2
-      title: Dayton Literary Peace Prize
-      reviewers:
-        - id: 3
-          name: PersonC
-    - id: 3
+        - id: person-a
+          name: Person A
+    - id: example-book
       title: Example Book
--- a/examples/playground/overlays/example-append-map-to-array/append-map-to-array.yml
+++ b/examples/playground/overlays/example-append-map-to-array/append-map-to-array.yml
@@ -11,5 +11,5 @@ awards:
       #@overlay/match missing_ok=True
       reviewers:
         #@overlay/append
-        - id: 1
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@amalagaura Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/ytt&pullRequestId=236)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=k14s/ytt&pullRequestId=236) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@amalagaura Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=k14s/ytt&pullRequestId=236)!

amalagaura:
@pivotal-dean I added the overlay example from #223. I hope it is sufficient.

examples/playground/overlays/example-append-map-to-array/_data.yml
```diff
@@ -0,0 +1,17 @@
...
+    - id: 1
+      title: International Botev
+      reviewers:
+        - id: 2
+          name: PersonB
+    - id: 2
+      title: Dayton Literary Peace Prize
+      reviewers:
+        - id: 3
+          name: PersonC
```
StevenLocke: Could you explain what we're illustrating with the first two `books` array items? I see them as similar and am wondering if it would make sense to have only one.

pkg/website/generated.go.txt
```diff
@@ -75,6 +75,7 @@ exampleSets = [
           {"id": "example-edit-map-value"},
           {"id": "example-remove-map-item"},
           {"id": "example-rename-key-in-map"},
+          {"id": "example-append-map-to-array"},
```
StevenLocke, line 78: This example may follow logically after `Append array item`, how would you feel about this example being number 11?

examples/playground/overlays/example-append-map-to-array/append-map-to-array.yml
```diff
@@ -0,0 +1,15 @@
...
+#@overlay/match by=overlay.all
+---
+awards:
+  books:
+    #@overlay/match by=overlay.all, expects="1+"
+    -
+      #@overlay/match missing_ok=True
+      reviewers:
+        #@overlay/append
+        - id: 1
```
StevenLocke: When this example runs, it outputs the following as the first array item under the `books` key:
```
- id: 1
    title: International Botev
    reviewers:
    - id: 2
      name: PersonB
    - id: 1
      name: Person A
```

Because we're `@overlay/append`ing it puts the second id at the end of the `reviewers` array (after the first id). This may be confusing for people who see the ids out of order. Instead, could we replace the `id` field with something non-order dependent?

amalagaura:
@StevenLocke Thanks for the feedback. I liked your suggestions and simplified accordingly, hope it is better. I left an id field, maybe it is not needed and we can simplify the map further, I am not sure.. I put the order of the example according to what @pivotal-dean suggested in issue #223. I think after `Append array item` does make sense.

StevenLocke (approved):
Great updates! This looks good to me.

@pivotal-dean 
Final word is yours on the order :)

cari-lynn:
Looks great. Thanks for the addition @amalagaura !

PR #237, opened 2020-11-16
Experimental Schemas: Add support for default values fill in

joaopapereira:
When a value is missing in the data file but present in the schema, ytt
will use this value as the default and it will populate the map
structure with it.

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -119,6 +119,7 @@ github.com/spf13/pflag v1.0.3/go.mod h1:DYY7MBk1bdzusC3SYhjObp+wFpr4gzcvqqNjLnIn
 github.com/spf13/pflag v1.0.5 h1:iy+VFUOCP1a+8yFto/drg2CJ5u0yRoB7fZw3DKv/JXA=
 github.com/spf13/pflag v1.0.5/go.mod h1:McXfInJRrz4CZXVZOBLb0bTZqETkiAhM9Iw0y3An2Bg=
 github.com/spf13/viper v1.4.0/go.mod h1:PTJ7Z/lr49W6bUbkmS1V3by4uWynFiR9p7+dSq/yZzE=
+github.com/stretchr/objx v0.1.1 h1:2vfRuCMp5sSVIDSqO8oNnWJq7mPa6KVP3iPIwFBuy8A=
 github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/testify v1.2.2 h1:bSDNvY7ZPG5RlJ8otE/7V6gMiyenm9RtJ7IUVIAoJ1w=
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
--- a/go.sum
+++ b/go.sum
@@ -119,7 +119,6 @@ github.com/spf13/pflag v1.0.3/go.mod h1:DYY7MBk1bdzusC3SYhjObp+wFpr4gzcvqqNjLnIn
 github.com/spf13/pflag v1.0.5 h1:iy+VFUOCP1a+8yFto/drg2CJ5u0yRoB7fZw3DKv/JXA=
 github.com/spf13/pflag v1.0.5/go.mod h1:McXfInJRrz4CZXVZOBLb0bTZqETkiAhM9Iw0y3An2Bg=
 github.com/spf13/viper v1.4.0/go.mod h1:PTJ7Z/lr49W6bUbkmS1V3by4uWynFiR9p7+dSq/yZzE=
-github.com/stretchr/objx v0.1.1 h1:2vfRuCMp5sSVIDSqO8oNnWJq7mPa6KVP3iPIwFBuy8A=
 github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/testify v1.2.2 h1:bSDNvY7ZPG5RlJ8otE/7V6gMiyenm9RtJ7IUVIAoJ1w=
 github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
--- a/go.sum
+++ b/go.sum
@@ -66,8 +66,8 @@ github.com/inconshreveable/mousetrap v1.0.0 h1:Z8tu5sraLXCXIcARxBp/8cbvlwVa7Z1NH
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
 github.com/jonboulle/clockwork v0.1.0/go.mod h1:Ii8DK3G1RaLaWxj9trq07+26W01tbo22gdxWY5EU2bo=
 github.com/julienschmidt/httprouter v1.2.0/go.mod h1:SYymIcj16QtmaHHD7aYtjjsJG7VTCxuUUipMqKk8s4w=
-github.com/k14s/difflib v0.0.0-20200108171459-b101e55e0592 h1:Qt52IIUgcCU2lHWbmBTTzp5IDJ0UM4Kc5BKmGmhTZJ0=
-github.com/k14s/difflib v0.0.0-20200108171459-b101e55e0592/go.mod h1:B0xN2MiNBGWOWi9CcfAo9LBI8IU4J1utlbOIJCsmKr4=
+github.com/k14s/difflib v0.0.0-20201117154628-0c031775bf57 h1:CwBRArr+BWBopnUJhDjJw86rPL/jGbEjfHWKzTasSqE=
+github.com/k14s/difflib v0.0.0-20201117154628-0c031775bf57/go.mod h1:B0xN2MiNBGWOWi9CcfAo9LBI8IU4J1utlbOIJCsmKr4=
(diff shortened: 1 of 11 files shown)


pkg/cmd/template/schema_test.go
```diff
@@ -63,6 +63,81 @@ rendered: true`
...
+rendered: true
+db:
+  hostname: #@ data.values.db_conn.hostname
+  port: #@ data.values.db_conn.port
+  username: #@ data.values.db_conn.username
+  password: #@ data.values.db_conn.password
+  tls_only: #@ data.values.db_conn.tls_only
+  metadata: #@ data.values.db_conn.metadata
+  other_key: #@ data.values.db_conn.other_key.missing_key.something
+  inner_key: #@ data.values.db_conn.other_key.inner_key
```
pivotaljohn: Consider taking advantage of the fact that `structs` can be rendered, as is:

```yaml
#@ load("@ytt:data", "data")
---
data.values: #@ data.values
```

pkg/cmd/template/schema_test.go
```diff
@@ -63,6 +63,81 @@ rendered: true`
...
+---
+db_conn:
+  username: sa
+  password: changeme
+  tls_only: true
+  other_key:
+    missing_key:
+      something: yeet
+`
+	templateYAML := `#@ load("@ytt:data", "data")
```
pivotaljohn: Consider shaping the test fixture so that it strikes a balance between using real-worldish data values (which is present, here), and having elements that are telling of what's happening.

Otherwise, the reader must digest the whole example, line-by-line and build up a mental mapping of what's being tested ==> "too" much cognitive load.

For example:
- `missing_key` is not missing, but `inner_key` _is_ missing (from the data values).
- there appear to be multiple places where defaults are overridden in the data values. consider to limit this to — at most — one per type? (both `username` and `password` are defaulted strings that are overridden.

pkg/cmd/template/schema_test.go
```diff
@@ -63,6 +63,81 @@ rendered: true`
...
+  password: changeme
+  tls_only: true
+  metadata:
+    run: jobName
+  other_key: yeet
+  inner_key: some string
+`
+	if string(out.Files[0].Bytes()) != expected {
+		t.Fatalf("Expected output to only include template YAML, but got: %s", out.Files[0].Bytes())
+	}
```
pivotaljohn: An emerging pattern is to use our text diffing library to make it easier to spot the differences between expected and actual.

Let's pull that in here.

pkg/yamlmeta/type.go
```diff
@@ -103,18 +105,48 @@ func (t *MapType) AssignTypeTo(typeable Typeable) (chk TypeCheck) {
...
+}
+
+func contains(haystack []interface{}, needle interface{}) bool {
+	for _, key := range haystack {
+		if key == needle {
+			return true
+		}
+	}
+	return false
+}
```
pivotaljohn, line 152: This made me smile. Thanks for the tongue-in-cheek idiom! ❤️ 

Does `Key`s are technically any type. Does `==` do a deep equals? 🤔

pkg/yamlmeta/type.go
```diff
@@ -48,8 +48,10 @@ type MapType struct {
 	Items []*MapItemType
 }
 type MapItemType struct {
-	Key       interface{} // usually a string
-	ValueType Type
+	Key          interface{} // usually a string
+	ValueType    Type
+	DefaultValue interface{}
```
pivotaljohn, line 53: `ValueType` has access to its source which contains the original node (and therefore the default value)...

Was that in the mix when we considered introducing `DefaultValue`?

go.sum
```diff
@@ -119,6 +119,7 @@ github.com/spf13/pflag v1.0.3/go.mod h1:DYY7MBk1bdzusC3SYhjObp+wFpr4gzcvqqNjLnIn
 github.com/spf13/pflag v1.0.5 h1:iy+VFUOCP1a+8yFto/drg2CJ5u0yRoB7fZw3DKv/JXA=
 github.com/spf13/pflag v1.0.5/go.mod h1:McXfInJRrz4CZXVZOBLb0bTZqETkiAhM9Iw0y3An2Bg=
 github.com/spf13/viper v1.4.0/go.mod h1:PTJ7Z/lr49W6bUbkmS1V3by4uWynFiR9p7+dSq/yZzE=
+github.com/stretchr/objx v0.1.1 h1:2vfRuCMp5sSVIDSqO8oNnWJq7mPa6KVP3iPIwFBuy8A=
```
pivotaljohn: I'm still learning about Go modules... what does it mean for the version number not to change but the checksum does not?

pkg/yamlmeta/type.go
```diff
@@ -103,18 +105,48 @@ func (t *MapType) AssignTypeTo(typeable Typeable) (chk TypeCheck) {
...
+		}
+
+		val := &MapItem{
+			Key:      item.Key,
+			Value:    item.DefaultValue,
+			Position: item.Position,
+		}
+		childCheck := item.AssignTypeTo(val)
+		chk.Violations = append(chk.Violations, childCheck.Violations...)
+		mapNode.AddValue(val) // TODO: How might we surface this error?
```
pivotaljohn: Open `TODO:`.

I think this is an unexpected situation. Are there cases where applying the default value _could_ result in a **type-check** violation (as opposed to a validation violation)?

Given that considering, what should we do, here?

pivotaljohn (changes requested):
Looks really good, @joaopapereira.

Some nip and tuck still suggestions, below.

pkg/yamlmeta/type.go
cppforlife: if we are not propagating error up then we should panic. definitely do not want to swallow.

pkg/yamlmeta/type.go
joaopapereira: Yes indeed `ValueType` is of the type `Type` (lol) but we would have to cast it to the correct type in line 133 of this file in order to retrieve the value from this new type.
To simplify that logic, we decided to add another field that contains the raw value we want to attach

pkg/yamlmeta/type.go
joaopapereira: To get an error we would need to pass a value to the method `AddValue` that is not of the type `*MapItem`. This cannot happen because we create the variable `val` 7 lines above this invocation.
Nevertheless, we can always add here a panic, not really a fan of panic in mid execution 😢  that was why I added a TODO so we can discuss the best option

pkg/yamlmeta/type.go
joaopapereira: Key should be a string (or perhaps a number???) which should render this `==` to be a simple string comparison.
This method is already used in other parts of the code to ensure we are handly the value from the correct key. (type.go:112 as an example)

pkg/cmd/template/schema_test.go
joaopapereira: done

pkg/cmd/template/schema_test.go
joaopapereira: done

pkg/cmd/template/schema_test.go
joaopapereira: done

pkg/yamlmeta/type.go
joaopapereira: If you know of a better way let me know and I am happy to update it 😃

pkg/yamlmeta/type.go
cppforlife: > Nevertheless, we can always add here a panic, not really a fan of panic in mid execution

`panic("Unreachable")` is a very common pattern to assert on internal constraints.

joaopapereira: done

go.sum
joaopapereira: I added a package to do a check, and that updated the `go.sum`. I removed this line and everything is working now

pkg/yamlmeta/type.go
```diff
@@ -103,18 +105,52 @@ func (t *MapType) AssignTypeTo(typeable Typeable) (chk TypeCheck) {
...
+		val := &MapItem{
+			Key:      item.Key,
+			Value:    item.DefaultValue,
+			Position: item.Position,
+		}
+		childCheck := item.AssignTypeTo(val)
+		chk.Violations = append(chk.Violations, childCheck.Violations...)
+		err := mapNode.AddValue(val)
+		if err != nil {
+			panic("Unreachable")
```
cppforlife: i would go with something like this actually `fmt.Sprintf("Internal inconsistency: adding map item: %s", err)`

pkg/cmd/template/schema_test.go
```diff
@@ -361,3 +429,27 @@ rendered: true`
 		t.Fatalf("Expected an error about schema enabled but no schema provided, but got: %v", out.Err.Error())
 	}
 }
+
+func diffText(left, right string) string {
```
cppforlife: seems like this got copied from elsewhere. lets pull that out into difflib.

pkg/yamlmeta/type.go
joaopapereira: fixed

pkg/cmd/template/schema_test.go
joaopapereira: extracted and updated the library

pkg/yamlmeta/type.go
```diff
@@ -137,7 +137,7 @@ func (t *MapType) applySchemaDefaults(foundKeys []interface{}, chk TypeCheck, ma
 		chk.Violations = append(chk.Violations, childCheck.Violations...)
 		err := mapNode.AddValue(val)
 		if err != nil {
-			panic("Unreachable")
+			panic(fmt.Sprintf("Internal inconsistency: adding map item: %s", err))
```
pivotaljohn, line 140: Improved! 👍🏼 

In general, when I'm faced with reporting particularly rare/unexpected situations, I try to report as much context as I can.  So, I scan around the function that is going to emit this panic and look for values that are essential to the failed operation and make sure they are either included in the Error message, or report it myself.  For example, is `val` in the message of `err`? And what's the state of `mapNode`?

I think this is okay, for now, but would be worth the little extra effort to make sure those surprising moments come more wrapped in a bow.

PR #243, opened 2020-12-01
Switch Back to Original Example ID Condition Check

danielhelfand:
Using strings.Contains() can cause examples that have names that are substrings of other example names to load wrong examples. An example of this is `example-load` and `examples-load-data-values`. Reverting back to old condition to check for example ID.


joaopapereira (approved):
Looks good

PR #244, opened 2020-12-01
Schema nullable refactoring

joaopapereira:
In this PR some refactoring was done to try to address comments on the commit original commit for the nullable schema annotation implementation.

https://github.com/vmware-tanzu/carvel-ytt/commit/66c26bad47cc0d30f4224c205f054a145793d6b0#r44427722
https://github.com/vmware-tanzu/carvel-ytt/commit/66c26bad47cc0d30f4224c205f054a145793d6b0#r44427606

In broad strokes, the annotation concept is now part of the schema types and instead of relying on magic strings we now use `structmeta.AnnotationName` and implemented a method that checks if a type can be nullable.

### DIFF
--- a/pkg/cmd/template/cmd.go
+++ b/pkg/cmd/template/cmd.go
@@ -9,6 +9,7 @@ import (
 
 	cmdcore "github.com/k14s/ytt/pkg/cmd/core"
 	"github.com/k14s/ytt/pkg/files"
+	"github.com/k14s/ytt/pkg/schema"
 	"github.com/k14s/ytt/pkg/workspace"
 	"github.com/k14s/ytt/pkg/yamlmeta"
 	"github.com/spf13/cobra"
@@ -131,10 +132,10 @@ func (o *TemplateOptions) RunWithFiles(in TemplateInput, ui cmdcore.PlainUI) Tem
 	if err != nil {
 		return TemplateOutput{Err: err}
 	}
-	var schema yamlmeta.Schema = &yamlmeta.AnySchema{}
+	var schemaVar workspace.Schema = &schema.AnySchema{}
 	if len(schemaDocs) > 0 {
 		if o.SchemaEnabled {
-			schema, err = yamlmeta.NewDocumentSchema(schemaDocs[0])
+			schemaVar, err = schema.NewDocumentSchema(schemaDocs[0])
 			if err != nil {
 				return TemplateOutput{Err: err}
 			}
@@ -151,7 +152,7 @@ func (o *TemplateOptions) RunWithFiles(in TemplateInput, ui cmdcore.PlainUI) Tem
 		}
 	}
 
-	values, libraryValues, err := libraryLoader.Values(valuesOverlays, schema)
+	values, libraryValues, err := libraryLoader.Values(valuesOverlays, schemaVar)
 	if err != nil {
 		return TemplateOutput{Err: err}
 	}
--- a/pkg/workspace/data_values_pre_processing.go
+++ b/pkg/workspace/data_values_pre_processing.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/k14s/starlark-go/starlark"
 	"github.com/k14s/ytt/pkg/filepos"
+	"github.com/k14s/ytt/pkg/schema"
 	"github.com/k14s/ytt/pkg/yamlmeta"
 	"github.com/k14s/ytt/pkg/yamltemplate"
 	yttoverlay "github.com/k14s/ytt/pkg/yttlibrary/overlay"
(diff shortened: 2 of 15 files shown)


pkg/yamlmeta/type.go
```diff
@@ -45,22 +46,26 @@ type DocumentType struct {
 	ValueType Type // typically one of: MapType, ArrayType, ScalarType
 }
 type MapType struct {
-	Items []*MapItemType
+	Items       []*MapItemType
+	annotations template.NodeAnnotations
```
cppforlife: im -1 on this. this violates layering that's been setup. yamlmeta package should NOT know anything about templating.

joaopapereira: Can you explain what do you mean by layering?

cppforlife: we would like to maintain boundaries between various packages so that there is a clean layering going on in the "cake" (project). bottom layers of the cake should not depend on higher layers of the cake.

one way to "test this" is to imagine if to use particular package in isolation. eg my other project needs to parse yamls and extract metadata (comments). it doesnt need any concept of templating and i want to interpret comments in my own way (eg `# foo` is something special, everything else is plain comment).

the way that ytt cake is set up is something like this:

- yamlmeta - standalone package that knows how to parse YAML and produce YAML AST.
  - no dependency
- template - standalone generic starlark templating with a high level instructions interface for building structures
  - no dependency
- texttemplate/yamltemplate - language specific (YAML, text, may be JSON in future, etc)
  - yamltemplate -> depends on -> yamlmeta, and template
  - texttemplate -> depends on -> template

(more pkgs descriptions here: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/docs/dev.md#source-code-structure).

cppforlife:
> Stop yamlmeta package depending on the template package

yamlmeta pkg still continues to depend on template pkg (https://github.com/vmware-tanzu/carvel-ytt/pull/244/commits/6a351599276b8420823d021e37298452a8668bbc#diff-ae30da157e6a7f9c338ef9aeccf6e4f5ae90e3a460619a3a6d87987597ba4153R9)

StevenLocke:
Hmmm, that's a good point. That commit was inaccurately named.

More accurately, the template dependency is now only in the schema type construction, it was removed from the more general type.go. Perhaps the schema code doesn't belong in the yamlmeta package to keep this layering, but the "types" needs to be able to check those annotations at some point. It has to be able to read "#@schema/nullable" somewhere.

pkg/cmd/template/cmd.go
```diff
@@ -5,6 +5,7 @@ package template
 
 import (
 	"fmt"
+	"github.com/k14s/ytt/pkg/schema"
```
joaopapereira: Please move the import to the other group

pkg/workspace/data_values_pre_processing.go
```diff
@@ -5,6 +5,7 @@ package workspace
 
 import (
 	"fmt"
+	"github.com/k14s/ytt/pkg/schema"
```
joaopapereira: Please move the import to the other group

pkg/workspace/library_loader.go
```diff
@@ -5,6 +5,7 @@ package workspace
 
 import (
 	"fmt"
+	"github.com/k14s/ytt/pkg/schema"
```
joaopapereira: Please move the import to the other group

pkg/workspace/library_module.go
```diff
@@ -5,6 +5,7 @@ package workspace
 
 import (
 	"fmt"
+	"github.com/k14s/ytt/pkg/schema"
```
joaopapereira: Please move the import to the other group

pkg/workspace/template_loader.go
```diff
@@ -5,6 +5,7 @@ package workspace
 
 import (
 	"fmt"
+	"github.com/k14s/ytt/pkg/schema"
```
joaopapereira: Please move the import to the other group

pkg/yamlmeta/type.go
```diff
@@ -6,6 +6,11 @@ package yamlmeta
 import (
 	"fmt"
 	"github.com/k14s/ytt/pkg/filepos"
+	"github.com/k14s/ytt/pkg/structmeta"
+)
+
+const (
+	AnnotationSchemaNullable structmeta.AnnotationName = "schema/nullable"
```
joaopapereira: Feels like this is a schema annotation. Shouldn't this be in the schema package?

joaopapereira (commented):
I added some comments, specially around import ordering.

Since we created a schema package now I am thinking that maybe we should not have the Types inside yamlmeta, since the Types make sense associated with a schema, not by themselves.

Not sure what implications this would have for yamlmeta and the dependencies between packages.

Issue #245, opened 2020-12-01
Fix Broken Link for Schema Error


danielhelfand:
Running `ytt` with `--enable-experiment-schema` with an invalid schema definition, such as below, returns the error underneath the invalid schema.

Invalid schema:
```
#@schema/definition data_values=True
---
vpc:
  name: ""
  subnet_ids:
  - 0
  #@schema/nullable
  subnet_config:
  - id: 0
    mask: "255.255.0.0"
    private: true
```

Error:
```
ytt: Error: Schema experiment flag was enabled but no schema document was provided. 
(See this propsal for details on how to include a schema document: https://github.com/k14s/design-docs/blob/develop/ytt/001-schemas/README.md#defining-a-schema-document)
```

When clicking on the link in the error to get more information, a 404 is returned.

gcheadle-vmware:
Link removed in merged pull request #246

PR #246, opened 2020-12-02
Remove Broken Schema Definition Link

danielhelfand:
Until docs are in more mature state, we should not provide a link with this error. Currently docs live in the design docs repo (https://github.com/k14s/design-docs/blob/develop/ytt/), but they are not up to date with the implementation and may cause confusion. A link should be added back when schema docs are better defined/have a more permanent location.


danielhelfand:
Closes #246

danielhelfand:
Closes #245

PR #250, opened 2020-12-04
add docs site installation instructions to readme

DanielJonesEB:
Hey folks, I didn't stumble across the docs site which mentions the homebrew tap until I'd done it the manual way. Hopefully this will make `ytt` marginally easier to install for those who don't find the full-form docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,44 @@
 
-* Install: Grab prebuilt binaries from the [Releases page](https://github.com/vmware-tanzu/carvel-ytt/releases)
+* Install: [See below](#installation), or grab prebuilt binaries from the [Releases page](https://github.com/vmware-tanzu/carvel-ytt/releases)
 * Play: Jump right in by trying out the [online playground](https://get-ytt.io/#playground)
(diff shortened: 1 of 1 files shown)


cppforlife:
it would be good to figure how to address visibility of available installation methods but also not have to maintain repeated copy for every single carvel repository. do you think having a dedicated doc page (on carvel.dev [0]) that we could link from line 2 in README.md would be sufficient?

[0] we will be launching new copy of carvel.dev site that has docs for each tool (and probably would also have generic docs for all tools).

gcheadle-vmware:
> do you think having a dedicated doc page (on carvel.dev) that we could link from line 2 in README.md would be sufficient?

+1. I agree that linking the installation guide would make this information easier to maintain.

aaronshurley:
@DanielJonesEB We have this information on our website (https://carvel.dev/), do you think that this is easy enough to discover or would having installation instructions within this repo's README be beneficial, as well? We'd like to strike the right balance between discoverability and maintainability 🤔

cari-lynn:
Hey @DanielJonesEB, thanks for the time and effort put into this PR. At this time we want to keep the installation instructions in one place, rather than in each carvel repository. Please do let us know if you have any feedback on how to make the installation instructions from the carvel.dev site more discoverable in ytt.

PR #256, opened 2020-12-15
Make overlay error order deterministic

hanlins:
Currently when ytt overlay hit an error, the error line orders are not deterministic. However many CI systems could make assumptions on the order of the error messages in test cases. Now sort the lines in alphabetical order to make it deterministic.


hanlins:
cc @StevenLocke, PTAL.

gcheadle-vmware:
Accepting PR as a fix for the determinism until upstream source is identified and addressed.

PR #263, opened 2021-01-04
Adding Check for if Issue Closed Timestamp != Comment Created Timestamp

danielhelfand:
Adding a check to prevent the `carvel-triage` label from being added when an issue is closed with a comment.

Issues being closed with a comment will have the same time stamp for both the issue closing/comment, so this check makes sure the time stamps are different (i.e. the comment happened after the issue was closed so the carvel-triage label should be reapplied).

### DIFF
--- a/.github/workflows/closed-issue-label.yml
+++ b/.github/workflows/closed-issue-label.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: docker://k8slt/closed-issue-label-action@sha256:84360552559e0de4dfaaeefcf1ea0a1070a6250621f98539caf3eaad9395feb7
-        if: github.event_name == 'issues' || ( github.event_name == 'issue_comment' && github.event.issue.state == 'closed')
+        if: github.event_name == 'issues' || ( github.event_name == 'issue_comment' && github.event.issue.state == 'closed') && (github.event.issue.closed_at != github.event.comment.created_at)
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           ignore-comments: true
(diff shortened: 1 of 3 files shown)


.github/workflows/closed-issue-label.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: docker://k8slt/closed-issue-label-action@sha256:84360552559e0de4dfaaeefcf1ea0a1070a6250621f98539caf3eaad9395feb7
-        if: github.event_name == 'issues' || ( github.event_name == 'issue_comment' && github.event.issue.state == 'closed')
+        if: github.event_name == 'issues' || ( github.event_name == 'issue_comment' && github.event.issue.state == 'closed') && (github.event.issue.closed_at != github.event.comment.created_at)
```
joaopapereira: I think the logic is not correct here, because the part you added should only apply if this is a new comment added so it need some parenthesis something like:
```suggestion
        if: github.event_name == 'issues' || ( ( github.event_name == 'issue_comment' && github.event.issue.state == 'closed') && (github.event.issue.closed_at != github.event.comment.created_at) )
```

But at this point maybe it makes more sense to split this into maybe 2 jobs, or even 2 workflows? I am suggesting this because the expression was originally "forced" and now that we are adding more boolean logic it feels like maybe this should be separated

What do you think?

joaopapereira (changes requested):
Take a look at the comment I left in the change

.github/workflows/closed-issue-label.yml
danielhelfand: Yeah, I think splitting these in two might make it easier to read. I can reorganize this action into one for reopening and one for applying the comment on a closed issue.

PR #264, opened 2021-01-06
Improve Non Yaml Warning Msg

### LINKED ISSUE #132: Inform user about options when inputs include non-YAML files and not output flag given
_edit: ⚠️ see [Cohen-style User Story format below](https://github.com/k14s/ytt/issues/132#issuecomment-725524292) ⚠️_

---

It's unclear from the result that when a user includes a non-YAML template and the destination is standard out, that its output is not included on purpose.

Let's give them a hint. For example, on stderr, we could say,
```
There are templates marked for output that are not included in this output.
Non-YAML templates are not rendered to standard output.
If you want to include those results, use the --output-directory flag.
```

---
cc: @pivotal-dean

danielhelfand:
Closes #132 

This pull request adds a warning for when non YAML files are provided without output options provided.

### DIFF
--- a/pkg/cmd/template/cmd_test.go
+++ b/pkg/cmd/template/cmd_test.go
@@ -961,6 +961,148 @@ func TestFileMarkedAsAYamlFileDoesNotSuggestsUsingOutputFlags(t *testing.T) {
 	}
 }
 
+func TestFileMarkedAsExclusiveForOutputFileDoesNotSuggestUsingOutputFlags(t *testing.T) {
+	expectedStdErr := ""
+
+	fileData := `EXCLUSIVE OUTPUT
+`
+	testFile, err := createTempFileWithContent([]byte(fileData), "txt")
+	if err != nil {
+		t.Fatalf("Expected writing to test file not to fail: %v", err)
+	}
+	defer os.Remove(testFile.Name())
+
+	outputDir, err := ioutil.TempDir(os.TempDir(), "fakedir")
+	if err != nil {
+		t.Fatalf("Expected creating a temp dir to not fail: %v", err)
+	}
+	defer os.RemoveAll(outputDir)
+
+	testFilePath := filepath.Join(outputDir, filepath.Base(testFile.Name()))
+	expectedStdOut := `creating: ` + testFilePath + "\n"
+
+	fakeStdErr, err := ioutil.TempFile(os.TempDir(), "fakedev")
+	if err != nil {
+		t.Fatalf("Expected creating a temp file to not fail: %v", err)
+	}
+	defer os.Remove(fakeStdErr.Name())
+
+	fakeStdOut, err := ioutil.TempFile(os.TempDir(), "fakedev")
+	if err != nil {
+		t.Fatalf("Expected creating a temp file to not fail: %v", err)
+	}
+	defer os.Remove(fakeStdOut.Name())
+
+	os.Stdout = fakeStdOut
+	os.Stderr = fakeStdErr
+
+	resetStdOutAndErr := func() {
+		os.Stdout = os.NewFile(uintptr(syscall.Stdout), "/dev/stdout")
+		os.Stderr = os.NewFile(uintptr(syscall.Stderr), "/dev/stderr")
+	}
+	defer resetStdOutAndErr()
+
+	templateOptions := &cmdtpl.TemplateOptions{}
+	cmd := cmdtpl.NewCmd(templateOptions)
+	cmd.SetArgs([]string{"-f", testFile.Name(), "--file-mark", fmt.Sprintf("%s:exclusive-for-output=true", filepath.Base(testFile.Name())), "--output-files", outputDir})
+	cmd.SilenceUsage = true
+
+	err = cmd.Execute()
+	if err != nil {
+		t.Fatalf("Unexpected error running command: %v", err)
+	}
+	resetStdOutAndErr()
+
+	stdErrContents, err := ioutil.ReadFile(fakeStdErr.Name())
+	if err != nil {
+		t.Fatalf("Expected reading stderr to not fail: %v", err)
+	}
+
+	if string(stdErrContents) != expectedStdErr {
+		t.Fatalf("Expected std err to have >>>%s<<< \n warning message, but was: >>>%s<<<", expectedStdErr, stdErrContents)
+	}
+
+	stdOutContents, err := ioutil.ReadFile(fakeStdOut.Name())
+	if err != nil {
+		t.Fatalf("Expected reading stdout to not fail: %v", err)
+	}
+
+	if string(stdOutContents) != expectedStdOut {
+		t.Fatalf("Expected std out to be >>>%s<<<\nBut was: >>>%s<<<", expectedStdOut, stdOutContents)
+	}
+
+	if string(stdErrContents) != expectedStdErr {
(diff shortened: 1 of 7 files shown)


pkg/cmd/template/regular_input.go
```diff
@@ -71,11 +71,22 @@ func (s *RegularFilesSource) Output(out TemplateOutput) error {
...
+	nonYamlFileNames := []string{}
 	switch {
 	case len(s.opts.outputDir) > 0:
 		return files.NewOutputDirectory(s.opts.outputDir, out.Files, s.ui).Write()
 	case len(s.opts.outputFiles) > 0:
 		return files.NewOutputDirectory(s.opts.outputFiles, out.Files, s.ui).WriteFiles()
+	default:
+		fileMarkMap := convertStrArrtoKeyValues(out.FileMarks)
+		for _, file := range out.Files {
+			if !strings.Contains(file.RelativePath(), ".yml") {
```
danielhelfand: Making a note here this should also include .yaml

cppforlife:
this code: https://github.com/vmware-tanzu/carvel-ytt/pull/264/files#diff-9e45ca4a50fe868e6b6af8c4ff161f5c1262f35f4872f7a8740a2adf4506bb72R80-R89 is trying to understand whether something is a yaml template or not. it seems like there quite a bit of friction in terms of how it's doing it: looking at extensions, trying to check on file marks -- it seems like it's harder than it should be. this leads me to thinking: where is a better place to identify what is yaml template and what is not. more natural place for keeping track of that is within the code that actually decides how to execute files.

separately how does this feature interact with:
- library feature
- files that are just metadata files (eg a csv) that is used as input data
- exclusive-for-output file mark

danielhelfand:
> this code: https://github.com/vmware-tanzu/carvel-ytt/pull/264/files#diff-9e45ca4a50fe868e6b6af8c4ff161f5c1262f35f4872f7a8740a2adf4506bb72R80-R89 is trying to understand whether something is a yaml template or not. it seems like there quite a bit of friction in terms of how it's doing it: looking at extensions, trying to check on file marks -- it seems like it's harder than it should be. this leads me to thinking: where is a better place to identify what is yaml template and what is not. more natural place for keeping track of that is within the code that actually decides how to execute files.

It would be nice if OutputFiles in this case know whether they have been marked/what type they are. `files.File` has something similar to this, so maybe it's a matter of adding to OutputFiles? It's not clear to me why there is a need for Output versus regular files. It would be more ideal to use `files.File` in this scenario than the OutputFile.

> separately how does this feature interact with:
> 
> * library feature

Is there an example you could recommend for this feature to be tested against?

> * files that are just metadata files (eg a csv) that is used as input data

Is there an example you could recommend for this feature to be tested against?

> * exclusive-for-output file mark

We can add some additional testing if needed but at the very least can manually verify once learning a bit more about the option.

cppforlife:
> It's not clear to me why there is a need for Output versus regular files.

just two different concepts. OutputFile doesnt need to do things that files.File does.

i dont believe OutputFile is even created for things that arent included in the output. you may want to consider collecting that separately or expand OutputFile to include indication whether it's included or not. 

> separately how does this feature interact with:

this was about how these features overlap in functionality (does it make sense or not). not whether they "work".

pkg/files/output_file.go
```diff
@@ -11,10 +11,11 @@ import (
 type OutputFile struct {
 	relativePath string
 	data         []byte
+	MarkedType   Type
```
cppforlife: might be worth changing MarkedType to private and adding a readonly accessor similar to RelativePath() and Bytes()

pkg/cmd/template/regular_input.go
```diff
@@ -101,5 +108,10 @@ func (s *RegularFilesSource) Output(out TemplateOutput) error {
 	s.ui.Debugf("### result\n")
 	s.ui.Printf("%s", combinedDocBytes) // no newline
 
+	if len(nonYamlFileNames) > 0 {
+		s.ui.Warnf("\n"+`Warning: Non-YAML templates are not rendered to standard output.
+If you want to include those results, use the --output-files or --dangerous-emptied-output-directory flag.
+Non-YAML files are: [%s]`+"\n", strings.Join(nonYamlFileNames, ", "))
```
cppforlife: whats should we do when there is a lot of files? imagine one yaml template that uses data.read of 100 csv files. we would end up with a enormous warning and in this case use knows that it's perfectly fine.

pkg/cmd/template/regular_input.go
```diff
@@ -71,11 +71,18 @@ func (s *RegularFilesSource) Output(out TemplateOutput) error {
...
 	}
 
+	nonYamlFileNames := []string{}
 	switch {
 	case len(s.opts.outputDir) > 0:
 		return files.NewOutputDirectory(s.opts.outputDir, out.Files, s.ui).Write()
 	case len(s.opts.outputFiles) > 0:
 		return files.NewOutputDirectory(s.opts.outputFiles, out.Files, s.ui).WriteFiles()
+	default:
+		for _, file := range out.Files {
```
cppforlife, line 80: this looks much cleaner compared to previous iteration. i think we are on the right path!

pkg/cmd/template/regular_input.go
```diff
@@ -20,8 +21,7 @@ const (
 )
 
 type RegularFilesSourceOpts struct {
-	files     []string
-	fileMarks []string
+	files []string
```
cppforlife, line 23: this isnt unsed right?

pkg/cmd/template/cmd_test.go
```diff
@@ -826,3 +831,513 @@ func TestLoadYTTModuleFailEarly(t *testing.T) {
 		t.Fatalf("Expected RunWithFiles to fail with error, but was '%s'", out.Err.Error())
 	}
 }
+
+func TestNonYamlFilesSuggestsUsingOutputFlags(t *testing.T) {
```
cppforlife: if we are going to have top level test for this feature, we should do something similar to this: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/cmd/template/cmd_test.go#L15. you may have to bubble up to testing against Run() function and in that case you should inject your own ui as interface instead of trying to capture os.* fds.

pkg/files/output_file.go
danielhelfand: This has been made not public and is available via a Type() function for OutputFiles.

pkg/cmd/template/regular_input.go
danielhelfand: We have decided to remove the files as part of the warning message. Perhaps this can be revisited in the future with some kind of verbose option, the message by itself seems okay for now.

pkg/cmd/template/regular_input.go
danielhelfand: It is not used.

pkg/cmd/template/cmd_test.go
danielhelfand: @pivotal-dean and I reviewed the tests and decided to reduce the amount of tests included. We decided to keep testing at the command level versus using `Run()` because it is easier to understand as far as what actual command is being run and the expected output.

We still need to use the redirecting of stderr/out because the stderr is only written via `ui.Warnf` and no error is ever returned from `Output()`.

We were able to refactor the test to focus on `Output()` for the main test in regular_input_test.go, but the higher level command tests we thought were valuable for including tests with `--output-files`.

PR #265, opened 2021-01-09
Use defaults from Schema as base for Data Values

pivotaljohn:


pivotaljohn:
@cppforlife — if you get a moment, this is a good spot for review/feedback of all kinds.

pkg/cmd/template/schema_test.go
```diff
@@ -13,8 +13,115 @@ import (
...
+  username: ""
+  password: ""
+  metadata:
+    run: jobName
+top_level: ""
+`
+		dataValuesYAML := `#@data/values
+---
+db_conn:
+#@overlay/append
```
StevenLocke, line 77: Nice to have:
This feels off to me. We're not really appending anything, we need an "overlay/match" annotation and append gives us one for free. This likely already existed before this commit :)

pkg/cmd/template/schema_test.go
```diff
@@ -13,8 +13,115 @@ import (
...
+		schemaYAML := `#@schema/match data_values=True
+---
+- ""
+`
+		dataValuesYAML := `#@data/values
+---
+#@overlay/append
+- first
+#@overlay/append
+- second
```
StevenLocke, line 120: Same nice to have, this doesn't seem like an append is what we're aiming for, just trying to match.

pkg/cmd/template/schema_test.go
```diff
@@ -24,666 +131,523 @@ vpc:
...
 `
 
-	testSchemaTemplates(t, schemaYAML, dataValuesYAML, templateYAML, expected)
+		testSchemaTemplates(t, schemaYAML, dataValuesYAML, templateYAML, expected)
+	})
 }
 
-func TestMapOnlySchemaChecksOk(t *testing.T) {
-	schemaYAML := `#@schema/match data_values=True
+func TestDataValueNotConformingToSchemaFails(t *testing.T) {
```
StevenLocke, line 296: Nice to have:
It would be great if all of these "expected failure" tests could use a test helper like the success cases. Instead of performing the setup and checking out.Err, passing an expected error string may be sufficient.

pkg/cmd/template/schema_test.go
```diff
@@ -24,666 +131,523 @@ vpc:
...
+    private: true
+`
+		dataValuesYAML := `#@data/values
+---
+vpc:
+  subnet_config:
+  #@overlay/append
   - id: 2
+  #@overlay/append
   - id: 3
```
StevenLocke, line 496: Same NTH, match > append?

pkg/cmd/template/schema_test.go
```diff
@@ -24,666 +131,523 @@ vpc:
...
+
+		expectedErr := "Array items cannot be annotated with #@schema/nullable (schema.yml:6). If this behaviour would be valuable, please submit an issue on https://github.com/vmware-tanzu/carvel-ytt"
+		if !strings.Contains(out.Err.Error(), expectedErr) {
+			t.Fatalf("Expected an error about a schema check failure, but got: %s", out.Err.Error())
+		}
 	})
+}
 
-	ui := cmdcore.NewPlainUI(false)
+func TestSchemaFeatureIsNotEnabledButSchemaIsPresentReportsAWarning(t *testing.T) {
```
StevenLocke, line 610: Are we missing an assertion here? This name seems to indicate that a warning should exist, but I don't see any verification for it.

pkg/cmd/template/schema_test.go
```diff
@@ -711,57 +673,25 @@ rendered: true`
...
-	if out.Err == nil {
-		t.Fatalf("Expected RunWithFiles to fail with message about schema enabled but no schema provided, but was a success")
-	}
-
-	if !strings.Contains(out.Err.Error(), "Schema experiment flag was enabled but no schema document was provided") {
-		t.Fatalf("Expected an error about schema enabled but no schema provided, but got: %v", out.Err.Error())
-	}
-}
-
 func testSchemaTemplates(t *testing.T, schemaYAML string, dataValuesYAML string, templateYAML string, expected string) {
```
StevenLocke: This is an awesome test helper! It makes the tests so much cleaner.

As we've been reviewing the changes, we thought of a couple ways to augment it.
Accept an error message parameter to denote when the `RunWithFiles ` call should fail?
Switch the file string params to a `filesToProcess` parameter? This would allow not providing the dataValues or providing multiple.

pkg/cmd/template/cmd.go
```diff
@@ -132,30 +131,32 @@ func (o *TemplateOptions) RunWithFiles(in TemplateInput, ui cmdcore.PlainUI) Tem
...
-	if len(schemaDocs) > 0 {
-		if o.SchemaEnabled {
-			schemaVar, err = schema.NewDocumentSchema(schemaDocs[0])
+
+	var currSchema workspace.Schema
+	var values *workspace.DataValues
+	var libraryValues []*workspace.DataValues
+
+	if o.SchemaEnabled {
+		if len(schemaDocs) > 0 {
```
StevenLocke, line 139: Will there ever be a case where we want to support multiple schema files? If not, should we simply error here instead of taking the first one? (This has been the case for a while, not directly related to this change set)

pkg/workspace/data_values_pre_processing.go
```diff
@@ -39,6 +39,7 @@ func (o DataValuesPreProcessing) Apply() (*DataValues, []*DataValues, error) {
 
 func (o DataValuesPreProcessing) apply(files []*FileInLibrary) (*DataValues, []*DataValues, error) {
 	var values *yamlmeta.Document
+	values = o.loader.schema.AsDataValue()
```
StevenLocke: Nit: Any reason not to use `:=`?

pkg/schema/schema.go
```diff
@@ -5,23 +5,43 @@ package schema
...
 
+type NullSchema struct {
+}
+
 type DocumentSchema struct {
-	Name    string
-	Source  *yamlmeta.Document
-	Allowed *DocumentType
+	Name       string
+	Source     *yamlmeta.Document
```
StevenLocke, line 20: These don't seem to be read anywhere. Is there a future requirement we're setting ourselves up for?

pkg/schema/schema.go
```diff
@@ -5,23 +5,43 @@ package schema
...
 	"fmt"
-
 	"github.com/k14s/ytt/pkg/template"
 	"github.com/k14s/ytt/pkg/yamlmeta"
 )
 
 type AnySchema struct {
 }
 
+type NullSchema struct {
```
StevenLocke, line 15: As far as I can tell, the only distinction between AnySchema and NullSchema is in `ValidateWithValues`. Are there other benefits to creating this struct?

StevenLocke (commented):
In the [glossary](https://hackmd.io/pODV3wzbT56MbQTxbQOOKQ#Glossary) we define "type checking" but our error messages call it "typechecking". Not directly related to this change set, but wanted to note it somewhere.

pkg/cmd/template/schema_test.go
pivotaljohn: Yup, @DennisDenuto and I saw this too. We decided to defer this to a chore: https://www.pivotaltracker.com/story/show/176465929

danielhelfand: @pivotal-dean and I added a NewFakeUI approach in this [commit](https://github.com/vmware-tanzu/carvel-ytt/commit/3429491437f2a42ed066f1cee3787eac8e51d136). Hopefully should help with these cases in future.

pkg/schema/schema.go
```diff
@@ -143,10 +158,63 @@ func newCollectionItemValueType(collectionItemValue interface{}) (yamlmeta.Type,
 	return nil, fmt.Errorf("Collection item type did not match any known types")
 }
 
+func DefaultDataValues(doc *yamlmeta.Document) (*yamlmeta.Document, error) {
```
joaopapereira: What is the intent behind make this a public method? Is the idea that we will be able to use this externally?

Also, this feels like a factory, should it be named `NewDefaultDataValues`?

pkg/schema/schema.go
```diff
@@ -5,23 +5,43 @@ package schema
...
-	Allowed *DocumentType
+	Name       string
+	Source     *yamlmeta.Document
+	defaultDVs *yamlmeta.Document
+	Allowed    *DocumentType
 }
 
 func NewDocumentSchema(doc *yamlmeta.Document) (*DocumentSchema, error) {
-	docType := &DocumentType{Source: doc}
+	docType, err := NewDocumentType(doc)
```
joaopapereira, line 26: An idea:
What if DocumentSchema had a #Parse() error that would internally calculate both the docType and defaultDVs?
This would prevent us from have dangling functions

pkg/schema/schema.go
```diff
@@ -143,10 +158,63 @@ func newCollectionItemValueType(collectionItemValue interface{}) (yamlmeta.Type,
...
+func DefaultDataValues(doc *yamlmeta.Document) (*yamlmeta.Document, error) {
+	docCopy := doc.DeepCopyAsNode()
+	for _, value := range docCopy.GetValues() {
+		setDefaultValues(value)
+	}
+
+	return docCopy.(*yamlmeta.Document), nil
+}
+
+func setDefaultValues(node interface{}) {
```
joaopapereira: Just some food for thought
Having generic `interface{}` in a couple of these methods feels off. Personally, I would like to have the compiler help to ensure we are using the correct types here.
Maybe having an interface called `yamlmeta.Element` or `yamlmeta.Node` or `yamlmeta.Item` not sure naming is hard.

pkg/schema/schema.go
```diff
@@ -143,10 +158,63 @@ func newCollectionItemValueType(collectionItemValue interface{}) (yamlmeta.Type,
...
+	return s.defaultDVs
+}
+
+func (as *AnySchema) ValidateWithValues(valuesFilesCount int) error {
+	return nil
+}
+
+func (n NullSchema) ValidateWithValues(valuesFilesCount int) error {
+	if valuesFilesCount > 0 {
+		return fmt.Errorf("Schema experiment flag was enabled but no schema document was provided")
```
joaopapereira: This message feels like it is at the wrong level of abstraction. Should the Schema package be aware that there is a flag in cmd that is experimental?

pkg/schema/schema.go
```diff
@@ -143,10 +158,63 @@ func newCollectionItemValueType(collectionItemValue interface{}) (yamlmeta.Type,
...
+
+func (s *DocumentSchema) AsDataValue() *yamlmeta.Document {
+	return s.defaultDVs
+}
+
+func (as *AnySchema) ValidateWithValues(valuesFilesCount int) error {
+	return nil
+}
+
+func (n NullSchema) ValidateWithValues(valuesFilesCount int) error {
```
joaopapereira, line 215: I am a little bit confused about this function. Is the intent of the function to error out if a schema is NOT provided and no data values files are present?
Not sure if Schema should be responsible for this check. Maybe the LibraryLoader should have `#HasSchemaFiles() bool` and `#HasValuesFiles() bool`. This way at the `cmd` level we can check if the above is conditions are met to continue or not.

joaopapereira (commented):
Some comments around the structure.

pivotaljohn:
> In the [glossary](https://hackmd.io/pODV3wzbT56MbQTxbQOOKQ#Glossary) we define "type checking" but our error messages call it "typechecking". Not directly related to this change set, but wanted to note it somewhere.

Thanks, @StevenLocke — noted in the upcoming story about taking a thorough charge at error messaging. 👍🏼

pkg/cmd/template/schema_test.go
pivotaljohn: Yes. Good news is that this is not a degradation in UX: before schemas, Configuration Authors have supplied a base Data Values that illustrated that a particular Data Value is a list. Configuration Consumers have already had to supply some kind of `@overlay` annotation on their arrays.

Even better news is that in the very next story, we're aiming to improve that UX further by setting a default behavior (likely `@overlay/replace` on the Array itself) so that, Configuration Consumers can provide configuration _without_ having to learn about overlays, first. 👍🏼   Stay tuned!

pkg/cmd/template/schema_test.go
pivotaljohn: ... and to be clear, the `@overlay/append`s _are_ needed here... one needs to provide _both_ an `@overlay/match` _and_ an operation (e.g. `@overlay/replace`)... it's just that `@overlay/append` is semantically equivalent to `overlay/match` and `@overlay/insert after`.

pkg/cmd/template/schema_test.go
pivotaljohn: Yes! Let's rebase and make use of that approach, here.

pkg/cmd/template/schema_test.go
DennisDenuto: Good points.
filesToProcess is now a parameter that is passed to the helper function.

Also, instead of passing an error argument as suggested, we went with splitting the 'assertion' helper function into 2 separate functions:
1) assertYttWorkflowSucceedsWithOutput
2) assertYttWorkflowFailsWithErrorMessage

pkg/cmd/template/cmd.go
DennisDenuto: Yes we will eventually want to support multiple schemas being provided to ytt and is being prioritized and hopefully worked on soon.

For now, this experimental schema feature will quietly process the first schema file with the intent that multiple schema files will be allowed and processed in the future.

pkg/schema/schema.go
DennisDenuto: There is a near term future requirement where we want to report error messages relating to schema type check. And for that to work we will need to have a reference to the schema source and the name field.

pkg/cmd/template/schema_test.go
DennisDenuto: Nice! we proceeded with the testing approach here in `schema_test.go`
We made some changes, the main one being that cmds now accept a `UI` interface. This allowed for the test code to implement a 'fakeUI' and pass that to a command. The upside being that we could remove the 'NewFakeUI' function that lived in the `template` package.

pkg/workspace/data_values_pre_processing.go
DennisDenuto: 👍

pkg/schema/schema.go
DennisDenuto: Thanks, we have made this function private. This isn't a 'factory' function, it doesn't construct different data value structs, instead it modifies the defaults from the schema document to produce the base for data values.

pkg/schema/schema.go
DennisDenuto: It is a good observation. We think that the current implementation is sufficient for the moment. We should discuss this further. We are going to proceed with this PR as-is. 

Let's continue with your idea further on this github comment thread. And let's see if we can incorporate this idea as part of a future story/chore.

pkg/schema/schema.go
DennisDenuto: Thanks, we are now using the `yamlmeta.Node` type instead of an empty interface. And we are also guarding against values that are not nodes (i.e. scalars)

pkg/schema/schema.go
DennisDenuto: We are going to drop reference to the 'flag' word. And make the error message a bit more generic. "Schema feature is enabled but no schema document was provided"

pkg/schema/schema.go
DennisDenuto: > I am a little bit confused about this function. Is the intent of the function to error out if a schema is NOT provided and no data values files are present?

No, the intent is for when schema is enabled, no schema is provided, and there are data values. Then it should error out.

It is a good observation. We think that the current implementation is sufficient for the moment. We should discuss this further. We've added it to tech forum.

pkg/cmd/template/schema_test.go
DennisDenuto: the @overlay/appends are needed here... one needs to provide both an @overlay/match and an operation (e.g. @overlay/replace)... it's just that @overlay/append is semantically equivalent to overlay/match and @overlay/insert after.

pkg/schema/schema.go
DennisDenuto: Yes that is the only distinction between AnySchema and NullSchema. My understanding is that we are using polymorphism here to communicate different behavior for when a schema isn't provided vs when a schema is completely permissive. John has added this as a tech forum topic to discuss more as a team.

pkg/cmd/template/schema_test.go
DennisDenuto: Thanks. we have moved the failure test cases to a test helper as suggested.

pkg/cmd/template/regular_input.go
```diff
@@ -5,9 +5,9 @@ package template
 
 import (
 	"fmt"
+	ui2 "github.com/k14s/ytt/pkg/cmd/ui"
```
danielhelfand, line 8: Maybe it would make sense to not alias `github.com/k14s/ytt/pkg/cmd/ui`? It has two different aliases now (e.g. `ui2` and `cmdui`, so it would be nice to pick a single alias across the whole project.

pkg/cmd/template/regular_input_test.go
```diff
@@ -155,3 +164,24 @@ func assertStdoutAndStderr(stdout *bytes.Buffer, stderr *bytes.Buffer, expectedS
...
+	fmt.Fprintf(ui.stdout, str, args...)
+}
+
+func (ui fakeUI) Warnf(str string, args ...interface{}) {
+	fmt.Fprintf(ui.stderr, str, args...)
+}
+
+func (ui fakeUI) Debugf(str string, args ...interface{}) {
+	fmt.Fprintf(ui.stderr, str, args...)
+}
```
danielhelfand, line 187: Moving this into this block of tests makes it harder to find this pattern for asserting on ui output streams in my opinion. I think it would make sense to move under the ui package so it can be more easily found/used.

I still do think there should be a constructor for it or its fields should be public since tests are in a separate package.

PR #268, opened 2021-01-14
Remove alias for UI package and add back Mock UI for Testing

danielhelfand:
This is a follow up for pr #265. It addresses the following feedback points:
* Standardize how `github.com/k14s/ytt/pkg/cmd/ui` is aliased when imported (previously was ui2 or cmdui). I just removed the alias for all and refer to it as ui.
* Move the fakeUI concept into the ui package instead of in regular_input_tests. This makes this testing pattern easier to find/use. The changes in #265 make it so the `io.Writer` on the new `tty` struct aren't able to be set for testing since no fields are exported.

### DIFF
--- a/pkg/cmd/ui/tty.go
+++ b/pkg/cmd/ui/tty.go
@@ -9,33 +9,33 @@ import (
 	"os"
 )
 
-type tty struct {
+type TTY struct {
 	debug  bool
 	stdout io.Writer
 	stderr io.Writer
 }
 
-var _ UI = tty{}
+var _ UI = TTY{}
 
-func NewTTY(debug bool) tty {
-	return tty{debug, os.Stdout, os.Stderr}
+func NewTTY(debug bool) TTY {
+	return TTY{debug, os.Stdout, os.Stderr}
 }
 
-func (t tty) Printf(str string, args ...interface{}) {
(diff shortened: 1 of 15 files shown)


pkg/cmd/ui/tty.go
```diff
@@ -47,3 +47,14 @@ type noopWriter struct{}
 var _ io.Writer = noopWriter{}
 
 func (w noopWriter) Write(data []byte) (int, error) { return len(data), nil }
+
+// Used for testing whether TTY writes correct output to stdout/stderr
+func NewMockTTY(debug bool, stdout, stderr io.Writer) tty {
```
cppforlife: this isnt really a mock though. it's `tty` class that allows custom stdout, stderr.

cppforlife: in general public constructors should return public concrete types since it should be possible to describe returned type from the call site. e.g.

```go
var myFoo ____ = ui.NewMockTTY()
```

what is ____ above?

danielhelfand: Renamed to `NewCustomWriterTTY`

danielhelfand: Exported the TTY type. Was there a different concept name you prefer here? I am mainly following the change from #265.

PR #273, opened 2021-01-22
Change labeler

joaopapereira:
Update the action used to label our issues because the previous one did not allow us to remove issues

### DIFF
--- a/.github/workflows/closed-issue-comment.yml
+++ b/.github/workflows/closed-issue-comment.yml
@@ -7,9 +7,8 @@ jobs:
   add-labels:
     runs-on: ubuntu-latest
     steps:
-      - uses: docker://k8slt/closed-issue-label-action@sha256:84360552559e0de4dfaaeefcf1ea0a1070a6250621f98539caf3eaad9395feb7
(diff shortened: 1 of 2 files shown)


.github/workflows/release-published.yml
```diff
@@ -0,0 +1,20 @@
+name: Checks after release is published
+on:
+  release:
+    types: ['published']
+
+jobs:
+  validate-github-action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: k14s/setup-k14s-action@v1
```
danielhelfand: Might be nice to configure this action to only install `ytt`?  I think by default you get everything.

```
- uses: k14s/setup-k14s-action@v1
  with:
    only: ytt
```

Docs: https://github.com/k14s/setup-k14s-action#usage

danielhelfand (commented):
The issue labeling seems to work as expected. Just a minor not on the use of the k14s action.

Can we add a comment on what action is being used above `-uses:` for each?

.github/workflows/release-published.yml
joaopapereira: This file is not supposed to be part of this PR :( going to remove it, but add the suggestion to the PR #274

PR #274, opened 2021-01-22
When a release is published check that github action works

joaopapereira:


.github/workflows/release-published.yml
```diff
@@ -0,0 +1,20 @@
+name: Checks after release is published
+on:
+  release:
+    types: ['published']
+
+jobs:
+  validate-github-action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: k14s/setup-k14s-action@v1
```
cppforlife, line 10: are we trying to test `setup-k14s-action` or release that we made with this test?

joaopapereira: we are trying to test that we can retrieve the latest release of `ytt` using the `setup-k14s-action`

.github/workflows/release-published.yml
```diff
@@ -0,0 +1,22 @@
+name: Checks after release is published
+on:
+  release:
+    types: ['published']
```
StevenLocke, line 4: Do we only want to run this when a release is published? What if we have to pull a release because of a major breaking change? I'd still want to check that we're pulling the current functional one.

I'm not sure what all the release types mean – what's the difference between `published`, `created`, and `released`?

.github/workflows/release-published.yml
```diff
@@ -0,0 +1,22 @@
...
+  validate-github-action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: k14s/setup-k14s-action@v1
+        with:
+          only: ytt
+      - run: |
+          ytt version
+          version=`ytt version`
+          release_version="${GITHUB_REF:1}"
```
StevenLocke: Is github_ref the best way to get the release version? I'm not sure what gets set in there, it seems dependent on the type of trigger.

I see some StackOverflow answers suggesting
`${{ github.event.release.tag_name }}`

.github/workflows/release-published.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  validate-github-action:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: k14s/setup-k14s-action@v1
+        with:
+          only: ytt
+      - run: |
+          ytt version
+          version=`ytt version`
```
StevenLocke, line 15: We probably only need to do this once to capture the version.

.github/workflows/release-published.yml
```diff
@@ -0,0 +1,22 @@
...
+          release_version="${GITHUB_REF:1}"
+          if [[ "${version}" == *"$release_version"* ]];
```
StevenLocke: My own uncertainty: are *s wildcarding? 

If so, what's the reason behind not expecting an exact match?

PR #279, opened 2021-01-26
Schema Error Messages

### LINKED ISSUE #291: Present focusing, clear, complete, and actionable error messages
## Tasks
* [x] Ensure tests pass
* [x] for nil, print "null", not "<nil>"
* [ ] verify we've met the spirit of the acceptance criteria through some exploratory 'testing" — how would _you_ answer the questions from #285 ?

## Story

**As a** Configuration Author
**I want** — when my schema is invalid — helpful error messages
**so that** when I inevitably make mistakes, I can fix them quickly, ideally all by myself (increasing my confidence in the tool and my own skills with it)

and

**As a** Configuration Consumer
**I want** — when my data values do not conform to the schema — helpful error messages
**so that** when I inevitably make mistakes, I can fix them quickly, ideally all by myself (increasing my confidence in the tool and my own skills with it)

...

**where** "helpful error messages:"

1. focus my attention on the source of the problem,
1. gives _all_ of the relevant facts (and _none_ of the irrelevant ones)
1. consistently uses a set of accessible terms
1. makes clear what's wrong, and
1. helps me figure out how to fix it

## Context

https://hackmd.io/pODV3wzbT56MbQTxbQOOKQ#Part-8-Command-Line-Interface-Interactions

-----

### General

Messages provided in this user story are intended to be _a thoughtful starting point_ rather than an _authoritative spec_.

Error messages should...

1. be formatted so as to be effortlessly parsed by the human eye
1. present the most important information in the most prominent location (i.e. where the eye goes first)
1. limit information to what's _directly_ relevant to the error
1. use the fewest words possible

### 🔴 Invalid schema

**Given** A schema that is well-formed YAML, but invalid Schema
**When** I invoke `ytt`
**Then** It reports an error including the location of the violation, the contents there, and the expectation that was not met.

_(hint: the test group_ `TestSchemaIsInvalidItFails` _in_ [`schema_test.go`](https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/cmd/template/schema_test.go) _**ought** to enumerate all the ways in which Schema can be invalid.)_

#### Example 1a: Too many values defining the type of an array

`schema.yml:`
```yaml
#@schema/definition data_values=True
---
list:
- item 1
- item 2
```

```
$ ytt -f schema.yml -f data_values.yml --enable-experiment-schema
ytt: Error: ...

schema.yml:3 | list:
             |      ^^^
             |   found: 2 array items
             |   expected: exactly 1 array item
             |   (hint: to define an array, provide one item of the desired type; give default items in a separate data values document)
```

#### Example 1b: Not enough values defining the type of an array

`schema.yml:`
```yaml
#@schema/definition data_values=True
---
list: []
```

```
$ ytt -f schema.yml -f data_values.yml --enable-experiment-schema
ytt: Error: ...

schema.yml:3 | list:
             |      ^^^
             |   found: 0 array items
             |   expected: exactly 1 array item
             |   (hint: to define an array, provide one item of the desired type; the default value of arrays is an empty list)
```

#### Example 1c: Null value in schema 

`schema.yml:`
```yaml
#@schema/definition data_values=True
---
list: ~
```

```
$ ytt -f schema.yml -f data_values.yml --enable-experiment-schema
ytt: Error: ...

schema.yml:3 | list: ~
             |      ^^^
             |  expected: a non-null value, of the desired type
             |  (hint: to default to null, specify a value of the desired type and annotate with @schema/nullable)
```
Note: "hints" should...
- guard instructions on how to fix behind a guessed intent
- attempts to fill-in likely missing understanding
- contains words, that if used in a search would likely pinpoint relevant documentation

### 🔴 Type checking errors

**Given** a valid schema
**And** data values that do not conform to that schema
**When** I invoke `ytt`
**Then** it reports an error including: the location and line of the offending data value, the failed expectation, and the location in the schema where that expectation is expressed.

_(hint: the test group_ `TestDataValueNotConformingToSchemaFails` _in_ [`schema_test.go`](https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/cmd/template/schema_test.go) _**ought** to enumerate all the ways in which Data Values can fail to conform to schema.)_

#### Example 2a: Schema specifies a string, Data Values contain a boolean

`schema.yml:`
```yaml
#@schema/definition data_values=True
---
address: ""
```

`data_values.yml:`
```yaml
#@data/values
---
address: false
```

```
$ ytt -f schema.yml -f data_values.yml --enable-experiment-schema
ytt: Error: ...

data_values.yml:3 | address: false
                  |          ^^^^^
                  |  found: boolean
                  |  expected: string (by schema.yml:3)
```
Note:
- the file the user is most likely editing (i.e. `data_values.yml`) is most prominently placed.
- ancillary (yet still relevant) information (e.g. which line in the schema) is tucked in a less prominent place.
- each piece of information's proximity is in proportion to its affinity with other pieces (what is "found" is in the `data_values.yml` file), so `found` comes first.

#### Example 2b: Data Values contain a map item key not declared in schema

`schema.yml:`
```yaml
#@schema/definition data_values=True
---
address: ""
```

`data_values.yml:`
```yaml
#@data/values
---
home_address: ""
```

```
$ ytt -f schema.yml -f data_values.yml --enable-experiment-schema
ytt: Error: ...

data_values.yml:3 | home_address: ""
                  | ^^^^^^^^^^^^
                  |   unexpected key in map (as defined at schema.yml:2)
                  |   (hint: did you mean "address"?)
```

## Implementation Notes

- consider defining a set of terms (taking into account, but not necessarily being bound by prior use of terms) before writing copy for error messages
- don't be bound by current means of accumulating error context (i.e. appending strings while popping up the call stack)
- we're still figuring out the best way to message errors to users; separate the concerns of what information is included with error messages from the presentation of those error messages.

danielhelfand:
Adding error messages for invalid schemas and if data values violate schema.


pkg/schema/error.go
```diff
@@ -0,0 +1,158 @@
...
+		errorWithContent: errorWithContent{
+			Position: position,
+		},
+		Found:    found,
+		Expected: expected,
+		Hint:     hint,
+	}
+}
+
+type invalidError struct {
```
gcheadle-vmware: "invlaidError" makes me think that the error is invalid, not that the schema is invalid.

pkg/schema/error.go
```diff
@@ -0,0 +1,158 @@
...
+type errorWithContent struct {
+	contents string
+	Position *filepos.Position
+}
+
+func (e errorWithContent) Error() string {
+	panic("do not use directly")
+}
+
+func (e *errorWithContent) SetContext(doc *yamlmeta.Document) error {
```
gcheadle-vmware: Odd how in "SetContext" what is actually happening is the content that is being set. Maybe we stick with one of either content or context?

pkg/filepos/position.go
```diff
@@ -39,6 +39,10 @@ func (p *Position) Line() int {
 	return *p.line
 }
 
+func (p *Position) FileString() string {
```
gcheadle-vmware: "FileString" felt ambiguous, maybe an alternative like FileName describes what happens in this function better?

pkg/schema/error.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/k14s/ytt/pkg/filepos"
+	"github.com/k14s/ytt/pkg/yamlmeta"
+)
+
+type errorWithContent struct {
+	contents string
+	Position *filepos.Position
+}
+
+func (e errorWithContent) Error() string {
```
gcheadle-vmware: When deleting this function were found that all the tests still pass and there were no compilation errors. Is the purpose of this function just to make "errorWithContent" adhere to the error interface? Why is this here?

pkg/schema/type.go
```diff
@@ -66,101 +72,107 @@ func (m ScalarType) GetValueType() yamlmeta.Type {
...
+		chk.Violations = append(chk.Violations,
+			NewTypeError(node.ValueTypeAsString(), typeToString(m), node.GetPosition(), m.Position))
 		return
 	}
 
 	for _, item := range nodeMap.Items {
 		if !m.AllowsKey(item.Key) {
-			chk.Violations = append(chk.Violations, fmt.Sprintf("Map item '%s' at %s is not defined in schema", item.Key, item.Position.AsCompactString()))
+			chk.Violations = append(chk.Violations,
+				NewUnexpectedKeyError(item.Key.(string), item.Position, m.Position))
```
gcheadle-vmware: We’re casting "item.key.(string)", what makes us confident in this downcast?

pkg/cmd/template/schema_test.go
```diff
@@ -602,7 +662,29 @@ vpc:
...
-		expectedErr := "Array items cannot be annotated with #@schema/nullable (schema.yml:6). If this behaviour would be valuable, please submit an issue on https://github.com/vmware-tanzu/carvel-ytt"
+		expectedErr := `
+schema.yml:6 | - 0
+             |   ^^^
+             |  found: array item with an unexpected annotation
+             |  (hint: array items cannot be annotated with #@schema/nullable, if this behaviour would be valuable, please submit an issue on https://github.com/vmware-tanzu/carvel-ytt)`
+
+		assertYTTWorkflowFailsWithErrorMessage(t, filesToProcess, expectedErr)
+	})
+	t.Run("null value gives hint about nullable annotation", func(t *testing.T) {
```
gcheadle-vmware: This test seemed identical to the one below "null given as a value", maybe this test was used to TDD the hint? Do we need both of these tests?

pkg/cmd/template/schema_test.go
```diff
@@ -393,7 +418,31 @@ not_in_schema: "this should fail the type check!"
 			files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(schemaYAML))),
 			files.MustNewFileFromSource(files.NewBytesSource("values.yml", []byte(dataValuesYAML))),
 		})
-		expectedErr := "Typechecking violations found: [Expected node at values.yml:2 to be nil, but was a *yamlmeta.Map]"
+		expectedErr := "data values were found in data values file(s), but schema (schema.yml) has no values defined\n"
+		expectedErr += "(hint: define matching keys from data values files(s) in the schema, or do not enable the schema feature)"
+
+		assertYTTWorkflowFailsWithErrorMessage(t, filesToProcess, expectedErr)
+	})
+	t.Run("multiple data values given but schema is empty", func(t *testing.T) {
```
gcheadle-vmware: What is the value added by this test? found it difficult to see the difference between this and the one above, "data values is given but schema is empty"

gcheadle-vmware (commented):
@jtigger and I collected notes together during our review of this PR. I have provided the in line code comments and John will provide the higher level review.

pkg/cmd/template/schema_test.go
danielhelfand: It mainly was added to TDD. Originally, this error contained the values file name but only one if multiple were supplied. It helps reinforce the design of the error message.

pkg/schema/error.go
```diff
@@ -0,0 +1,158 @@
...
+	panic("do not use directly")
+}
+
+func (e *errorWithContent) SetContext(doc *yamlmeta.Document) error {
+	contents, err := doc.RawDataAtLine(e.Position)
+
```
cppforlife: nit: remove newline

pkg/yamlmeta/parser.go
```diff
@@ -97,8 +97,8 @@ func (p *Parser) parseBytes(data []byte, lineCorrection int) (*DocumentSet, erro
 			}
 			return nil, err
 		}
-
 		doc := &Document{
+			RawData:  data,
```
cppforlife: data hold entire file afaik in this case, seems a bit weird to store file contents on a document.

pkg/workspace/template_loader.go
```diff
@@ -187,6 +187,8 @@ func (l *TemplateLoader) EvalYAML(libraryCtx LibraryExecutionContext, file *file
 	if err != nil {
 		return nil, nil, err
 	}
+
+	resultVal.(*yamlmeta.DocumentSet).SetRawBytes(docSet)
```
cppforlife: this feels very weird. it's also not clear from the code why this is necessary.

pkg/schema/type.go
```diff
@@ -264,10 +294,14 @@ func (a *ArrayType) AssignTypeTo(typeable yamlmeta.Typeable) (chk yamlmeta.TypeC
 	}
 	return
 }
-func (a ArrayItemType) AssignTypeTo(typeable yamlmeta.Typeable) (chk yamlmeta.TypeCheck) {
+func (a *ArrayItemType) AssignTypeTo(typeable yamlmeta.Typeable) (chk yamlmeta.TypeCheck) {
 	arrayItem, ok := typeable.(*yamlmeta.ArrayItem)
 	if !ok {
-		chk.Violations = []string{fmt.Sprintf("Expected node at %s to be a %T, but was a %T", typeable.GetPosition().AsCompactString(), &yamlmeta.ArrayItem{}, typeable)}
+		chk.Violations = append(chk.Violations,
```
cppforlife: stylistic nit: would be nice to keep things on one line as long as possible, and then break to second line if it really runs too long.

```
chk.Violations = append(chk.Violations, NewTypeError(
  typeToString(typeable), typeToString(a), typeable.GetPosition(), a.Position))
```

pkg/schema/type.go
```diff
@@ -278,15 +312,18 @@ func (a ArrayItemType) AssignTypeTo(typeable yamlmeta.Typeable) (chk yamlmeta.Ty
 	} // else, at a leaf
 	return
 }
-
```
cppforlife, line 281: nit: add back new line

pkg/schema/type.go
```diff
@@ -66,101 +72,107 @@ func (m ScalarType) GetValueType() yamlmeta.Type {
...
 	default:
-		violation := fmt.Sprintf(violationErrorMessage, itemValueType, m.Type)
-		chk.Violations = append(chk.Violations, violation)
+		chk.Violations = append(chk.Violations,
+			NewTypeError(typeToString(itemValueType), typeToString(m.Type), node.GetPosition(), m.Position))
 	}
 	return
 }
 
+func typeToString(value interface{}) string {
```
cppforlife: this would be nice to invert onto yamlmeta.Typeable. e.g. Typeable could have AsDebugString() or something like this.

pkg/cmd/template/schema_test.go
```diff
@@ -4,13 +4,12 @@
...
 
 import (
-	"strings"
-	"testing"
-
 	"github.com/k14s/difflib"
 	cmdtpl "github.com/k14s/ytt/pkg/cmd/template"
 	"github.com/k14s/ytt/pkg/cmd/ui"
 	"github.com/k14s/ytt/pkg/files"
+	"strings"
```
cppforlife: nit: std lib imports separation

pkg/filepos/position.go
cppforlife: one of the rationals for not defining FileString() or similar method was to force consumers to use AsString() or AsCompatString() to provide consistent way of referencing to file location. do we really need to point to a file and not line number?

pkg/schema/error.go
```diff
@@ -0,0 +1,158 @@
...
+	Expected string
+	Message  string
+	Hint     string
+}
+
+func (i invalidError) Error() string {
+	var result string
+	position := i.Position.AsCompactString()
+	lineContent := strings.TrimSpace(i.contents)
+	colonPosition := strings.Index(lineContent, ":")
```
cppforlife: are we assuming that this would always be map key pair? what if key includes a ":"? is there a higher level to work with here instead of working at the string level?

pkg/schema/type.go
```diff
@@ -66,101 +72,107 @@ func (m ScalarType) GetValueType() yamlmeta.Type {
...
-func (t MapItemType) CheckType(node yamlmeta.TypeWithValues, prependErrorMessage string) (chk yamlmeta.TypeCheck) {
-	violationErrorMessage := prependErrorMessage + " was type %T when %T was expected"
-
+func (t *MapItemType) CheckType(node yamlmeta.TypeWithValues) (chk yamlmeta.TypeCheck) {
 	mapItem, ok := node.(*yamlmeta.MapItem)
 	if !ok {
-		chk.Violations = append(chk.Violations, fmt.Sprintf(violationErrorMessage, node.GetValues()[0], node))
-		return
+		// A Map must've yielded a non-MapItem which is not valid YAML
+		log.Panicf("MapItem type check was called on a non-MapItem: %#v", node)
```
cppforlife: should be using plain `panic(...)` for internal consistency checks.

pkg/yamlmeta/deep_copy.go
```diff
@@ -3,96 +3,96 @@
 
 package yamlmeta
 
-func (n *DocumentSet) DeepCopyAsNode() Node { return n.DeepCopy() }
```
cppforlife, line 6: i would typically recommend isolating renaming work into separate commits from feature work. it's harder to grok if something meaningful changed.

pkg/yamlmeta/document.go
```diff
@@ -4,27 +4,42 @@
...
-	if typedArray, isArray := n.Value.(*Array); isArray {
+	if typedArray, isArray := d.Value.(*Array); isArray {
 		return len(typedArray.Items) == 0
 	}
 	return false
 }
 
-func (n *Document) AsYAMLBytes() ([]byte, error) {
-	return yaml.Marshal(convertToLowYAML(convertToGo(n.Value)))
+func (d *Document) RawDataAtLine(position *filepos.Position) (string, error) {
```
cppforlife: it would be good to figure out how to take advantage of existing infrastructure around correlating lines to content: https://github.com/vmware-tanzu/carvel-ytt/blob/e24944a5abee281783e1ad62a483d239a80bcc04/pkg/template/compiled_template_error.go#L183.

pkg/yamlmeta/document_set.go
```diff
@@ -26,32 +26,51 @@ func NewDocumentSetFromBytes(data []byte, opts DocSetOpts) (*DocumentSet, error)
...
-func (n *DocumentSet) AsBytesWithPrinter(printerFunc func(io.Writer) DocumentPrinter) ([]byte, error) {
+func (ds *DocumentSet) SetRawBytes(docSet *DocumentSet) {
+	// TODO: DocumentSet.originalBytes already contain this information
+	//       This sets RawData into the documents, but in the end it contains the full document set bits and not
+	//       a document specific raw data. Eventually it might be ok but we should consider this as it might be a problem
+	//       Some options that I see are:
+	//         rename RawData to originalBytesDocumentSet or something.
+	//         templating will split these originalBytes we have into per document string.
+	//       I think the second option would be the one I would like but I think it might be a little bit of a big change
+	//       because we will have to change the yaml parser to do this....
```
pivotaljohn: 🛑  This TODO needs to be resolved before this PR can be merged.

pivotaljohn (dismissed):
This PR includes a mechanism to re-connect the raw lines of input with the error messages that report against the corresponding node. As written, the approach introduces too much technical debt to be merged in:

- it is too complex and disperses coupling across packages (smell: [shotgun surgery](https://en.wikipedia.org/wiki/Shotgun_surgery)). To understand how raw input that resulted in a node is reconciled with it, one must:
  1. learn that `DocumentSet` attaches a copy of the raw bytes of the document set to the document;
  2. that one must call `TypeCheck.LoadContext()` to re-attach that raw data to the error message; and
  3. that error types must invoke something like `SetContext()` which delegates to `doc.RawDataAtLine()`.
- as is noted in the TODO in `pkg/yamlmeta/document_set.go`, this approach duplicates data that's present on in the `DocumentSet` struct already.
- all other context of an error condition associated with a node comes from the node, why not the line that produced the node?

pkg/schema/error.go
pivotaljohn: There are a couple of places in here that we're introducing an assumption that `MapItem.Key`'s are strings. We need to revisit and generalize.

Another noted here: https://github.com/vmware-tanzu/carvel-ytt/pull/279/files/eb67827d125a4243d43253a460b8c32eb62c17ed#r564928739

pkg/schema/type.go
pivotaljohn: Which "this" are you referring to?

pkg/schema/type.go
pivotaljohn: Agreed. At times, splitting lines by arguments can significantly improve readability, but this doesn't strike me as such a case.

pkg/schema/type.go
cppforlife: this function.

pkg/cmd/template/schema_test.go
pivotaljohn: For GoLand users:

image

pkg/schema/error.go
StevenLocke: 👍 Agreed. Renamed to invalidSchemaError

pkg/schema/error.go
StevenLocke: Removed the function entirely :D

pkg/schema/error.go
StevenLocke: Removed alongside the struct.

pkg/schema/error.go
StevenLocke: Removed setContext entirely.

pkg/yamlmeta/parser.go
StevenLocke: Agreed. The "source" now lives on Position, not the Document.

pkg/workspace/template_loader.go
StevenLocke: Removed by giving ownership of the original line content to Position.

pkg/filepos/position.go
StevenLocke: I agree, it feels bad to have a one-off function for this. It's currently only used in a single error message: one where non-empty data values, and an empty schema were provided. That error looks like this:

```
data values were found in data values file(s), but schema (schema.yml) has no values defined
(hint: define matching keys from data values files(s) in the schema, or do not enable the schema feature)
```

Is there a better way to report it? It felt pretty strange to see something like

```
data values were found in data_values.yml:2, but schema.yml:1 has none defined.
(hint: define matching keys from data values files(s) in the schema, or do not enable the schema feature)
```

What lines could we point to – the first instance of a data value? The document start of a schema?

------------------

After a conversation with @pivotaljohn I was convinced that the File itself isn't meaningful, it's the Document with a given annotation which marks it as a schema. I feel good in showing AsCompactString() of the schema doc, so removed the FileString function.

pkg/schema/type.go
StevenLocke: Updated to use regular panic instead of log.Panic'ing to remove redundant output

pkg/yamlmeta/document.go
StevenLocke: Position now has this responsibility. parse sets it up and the error messages can use it without needing to ask the Document for the content at that line.

pkg/schema/type.go
StevenLocke: TypeWithValues, and therefore Typeable, now has a "ValueTypeAsString" method which is implemented by each of the nodes. The function is still around (though we can stop exporting it) for the "item" nodes and scalars.

pkg/yamlmeta/document_set.go
StevenLocke: Position now has the responsibility of connecting Nodes to their source. parse sets it up and the error messages can use it without needing to ask the Document for the content at that line.

pkg/yamlmeta/deep_copy.go
pivotaljohn: The additional effort to separate portions of a commit that are renames vs. semantic changes is real. It notably took us some time to do that with this change set.

pkg/filepos/position.go
StevenLocke: Removed entirely :)
It isn't the file that matters, it's the appropriately annotated document. Went back to using AsCompactString().

pkg/schema/error.go
StevenLocke: Errors now receive the entire node, no need to worry about whether there was a colon in the key or not. We also stopped trying to point to the value with carets so determining the colon position is no longer required.

PR #298, opened 2021-02-09
Allow only specific types for doc/array/map items

### LINKED ISSUE #297: Overlay inserting docset into array item results in panic
**What steps did you take:**
Applied an overlay over a document.

https://carvel.dev/ytt/#gist:https://gist.github.com/pivotaljohn/6ed947ba79d31449dcb9b61921566429

```yaml
#! base.yml
---
map1:
  map2:
    map3:
    - ""
```

```yaml
#! overlay.yml
#@ load("@ytt:overlay", "overlay")

#@ def make_docset():
---
something: 1
#@ end

#@overlay/match by=overlay.all
---
map1:
  map2:
    map3:
    #@overlay/append
    - #@ make_docset()
```

**What happened:**

```
$ ytt -f base.yml -f overlay.yml
panic: Unexpected docset value within document
goroutine 1 [running]:
github.com/k14s/ytt/pkg/yamlmeta.convertToGo(0x30bd220, 0xc0000327e0, 0x0, 0x0)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/yamlmeta/convert.go:52 +0x734
github.com/k14s/ytt/pkg/yamlmeta.convertToGo(0x3017740, 0xc001068730, 0xc000a22690, 0x0)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/yamlmeta/convert.go:71 +0x47e
github.com/k14s/ytt/pkg/yamlmeta.convertToGo(0x30179c0, 0xc00102a960, 0xc000a22400, 0x0)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/yamlmeta/convert.go:64 +0x5fd
github.com/k14s/ytt/pkg/yamlmeta.convertToGo(0x30179c0, 0xc001025770, 0xc000a22380, 0x0)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/yamlmeta/convert.go:64 +0x5fd
github.com/k14s/ytt/pkg/yamlmeta.convertToGo(0x30179c0, 0xc000ff8050, 0x4f96d8, 0xc0011b6c8e)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/yamlmeta/convert.go:64 +0x5fd
github.com/k14s/ytt/pkg/yamlmeta.(*Document).AsYAMLBytes(0xc000fd2410, 0xc00050d6f8, 0x4, 0x4, 0x4, 0x0)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/yamlmeta/document.go:22 +0x3c
github.com/k14s/ytt/pkg/yamlmeta.(*YAMLPrinter).Print(0xc00113b620, 0xc000fd2410, 0x0, 0x0)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/yamlmeta/printers.go:33 +0x42
github.com/k14s/ytt/pkg/yamlmeta.(*DocumentSet).AsBytesWithPrinter(0xc0003e58f0, 0x3875c40, 0x7, 0x8, 0xc0010ae0c0, 0x7, 0x8)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/yamlmeta/document_set.go:55 +0xec
github.com/k14s/ytt/pkg/yamlmeta.(*DocumentSet).AsBytes(...)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/yamlmeta/document_set.go:40
github.com/k14s/ytt/pkg/workspace.(*LibraryLoader).Eval(0xc0014bacf0, 0xc000e7bbc0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /root/go/pkg/mod/github.com/k14s/ytt@v0.30.0/pkg/workspace/library_loader.go:111 +0x33c
```

translated:
- "the document contains a map
- ... that has a map item containing a map
- ... that has a map item containing a map
- ... that has a map item containing an array
- ... that has an array item containing a document set."

... with the help of these code snippets:

https://github.com/vmware-tanzu/carvel-ytt/blob/v0.30.0/pkg/yamlmeta/convert.go#L57-L66
https://github.com/vmware-tanzu/carvel-ytt/blob/v0.30.0/pkg/yamlmeta/convert.go#L68-L73
https://github.com/vmware-tanzu/carvel-ytt/blob/v0.30.0/pkg/yamlmeta/convert.go#L51-L52

**What did you expect:**
A user-friendly error message _when the overlay was applying_ that a document set is the inappropriate value to insert that the specified location.

Using current error message formatting, something like this:
```
Overlaying (in following order: overlay.yml):
  Document on line overlay.yml:9:
    Map item (key 'map1') on line overlay.yml:10:
      Map item (key 'map2') on line overlay.yml:11:
        Map item (key 'map3') on line overlay.yml:12:
          Array item (index 0) on line overlay.yml:14:
            Expected docset, but was *yamlmeta.Array
```

Was reported by @rosenhouse.

**Environment:**

- ytt version (use `ytt --version`): v0.30.0
- OS (e.g. from `/etc/os-release`):

pivotaljohn:
- fixes #297

### DIFF
--- a/pkg/yamlmeta/node.go
+++ b/pkg/yamlmeta/node.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 
 	"github.com/k14s/ytt/pkg/filepos"
+	"github.com/k14s/ytt/pkg/orderedmap"
 	"github.com/k14s/ytt/pkg/yamlmeta/internal/yaml.v2"
 )
 
@@ -48,8 +49,8 @@ func (ds *DocumentSet) SetValue(val interface{}) error {
 }
 
 func (d *Document) SetValue(val interface{}) error {
-	d.Value = val
-	return nil
+	d.ResetValue()
+	return d.AddValue(val)
 }
 
 func (m *Map) SetValue(val interface{}) error {
@@ -57,11 +58,8 @@ func (m *Map) SetValue(val interface{}) error {
 }
 
 func (mi *MapItem) SetValue(val interface{}) error {
-	if isMapOrArrayItem(val) {
-		return fmt.Errorf("cannot set map-or-array-item value (%T) into mapitem", val)
-	}
-	mi.Value = val
-	return nil
+	mi.ResetValue()
+	return mi.AddValue(val)
 }
 
 func (a *Array) SetValue(val interface{}) error {
@@ -69,11 +67,24 @@ func (a *Array) SetValue(val interface{}) error {
 }
 
 func (ai *ArrayItem) SetValue(val interface{}) error {
-	if isMapOrArrayItem(val) {
-		return fmt.Errorf("cannot set map-or-array-item value (%T) into arrayitem", val)
+	ai.ResetValue()
+	return ai.AddValue(val)
+}
+
+func isValidValue(val interface{}) bool {
+	switch val.(type) {
+	case *Map, *orderedmap.Map,
+		*Array, []interface{},
+		int, int8, int16, int32, int64,
+		uint, uint8, uint16, uint32, uint64,
+		float32, float64,
+		bool,
+		string,
+		nil:
+		return true
+	default:
+		return false
 	}
-	ai.Value = val
-	return nil
 }
 
 func (ds *DocumentSet) ResetValue() { ds.Items = nil }
@@ -92,8 +103,8 @@ func (ds *DocumentSet) AddValue(val interface{}) error {
 }
 
 func (d *Document) AddValue(val interface{}) error {
-	if isMapOrArrayItem(val) {
-		return fmt.Errorf("cannot add map-or-array-item value (%T) into document", val)
+	if !isValidValue(val) {
+		return fmt.Errorf("documents can only contain arrays, maps, or scalars; this is a %T", val)
 	}
 	d.Value = val
 	return nil
@@ -108,8 +119,8 @@ func (m *Map) AddValue(val interface{}) error {
 }
 
 func (mi *MapItem) AddValue(val interface{}) error {
-	if isMapOrArrayItem(val) {
-		return fmt.Errorf("cannot add map-or-array-item value (%T) into mapitem", val)
+	if !isValidValue(val) {
+		return fmt.Errorf("mapitems can only contain arrays, maps, or scalars; this is a %T", val)
 	}
 	mi.Value = val
 	return nil
@@ -124,20 +135,11 @@ func (a *Array) AddValue(val interface{}) error {
 }
 
 func (ai *ArrayItem) AddValue(val interface{}) error {
-	if isMapOrArrayItem(val) {
-		return fmt.Errorf("cannot add map-or-array-item value (%T) into arrayitem", val)
+	if !isValidValue(val) {
+		return fmt.Errorf("arrayitems can only contain maps, arrays, or scalars; this is a %T", val)
 	}
 	ai.Value = val
 	return nil
-}
-
-func isMapOrArrayItem(val interface{}) bool {
-	switch val.(type) {
-	case *MapItem, *ArrayItem:
-		return true
-	default:
-		return false
-	}
 }
 
 func (ds *DocumentSet) GetValues() []interface{} {
--- a/pkg/cmd/template/cmd_test.go
+++ b/pkg/cmd/template/cmd_test.go
@@ -826,3 +826,84 @@ func TestLoadYTTModuleFailEarly(t *testing.T) {
 		t.Fatalf("Expected RunWithFiles to fail with error, but was '%s'", out.Err.Error())
 	}
 }
+
+func TestBuildingInvalidYAMLFailsFast(t *testing.T) {
+	t.Run("setting document value to DocumentSet", func(t *testing.T) {
+		tpl := []byte(`
+#@ def/end docset():
+---
+foo: true
+
+--- #@ docset()
+`)
+
+		filesToProcess := files.NewSortedFiles([]*files.File{
+			files.MustNewFileFromSource(files.NewBytesSource("template.yml", tpl)),
+		})
+
+		ui := ui.NewTTY(false)
+		opts := cmdtpl.NewOptions()
+
+		out := opts.RunWithFiles(cmdtpl.Input{Files: filesToProcess}, ui)
+		if out.Err == nil {
+			t.Fatalf("Expected RunWithFiles to fail")
+		}
+
+		if !strings.Contains(out.Err.Error(), "documents can only contain arrays, maps, or scalars") {
+			t.Fatalf("Expected RunWithFiles to fail with error, but was '%s'", out.Err.Error())
+		}
+	})
+	t.Run("setting map item to DocumentSet", func(t *testing.T) {
+		tpl := []byte(`
+#@ def/end docset():
+---
+foo: true
+
+---
+bar: #@ docset()
+`)
+
+		filesToProcess := files.NewSortedFiles([]*files.File{
+			files.MustNewFileFromSource(files.NewBytesSource("template.yml", tpl)),
+		})
+
+		ui := ui.NewTTY(false)
+		opts := cmdtpl.NewOptions()
+
+		out := opts.RunWithFiles(cmdtpl.Input{Files: filesToProcess}, ui)
+		if out.Err == nil {
+			t.Fatalf("Expected RunWithFiles to fail")
+		}
+
+		if !strings.Contains(out.Err.Error(), "mapitems can only contain arrays, maps, or scalars") {
+			t.Fatalf("Expected RunWithFiles to fail with error, but was '%s'", out.Err.Error())
+		}
+	})
+	t.Run("setting array item to DocumentSet", func(t *testing.T) {
+		tpl := []byte(`
+#@ def/end docset():
+---
+foo: true
+
+---
+bar:
+- #@ docset()
+`)
+
+		filesToProcess := files.NewSortedFiles([]*files.File{
+			files.MustNewFileFromSource(files.NewBytesSource("template.yml", tpl)),
+		})
+
+		ui := ui.NewTTY(false)
+		opts := cmdtpl.NewOptions()
+
+		out := opts.RunWithFiles(cmdtpl.Input{Files: filesToProcess}, ui)
+		if out.Err == nil {
+			t.Fatalf("Expected RunWithFiles to fail")
+		}
+
+		if !strings.Contains(out.Err.Error(), "arrayitems can only contain maps, arrays, or scalars") {
+			t.Fatalf("Expected RunWithFiles to fail with error, but was '%s'", out.Err.Error())
+		}
+	})
+}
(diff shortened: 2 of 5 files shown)


pkg/yamlmeta/node.go
```diff
@@ -6,6 +6,7 @@ package yamlmeta
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/k14s/ytt/pkg/orderedmap"
```
cppforlife: dont think we should be taking on orderedmap dependency from yamlmeta pkg.

pkg/yamlmeta/node.go
```diff
@@ -69,13 +70,26 @@ func (a *Array) SetValue(val interface{}) error {
...
+	if !(val == nil || isMapOrArrayOrScalar(val)) {
+		return fmt.Errorf("arrays can contain only maps, arrays, or scalars; this is a %T", val)
 	}
 	ai.Value = val
 	return nil
 }
 
+func isMapOrArrayOrScalar(val interface{}) bool {
+	switch val.(type) {
+	case *Map, *orderedmap.Map:
```
cppforlife: do values actually end up being orderedmap or []interface{} at this level?

pkg/yamlmeta/node.go
```diff
@@ -57,8 +58,8 @@ func (m *Map) SetValue(val interface{}) error {
 }
 
 func (mi *MapItem) SetValue(val interface{}) error {
```
cppforlife, line 60: what about AddValue(...)?

pkg/yamlmeta/node.go
cppforlife: i guess we already do take it on: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/yamlmeta/convert.go#L10

pkg/yamlmeta/node.go
```diff
@@ -69,13 +70,26 @@ func (a *Array) SetValue(val interface{}) error {
...
 	return nil
 }
 
+func isMapOrArrayOrScalar(val interface{}) bool {
+	switch val.(type) {
+	case *Map, *orderedmap.Map:
+		return true
+	case *Array, []interface{}:
+		return true
+	case int, int8, int16, int32, int64, bool, string, float32, float64:
```
cppforlife: uints? not sure if that matters.

pkg/yamlmeta/node.go
cppforlife: spacing from std imports.

pkg/yamlmeta/node.go
cppforlife: should we implement SetValue as ResetValue + AddValue?

pkg/yamlmeta/node.go
```diff
@@ -69,13 +70,26 @@ func (a *Array) SetValue(val interface{}) error {
 }
 
 func (ai *ArrayItem) SetValue(val interface{}) error {
-	if isMapOrArrayItem(val) {
-		return fmt.Errorf("cannot set map-or-array-item value (%T) into arrayitem", val)
+	if !(val == nil || isMapOrArrayOrScalar(val)) {
```
cppforlife: it would be good i think to move nil check within isMapOrArrayOrScalar. (if conditional looks a bit confusing at first glance.)

pkg/yamlmeta/node.go
cppforlife, line 60: looks like there is couple places where SetValue isnt checked for error return (in overlay pkg). probably worth throwing it in as well.

cppforlife, line 60: what about Document's SetValue?

pkg/yamlmeta/node.go
pivotaljohn: I too went back-and-forth on this a bit.

- acknowledged: `yamlmeta` does not depend on the _behavior_ of `orderedmap.Map`
- the spirit of this PR is to shift from deny-listing values from Nodes to allow-listing. I believe that necessitates naming types explicitly, including `orderedmap.Map`.
- `orderedmap` is a utility-level package providing a generic data structure. As a dependency, while not massless, it's light.

pkg/yamlmeta/node.go
pivotaljohn: They do: when nodes originate as Starlark values:

Dict:
https://github.com/vmware-tanzu/carvel-ytt/blob/01e2421ff04a2fa22c634b7cb3eedf04e054d7aa/pkg/template/core/starlark_value.go#L107-L116

List:
https://github.com/vmware-tanzu/carvel-ytt/blob/01e2421ff04a2fa22c634b7cb3eedf04e054d7aa/pkg/template/core/starlark_value.go#L140-L150

pkg/yamlmeta/node.go
pivotaljohn: 🤦🏼 , 🙏🏼 , 👍🏼 .

pkg/yamlmeta/node.go
pivotaljohn: Yes. The point is to permit _all_ legit values. Thank you!

pkg/yamlmeta/node.go
pivotaljohn: 👍🏼

pkg/yamlmeta/node.go
pivotaljohn: > what about AddValue(...)?

re:🤦🏼  Yup!

> should we implement SetValue as ResetValue + AddValue?

Will seriously consider this (for this exact reason).

pkg/yamlmeta/node.go
pivotaljohn: 🤔 ... and it also puts pressure on revealing the intention of the function name: to ensure that `val` is assignable as a value to a node within a document.

pkg/yamlmeta/node.go
pivotaljohn: Thanks for looking into that. I have a (apparently false) sense of security that we do a good job of error checking.

pivotaljohn: Yeah, I thought about it for a minute...

This whole thing came about because within a template, one can break YAML structure through an expression. But AFAIK, there's no way to make the value of a Document contain a Document or DocumentSet.

But without it, there's this question of whether that omission is design intent, partial implementation, or a miss.

I think it's worth clearing up that ambiguity. 👍🏼

cppforlife: > But AFAIK, there's no way to make the value of a Document contain a Document or DocumentSet.

```yaml
--- #@ docset()
```

cppforlife: it's typically good. i just like to double check.

pkg/yamlmeta/node.go
cppforlife: yup, i ended up looking at go docs to understand what other types are available. this also just reminded me of this: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/template/core/go_value.go#L39

it would probably be good to also panic for unordered maps, similar to this: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/template/core/go_value.go#L39

pkg/yamlmeta/node.go
pivotaljohn: > ``` 
> --- #@ docset()
> ```

As I was writing that sentence I thought to myself, "and saying this is likely the fastest way to learn how to do so." 😬 

And yes, of course, we do this constantly when we inject the evaluated contents of a private library... :)

pkg/yamlmeta/node.go
pivotaljohn: >yup, i ended up looking at go docs to understand what other types are available. this also just reminded me of this: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/template/core/go_value.go#L39

🤔 

👉🏼  `yamlmeta` viewed as a mini-library will exhaustively allow-list values onto nodes.
👉🏼  `template.core` in so much that Starlark is an intrinsic integration can safely rely on the specific types emitted through the Go port of the language.

pivotaljohn: > it would probably be good to also panic for unordered maps, similar to this: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/template/core/go_value.go#L39

One advantage of returning an `error` over `panic`'ing is being able to report more context... it's the symptom that drove this PR in the first place. It helps answer the question, "from where did that raw Go map come from?"

pkg/yamlmeta/node.go
pivotaljohn: I just created a targeted inspect profile in my IDE to keep a pulse on this. There's one other place that's suspicious, will take a look after getting this PR landed.

https://github.com/vmware-tanzu/carvel-ytt/blob/01e2421ff04a2fa22c634b7cb3eedf04e054d7aa/pkg/template/core/go_value.go#L108

pkg/yamlmeta/node.go
```diff
@@ -48,32 +49,41 @@ func (ds *DocumentSet) SetValue(val interface{}) error {
...
 }
 
 func (ai *ArrayItem) SetValue(val interface{}) error {
-	if isMapOrArrayItem(val) {
-		return fmt.Errorf("cannot set map-or-array-item value (%T) into arrayitem", val)
+	ai.ResetValue()
+	return ai.AddValue(val)
+}
+
+func isValidValue(val interface{}) bool {
```
cppforlife, line 74: looks like isMapOrArrayItem() is dead code after this.

pkg/yamlmeta/node.go
```diff
@@ -48,32 +49,41 @@ func (ds *DocumentSet) SetValue(val interface{}) error {
...
+	switch val.(type) {
+	case *Map, *orderedmap.Map,
```
cppforlife, line 76: nit: probably better to use `default` case to return false

pkg/yamlmeta/node.go
pivotaljohn: It is... I meant to delete but forgot. ty.

pkg/yamlmeta/node.go
pivotaljohn: 🤔 

So, `default` is optional in Go...
I don't see any semantic difference between a `default` over a final-line return.

Hmmm...

In what way is using the `default` case over a final return statement better?

Is it ... that this `switch` guarantees it covers _all_ cases ... and so is grammatically "tighter?"

cppforlife: > Is it ... that this switch guarantees it covers all cases ... and so is grammatically "tighter?"

this is purely my readability bias. feel free to ignore. in this case seems like we are trying to cover all cases, hence default in my head is natural exit point to have. (if default is added, thankfully go would be smart enough to see that all cases have returns and be happy with that function without trailing return.)

pivotaljohn: Yup, that makes sense. I appreciate turning over even these little things... however subjective or not. Please keep including these kinds of reflections alongside the meatier ones; I enjoy it all.

Issue #114, opened 2020-04-06
[lang] stricter annotation building


cppforlife:
- [ ] check for unknown annotations
  - currently does not blow up if annotation is not known
- [ ] check for annotations used multiple times
  - currently that's allowed (last one wins)

cppforlife:
it will be useful to catch problems of missing space as well eg `#@end`

pivotaljohn:
The allowing multiple annotations issue can result in very unexpected outcomes, like #223. This can be a particularly rough UX.

PR #301, opened 2021-02-16
Default overlay array merge behavior is append

cari-lynn:


cppforlife:
this might be semantics, but is there a difference between default operation being append vs merge behaviour for array items being append (if no match annotation is specified)? im asking this to better understand should we be augmenting merge behaviour for arrays or should we be doing what this PR is doing and switching default operation.

cppforlife:
here is another way to put it, are we actually talking about default for overlay/match for array items vs different default operation...

cari-lynn:
There is no difference. I can see how keeping the default operation for all types merge simplifies the implementation by isolating the special case of 'append' merge behavior to array items only. I'll update.

cari-lynn:
Heres some more feedback from talking with @pivotaljohn:

- [x] Order test variables so that expected is separate from inputs
- [x] Modify `TestDocumentOverlayArrayItems` to remove unnecessary keys to simplify test
- [x] Move data values overlay tests from data values test to the overlay test file
- [ ]  ~~(To think about) Would it make sense to assert on `--data-values-inspect` output for data values overlays test instead of template output~~

gcheadle-vmware:
@pivotaljohn @cppforlife, updated PR with changes, feel free to take a look.

examples/playground/overlays/example-append-array-item/append-array-item.yml
```diff
@@ -3,13 +3,12 @@
...
 awards:
-#@overlay/append
 - Ovid Prize
 
 #! Each edit requires two parts:
 #! 1. which item(s) in the collection to target (match)
 #! 2. what operation to perform
 
-#! `@overlay/append`:
+#! No annotation implies a default 'append' operation:
```
StevenLocke:
```suggestion
Omitting an annotation on the array item implies a default 'append' operation:
````

File: examples/playground/getting-started/example-append-array-items/__readme.yaml [hunk 2]
```diff
@@ -4,28 +4,24 @@ readme: |
   Appending Array Items
   =====================
 
-  Map items are easy to match: if the key of the "left" and
-  the key of the "right" are equal, we can assume they
-  refer to the same item. Lemon squeezy, default behavior.
+  In ytt v0.32.0+, array items are appended to the end of
+  a list.
```
StevenLocke:
```suggestion
  In ytt v0.32.0+, unannotated array items are appended to the
  end of the list. This append operation ignores matching
  annotations and always inserts after the last item.
```

examples/playground/getting-started/example-append-array-items/__readme.yaml
```diff
@@ -4,28 +4,24 @@ readme: |
...
 
-  Array items don't have keys; they are referenced by index.
-  We can not assume two items at the same position in the
-  array refer to the same item.
+  For other operations you must specify how to match array
+  items. Do this using the same `#@overlay/match by=...`
+  annotation used on documents.
```
StevenLocke:
```suggestion
  All other array item operations must specify how to match
  as they are not indexed by keys. Do so using the same 
  `#@overlay/match by=...` annotation used on documents.
```

examples/playground/getting-started/example-append-array-items/__readme.yaml
```diff
@@ -4,28 +4,24 @@ readme: |
...
   - `@overlay/merge`
   - `@overlay/replace`
   - `@overlay/remove`
   - `@overlay/insert`
 
-  The one exception is `@overlay/append` which ignores any
-  `@overlay/match` and always inserts after the last item.
+  The one exception is the default 'append' behavior which
+  ignores any `@overlay/match` and always inserts after
+  the last item.
```
StevenLocke:
```suggestion
```

examples/playground/getting-started/example-append-array-items/__readme.yaml
```diff
@@ -4,28 +4,24 @@ readme: |
...
+  items. Do this using the same `#@overlay/match by=...`
+  annotation used on documents.
 
-  In ytt, you must specify how to match array items. Do this
-  using the same `#@overlay/match by=...` annotation used on
-  documents.
-
-  So, all of these operations must be paired with a
+  So, all other operations must be paired with a
   `@overlay/match`:
```
StevenLocke:
```suggestion
  So, all of the following action annotations must be paired with a 
  matching annotation (`@overlay/match`)
```

pkg/cmd/template/cmd_overlays_test.go
```diff
@@ -4,14 +4,16 @@
...
 	"testing"
 
+	"github.com/k14s/difflib"
 	cmdtpl "github.com/k14s/ytt/pkg/cmd/template"
 	"github.com/k14s/ytt/pkg/cmd/ui"
 	"github.com/k14s/ytt/pkg/files"
 )
 
-func TestDocumentOverlays(t *testing.T) {
+func TestDocumentOverlayWithNewKeyAsFunction(t *testing.T) {
```
StevenLocke: Good renames! :clap:

pkg/cmd/template/cmd_overlays_test.go
```diff
@@ -73,7 +75,7 @@ yamlfunc: yamlfunc`)
 	}
 }
 
-func TestDocumentOverlays2(t *testing.T) {
+func TestDocumentOverlayAddAndRemoveKey(t *testing.T) {
```
StevenLocke: :clap::clap:

pkg/cmd/template/cmd_overlays_test.go
```diff
@@ -238,101 +240,531 @@ overlayed: true
 	}
 }
 
-func TestMultipleDataValuesOneEmptyAndOneNonEmpty(t *testing.T) {
-	yamlTplData := []byte(`#@ load("@ytt:data", "data")
+func TestDocumentOverlayArrayItems(t *testing.T) {
+	t.Run("append with and without 'append' annotation", func(t *testing.T) {
```
StevenLocke: Does this need to be an inner `t.Run`? Are we looking to make the `TestDocumentOverlayArrayItems` extendable? If so, maybe we could split the test to overlaying with the annotation, and another to overlaying without? Though that feels like a lot of overhead.


On a second pass, what's the difference between this test and `t.Run("appends array items"`...?

pkg/cmd/template/schema_test.go
```diff
@@ -14,16 +13,10 @@ import (
 	"github.com/k14s/ytt/pkg/files"
 )
 
-var opts *cmdtpl.Options
-
-func TestMain(m *testing.M) {
-	opts = cmdtpl.NewOptions()
```
StevenLocke, line 21: Removing TestMain looks like it lead to some duplicate `opts` creation between tests. I trust that it was causing issues, are there other ways for us to have a `BeforeAll` call?

pkg/yttlibrary/overlay/array.go
```diff
@@ -4,13 +4,18 @@
 package overlay
```
StevenLocke: I love that the actual code change is only a few lines 👨‍🍳

pkg/yttlibrary/overlay/array.go
```diff
@@ -4,13 +4,18 @@
...
 import (
+	"github.com/k14s/ytt/pkg/template"
 	"github.com/k14s/ytt/pkg/yamlmeta"
 )
 
 func (o Op) mergeArrayItem(
 	leftArray *yamlmeta.Array, newItem *yamlmeta.ArrayItem,
 	parentMatchChildDefaults MatchChildDefaultsAnnotation) error {
 
+	if !template.NewAnnotations(newItem).Has(AnnotationMatch) {
```
StevenLocke: It feels a little strange to me to see `NewAnnotations` here instead of `GetAnnotations`.
New is nice because it's properly typed, while Get simply returns an interface{}.

Does this feel any better to you?
`	if !newItem.GetAnnotations().(template.NodeAnnotations).Has(AnnotationMatch) {`

(Though this lacks the cast check)


I think what I'm really suggesting is to rename `template.NewAnnotations`. It's much more related to getting than new-ing, though I want to make sure we capture the distinction between `node.GetAnnotations` and `template.GetAnnotations(node)`

cppforlife: what do you think encapsulating interpretation of overlay/match for array items within NewArrayItemMatchAnnotation class. it seems that it already does a similar check here: https://github.com/vmware-tanzu/carvel-ytt/blob/767de6eada88abc6720dcb2422e6b2f6586f9bc2/pkg/yttlibrary/overlay/array_item_match_annotation.go#L36-L39

cppforlife: if you look at impl of NewAnnotations it does exactly what you are suggesting (https://github.com/vmware-tanzu/carvel-ytt/blob/08ad8d2907e59743ee066f9724c1e3ec30060069/pkg/template/annotations.go#L27).

> I think what I'm really suggesting is to rename template.NewAnnotations. It's much more related to getting than new-ing

im not sure i have strong opinion on this (aside from if we do decide to a rename, it should be a separate commit), but i'll say that new-ing to me implies new-ing up an object (Annotations). it's an implementation detail of NewAnnotations that it's actually reading it off from an existing thing aka get-ing. im guessing based on naming it was only meant to indicate that this is a new object (all other constructors typically start with New).

pkg/cmd/template/schema_test.go
cari-lynn: The way I was thinking of this is: As I am adding tests for `cmd_data_values_test` I want to follow the pattern that we began with `schema_test` like using sub-tests. I am not able to use TestMain in `cmd_data_values_test` to do test setup since it's package public, so I deleted it from schema since we cannot use TestMain as a pattern for all tests. 

It does add duplicate code, but we do need different opts flag setup for tests, take [this schema disabled line 691](https://github.com/vmware-tanzu/carvel-ytt/pull/301/files#diff-2ab902ad23ffe96207ff6466403834df5538e509639024bfc503aed26a377cb8R691) and [this schema enabled line 598](https://github.com/vmware-tanzu/carvel-ytt/pull/301/files#diff-2ab902ad23ffe96207ff6466403834df5538e509639024bfc503aed26a377cb8R597-R598) test. It also shows the prerequisites for each test group in an easy to read way. 

An alternative way we could do this is a setup() function but then we must return the opts. I'm open to thoughts on this.

pkg/cmd/template/cmd_overlays_test.go
cari-lynn: The difference between the two tests is that one is overlaying onto a Document and one is overlaying onto Data Values. We test these separately because the way we match the overlay is different, for Documents:
```
#@overlay/match by=overlay.all
---
```
For Data values (No 'match' annotation because multi data values are an implied overlay):
```
#@data/values
---
```

Agreed on the `t.Run`. It's removed.

pkg/yttlibrary/overlay/array.go
cari-lynn: Great suggestion. I changed `NewArrayItemMatchAnnotation` to return a typed error if a `overlay/match` is not present, and then have `array.go` check the error type. Let me know if that doesn't capture the encapsulation you were thinking of.

cppforlife: i was hinting at to go even further. currently you have a special conditional that short circuits and does the append if overlay/match isnt defined. NewArrayItemMatchAnnotation encapsulates all of the thinking behind how to interpret overlay/match and currently errors if it's not provided. instead of catching that error, we can expand NewArrayItemMatchAnnotation to note that overlay/match was not provided and hence return 0 matches from MatchNodes function [0]. this would keep all the matching logic within that class and would rely on existing check for 0 matched nodes to do appending [1]

[0] https://github.com/vmware-tanzu/carvel-ytt/pull/301/files#diff-880c2538f7db2e19b5ac04ffbc8f4a059e07cd8a6a1444bd0228699d6cdc83ccR87

[1] https://github.com/vmware-tanzu/carvel-ytt/pull/301/files#diff-1ec3c53bb1375f91da8a493c974f9f8f64dd1a3c1a3cc19059ec04bb395ca247R35-R37

gcheadle-vmware: Steven and I are looking into this and we found that if you misspell `#@overlay/match` (making `if !anns.Has(AnnotationMatch)` true) the array item will be appended by default. However we feel like this should be an error case, rather than a default append case.

cppforlife: ^ this is probably out of scope for this issue. typo-ing annotations is a possibility for all annotations so solution needs to be holistic. related issue: https://github.com/vmware-tanzu/carvel-ytt/issues/114

pkg/yttlibrary/overlay/array_item_match_annotation.go
```diff
@@ -76,6 +83,10 @@ func (a ArrayItemMatchAnnotation) Indexes(leftArray *yamlmeta.Array) ([]int, err
 }
 
 func (a ArrayItemMatchAnnotation) MatchNodes(leftArray *yamlmeta.Array) ([]int, []*filepos.Position, error) {
+	if a.unannotated {
+		return []int{}, []*filepos.Position{leftArray.GetPosition()}, nil
```
cppforlife: this should be `return nil, nil, nil` since there are no matches.

pkg/yttlibrary/overlay/array_item_match_annotation.go
```diff
@@ -18,21 +18,28 @@ type ArrayItemMatchAnnotation struct {
...
-		thread:  thread,
-		expects: MatchAnnotationExpectsKwarg{thread: thread},
+		newItem:     newItem,
+		thread:      thread,
+		expects:     MatchAnnotationExpectsKwarg{thread: thread},
+		unannotated: false,
 	}
 	anns := template.NewAnnotations(newItem)
 
+	if len(anns) == 0 {
```
cppforlife: this is a bit problematic. what if my node has #@yaml/text-templated-strings or any other annotation. i think we need to push this functionality in the if conditional below.

pkg/yttlibrary/overlay/array_item_match_annotation.go
StevenLocke: You're right, there are no matches... The intent of this change is to allow the `MatchAnnotationExpectsKwarg.Check` function to pass without error. This may be the wrong way, or it may have been too indirect.

We are really trying to return no matches, without erroring, which would mean we don't have to change 
`mergeArrayItem` in `array.go`:

```Golang
leftIdxs, err := ann.Indexes(leftArray)
if err != nil {
	if err, ok := err.(MatchAnnotationNumMatchError); ok && err.isConditional() {
		return nil
	}
	return err
}

if len(leftIdxs) == 0 {
	return o.appendArrayItem(leftArray, newItem)
}
```

cppforlife: are you saying that `a.expects.Check(matches)` fails because it expects 1 match? i imagine then we should  configure a.expects to expect 0 matches so that its Check succeeds without error.

StevenLocke: That's a good idea. I'm envisioning a change in the place that sets `unannotated = true`, where instead of needing that extra value, we would create a different `expects` for the matchAnnotation. That different "expects" could expect no matches, and return a nil error, which would then nicely default in array.go

examples/playground/overlays/example-append-array-item/append-array-item.yml
```diff
@@ -3,13 +3,13 @@
...
-#@overlay/append
 - Ovid Prize
 
 #! Each edit requires two parts:
 #! 1. which item(s) in the collection to target (match)
 #! 2. what operation to perform
 
-#! `@overlay/append`:
+#! Omitting an annotation on the array item implies a
+#! default 'append' operation:
```
cppforlife: this is no longer true per changes in this PR. it's still a merge operation that does not match any existing node, so append happens

examples/playground/overlays/example-append-array-item/append-array-item.yml
```diff
@@ -3,13 +3,13 @@
 #@overlay/match by=overlay.all
 ---
 awards:
-#@overlay/append
```
cppforlife, line 6: i actually think this example of using overlay/append should stay as is since it's showing how to use overlay/append.

cppforlife: but it could also have an additional variation that does not use overlay/append

examples/playground/getting-started/example-append-array-items/__readme.yaml
```diff
@@ -4,28 +4,23 @@ readme: |
   Appending Array Items
   =====================
 
-  Map items are easy to match: if the key of the "left" and
```
cppforlife, line 7: this section should also probably change a bit given we are just defaulting by for overlay/match. cc @jtigger you may have some opinion on wording/structure.

cppforlife:
implementation code looks fine at this point. circling back to wording in examples.

cppforlife:
actually one thing i missed was tests: i see that we have added a lot of tests at a different level. there are tests here https://github.com/vmware-tanzu/carvel-ytt/tree/develop/pkg/yamltemplate/filetests/ytt-library that exercise various ytt modules including overlays. in this case go tests should only be added for cross cutting functionality. changing overlay/match by default behaviour for array items is not cross cutting however (very isolated to overlay module). i'd like to probably see removal of some of the new tests added (would be good to chat about that since diff is really hard to read)

cari-lynn:
@cppforlife appreciate the feedback. I think I need some clarification on what action to take next. What I hear from 
> go tests should only be added for cross cutting functionality

unit tests (or go tests) should only be added for functionality that relies on or affects other parts of the codebase. Is this what you mean?
 
So then would you say that since this feature is very isolated to the overlay module, and not relying on other parts of the codebase we do not need unit tests that test more than the basic __overlay array item__ functionality here? Or is it that we do not need unit tests that test more than the basic __overlay__ functionality?

Concretely:  in this PR we added 4 unit tests in `cmd_overlays_test` this is a short description of each:
- ['document' overlay with and without overlay/append annotation](https://github.com/vmware-tanzu/carvel-ytt/blob/6dbf361df567222152cc77d9b6bf89183c1406a7/pkg/cmd/template/cmd_overlays_test.go#L243-L261)
- [data value array item overlay](https://github.com/vmware-tanzu/carvel-ytt/blob/6dbf361df567222152cc77d9b6bf89183c1406a7/pkg/cmd/template/cmd_overlays_test.go#L484-L512)
- [empty array data value edge case](https://github.com/vmware-tanzu/carvel-ytt/blob/6dbf361df567222152cc77d9b6bf89183c1406a7/pkg/cmd/template/cmd_overlays_test.go#L524-L546)
- [empty array overlay edge case](https://github.com/vmware-tanzu/carvel-ytt/blob/6dbf361df567222152cc77d9b6bf89183c1406a7/pkg/cmd/template/cmd_overlays_test.go#L558-L580)

If we move the tests above to the following locations does this address your concerns?

1. Keep the basic overlay array item functionality tests in `cmd_overlays_test`:
- 'document' overlay with and without overlay/append annotation
- data value array item overlay

2. Move the `edge case` tests into the [`https://github.com/vmware-tanzu/carvel-ytt/tree/develop/pkg/yamltemplate/filetests/ytt-library`](https://github.com/vmware-tanzu/carvel-ytt/tree/develop/pkg/yamltemplate/filetests/ytt-library) package?
- empty array data value edge case
- empty array overlay edge case

If not, can you elaborate on what testing structure you would like to see?

cari-lynn:
Here's a summary of the conversation @cppforlife and I had regarding this:

The level at which we need to test at depends on if the feature is "cross cutting functionality". For features that interact with others, for example testing data values and overlays together is "cross cutting" since they are two separate features and happen in ['go tests'](https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/cmd/template/cmd_data_values_test.go).

For this PR, we introduced a change in behavior for overlays which is not cross cutting. So it is appropriate to test at a lower level like ['yamltemplate'](https://github.com/vmware-tanzu/carvel-ytt/tree/develop/pkg/yamltemplate/filetests/ytt-library).

In addition to moving the tests to 'yamltemplate', we should decouple testing data values and overlays together, and instead just test overlay functionality.

Lastly, a refactor that was done as part of this PR to move data values overlay tests from [`cmd_data_values_test`](https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/cmd/template/cmd_data_values_test.go) to [`cmd_overlays_test`](https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/cmd/template/cmd_overlays_test.go) should be reverted since data values is a separate feature from overlays, and when we use data values __with__ overlays, that is a 'cross cutting' feature of data values and belongs to `cmd_data_values_test`.

cari-lynn:
@cppforlife Ready for review

PR #305, opened 2021-02-17
Fix schema test to assert on presence of warning message

gcheadle-vmware:
Signed-off-by: Cari Dean <cdean@vmware.com>

### DIFF
--- a/pkg/cmd/template/schema_test.go
+++ b/pkg/cmd/template/schema_test.go
@@ -5,7 +5,6 @@ package template_test
 
 import (
 	"bytes"
-	"fmt"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


pkg/cmd/template/schema_test.go
```diff
@@ -697,8 +701,20 @@ rendered: true`
 		files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(schemaYAML))),
 		files.MustNewFileFromSource(files.NewBytesSource("template.yml", []byte(templateYAML))),
 	})
-	expected := "rendered: true\n"
-	assertYTTWorkflowSucceedsWithOutput(t, filesToProcess, expected)
+
+	expectedStdErr := fmt.Sprintf(`Warning: schema document was detected, but schema experiment flag is not enabled. Did you mean to include --enable-experiment-schema?
```
cppforlife: fmt.Sprintf?

pkg/cmd/template/schema_test.go
```diff
@@ -684,8 +686,10 @@ schema.yml:4 |   subnet_ids: null
 }
 
 func TestSchemaFeatureIsNotEnabledButSchemaIsPresentReportsAWarning(t *testing.T) {
-	opts := cmdtpl.NewOptions()
```
cppforlife, line 687: how are these tests passing if we deleted opts and it's being used on the next line?

cppforlife: ah i see -- https://github.com/vmware-tanzu/carvel-ytt/pull/305/files#diff-2ab902ad23ffe96207ff6466403834df5538e509639024bfc503aed26a377cb8R22

having a global variable even in tests isnt great. how can we do this better?

cari-lynn: Agreed, the global is removed as part a refactor in #301.

Issue #294, opened 2021-02-08
Schema can be overlay'ed


pivotaljohn:
**As a** Configuration Consumer
**I want** to augment the Schema
**so that** I can declare additional data values for use in my own templates/overlays

---

## 🟢  Additional Schema Appends to Root Library's Schema

**Given** there are two or more YAML documents annotated as `@schema/definition`
**When** I run `ytt` that includes those schema
**Then** the effective schema is the merge of those two

```yaml
#! schema-1.yml
#@schema/definition data_values=True
---
foo:
  bar: 3
```

```yaml
#! schema-2.yml
#@schema/definition data_values=True
---
foo:
  #@overlay/match missing_ok=True
  ree: ""
```

```console
$ ytt -f . --data-values-inspect
foo:
  bar: 3
  ree: ""
```

## 🟢  Additional Library-targeted Schema Appends to That Library's Schema

**Given** there is a private library with a YAML document annotated as `@schema/definition`
**And** a schema overlay (from the root library) targets that private library
**When** I run `ytt` that includes those schema
**Then** the effective schema in that library is the merge of those two

```console
.
├── _ytt_lib
│   └── libby
│       ├── schema.yml
│       └── template.yml
├── more-schema.yml
└── root.yml
```

```yaml
#! _ytt_lib/libby/template.yml
#@ load("@ytt:data", "data")
---
libby_data_values: #@ data.values
```

```yaml
#! _ytt_lib/libby/schema.yml
#@schema/definition data_values=True
---
foo:
  bar: 3
```

```yaml
#! more_schema.yml
#@library/ref "@libby"
#@schema/definition data_values=True
---
foo:
  #@overlay/match missing_ok=True
  ree: ""
```

```yaml
#! root.yml
#@ load("@ytt:library", "library")
#@ load("@ytt:template", "template")
#@ load("@ytt:data", "data")

--- #@ template.replace(library.get("libby").with_data_values({"ree": "set from root"}).eval())
---
root_data_values: #@ data.values
```

```console
$ ytt -f .
---
libby_data_values:
  foo:
    bar: 3
    ree: set from root
---
root_data_values: {}
```

## 🟢  Programmatically targeted Schema Appends to That Library's Schema

**Given** there is a private library with a YAML document annotated as `@schema/definition`
**And** the root library configures additional schema, programmatically
**When** I run `ytt` that includes those schema
**Then** the effective schema in that library is the merge of the original schema overlay'ed with the programmatically supplied overlay.

```console
[...]

PR #319, opened 2021-03-12
Private lib schema

### LINKED ISSUE #296: Type check private library data values
**As a** Configuration Consumer using a private library
**I want** the set of data values supplied to the library to be subject to schema defined within the library.
**so that** when I misconfigure the library, I get specific, clear, and actionable feedback

---

### Scenario 1: Declarative Schema-Conforming Data Value

**Given** ytt input that includes a private library
**And** that private library contains schema
**And** a data values document targeting the private library that conforms to schema
**When** I invoke `ytt`
**Then** those data values are reflected in the evaluation of the private library

```
├── _ytt_lib
│   └── fooy
│       ├── schema.yml
│       └── template.yml
├── root.yml
└── values.yml
```

```yaml
#! fooy/schema.yml
---
foo: 42
```
```yaml
#! fooy/template.yml
#@ load("@ytt:data", "data")
---
foo: #@ data.values.foo
```
```yaml
#! root.yml
--- #@ template.replace(library.get("fooy").eval())
```
```yaml
#! values.yml
#@library/ref "@fooy"
#@data/values
---
foo: 7
```

```console
foo: 7
```

### Scenario 2: Declarative Schema-Non-Conforming Data Value

**Given** ytt input that includes a private library
**And** that private library contains schema
**And** a data values document targeting the private library that does *NOT* conform to schema
**When** I invoke `ytt`
**Then** I see a schema error message in the context of a library evaluation error

```yaml
#! values.yml
#@library/ref "@fooy"
#@data/values
---
foo: "wrong type"
```
_(all other files are identical from prior scenario)_

```console
$ ytt -f .
ytt: Error:
- library.eval: Evaluating library 'fooy':
    in <toplevel>
      root.yml:3 | --- #@ template.replace(library.get("fooy").eval())

    reason:
Overlaying data values (in following order: values.yml):
  Templating file 'values.yml':
values.yml:5 | foo: wrong type
             |
             |      found: string
             |   expected: integer (by schema.yml:4)
```

### Scenario 3: Programmatic Schema-Conforming Data Value

**Given** ytt input that includes a private library
**And** that private library contains schema
**And** a template in the root library supplies a data values document to the private library that conforms to the private library's schema
**When** I invoke `ytt`
**Then** those data values are reflected in the evaluation of the private library

```
.
├── _ytt_lib
│   └── fooy
│       ├── schema.yml
│       └── template.yml
└── root.yml
```

```
#! root.yml
---

#@ def dvs_from_root():
#@overlay/match missing_ok=True
foo: from "root" library
#@ end

--- #@ template.replace(library.get("fooy").with_data_values(dvs_from_root()).eval())
```

```
#! fooy/schema.yml
---
foo: ""
```

```
#! fooy/template.yml
#@ load("@ytt:data", "data")
---
foo: #@ data.values.foo
```

```console
foo: from "root" library
```

### Scenario 4: Programmatic Schema-Non-Conforming Data Value

**Given** ytt input that includes a private library
**And** that private library contains schema
**And** a template in the root library supplies a data values document to the private library that does *NOT* conform to the private library's schema
**When** I invoke `ytt`
**Then** I see a schema error message in the context of a library evaluation error

```
#! root.yml
---

#@ def dvs_from_root():
#@overlay/match missing_ok=True
foo: 13
#@ end

--- #@ template.replace(library.get("fooy").with_data_values(dvs_from_root()).eval())
```
_(all other files are identical from prior scenario)_
```console
$ ytt -f .
ytt: Error:
- library.eval: Evaluating library 'fooy':
    in <toplevel>
      root.yml:10 | --- #@ template.replace(library.get("fooy").with_data_values(dvs_from_root()).eval())

    reason:
Overlaying data values (in following order: values.yml):
  Templating file 'values.yml':
root.yml:7 | foo: 13
             |
             |      found: integer
             |   expected: string (by schema.yml:4)
```
_Note: the above error output is inaccurate due to the `values.yml` reference which is non-existent. We think it would `root.yml` in this case but haven't verified, yet._

pivotaljohn:
As part of this PR we decided on some UX decisions described below:
* When `--enable-experiement-schema` flag is **not** present, and a schema file is provided, a warning is shown. **New in this PR** the relative path to the file is also shown. (Consider the case where root library has a schema, and a private library has a schema, there will be **two** warning messages now).
* Rules around how the `--enable-experiement-schema` behaves: 
When the `--enable-experiement-schema` flag is present, and a schema file is given in the root or in a private library, it is used as the base data values and type checked.
When `--enable-experiement-schema` flag is **not** enabled, then if a schema file appears anywhere, the schema is IGNORED and a warning is given. This is because the schema flag is protecting the user against non-GA code.

⚠️   Update based on feedback ⚠️  
We decided to split this original PR into three separate ones.
```
develop 
|_overlay-position       <- PR to address position changes within overlay module (separate for this PR entirely)
|_ refactor-dvpp         <- PR to address refactoring done in data_values_pre_processing.go (must merge prior to this PR)
    |_ private-lib-schema <- this PR
```

### DIFF
--- a/pkg/cmd/template/cmd.go
+++ b/pkg/cmd/template/cmd.go
@@ -126,15 +126,10 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
 	libraryCtx := workspace.LibraryExecutionContext{Current: rootLibrary, Root: rootLibrary}
 	rootLibraryLoader := libraryExecutionFactory.New(libraryCtx)
 
-	currSchema, err := rootLibraryLoader.Schema()
-	if err != nil {
-		return Output{Err: err}
-	}
-
 	var values *workspace.DataValues
 	var libraryValues []*workspace.DataValues
 
-	values, libraryValues, err = rootLibraryLoader.Values(valuesOverlays, currSchema)
+	values, libraryValues, err = rootLibraryLoader.Values(valuesOverlays)
 	if err != nil {
 		return Output{Err: err}
 	}
--- a/pkg/cmd/template/cmd.go
+++ b/pkg/cmd/template/cmd.go
@@ -124,9 +124,9 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
 	})
 
 	libraryCtx := workspace.LibraryExecutionContext{Current: rootLibrary, Root: rootLibrary}
-	libraryLoader := libraryExecutionFactory.New(libraryCtx)
+	rootLibraryLoader := libraryExecutionFactory.New(libraryCtx)
 
-	currSchema, err := libraryLoader.Schemas()
+	currSchema, err := rootLibraryLoader.Schema()
 	if err != nil {
 		return Output{Err: err}
 	}
@@ -134,7 +134,7 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
 	var values *workspace.DataValues
 	var libraryValues []*workspace.DataValues
 
-	values, libraryValues, err = libraryLoader.Values(valuesOverlays, currSchema)
+	values, libraryValues, err = rootLibraryLoader.Values(valuesOverlays, currSchema)
 	if err != nil {
 		return Output{Err: err}
 	}
@@ -149,7 +149,7 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
 		}
 	}
 
-	result, err := libraryLoader.Eval(values, libraryValues)
+	result, err := rootLibraryLoader.Eval(values, libraryValues)
 	if err != nil {
 		return Output{Err: err}
 	}
--- a/pkg/cmd/template/cmd.go
+++ b/pkg/cmd/template/cmd.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/k14s/ytt/pkg/cmd/ui"
 	"github.com/k14s/ytt/pkg/files"
-	"github.com/k14s/ytt/pkg/schema"
 	"github.com/k14s/ytt/pkg/workspace"
 	"github.com/k14s/ytt/pkg/yamlmeta"
 	"github.com/spf13/cobra"
@@ -127,30 +126,13 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
 	libraryCtx := workspace.LibraryExecutionContext{Current: rootLibrary, Root: rootLibrary}
 	libraryLoader := libraryExecutionFactory.New(libraryCtx)
 
-	schemaDocs, err := libraryLoader.Schemas()
+	currSchema, err := libraryLoader.Schemas()
 	if err != nil {
 		return Output{Err: err}
 	}
 
-	var currSchema workspace.Schema
 	var values *workspace.DataValues
 	var libraryValues []*workspace.DataValues
-
-	if o.SchemaEnabled {
-		if len(schemaDocs) > 0 {
-			currSchema, err = schema.NewDocumentSchema(schemaDocs[0])
-			if err != nil {
-				return Output{Err: err}
-			}
-		} else {
-			currSchema = schema.NullSchema{}
-		}
-	} else {
-		if len(schemaDocs) > 0 {
-			ui.Warnf("Warning: schema document was detected, but schema experiment flag is not enabled. Did you mean to include --enable-experiment-schema?\n")
-		}
-		currSchema = &schema.AnySchema{}
-	}
 
 	values, libraryValues, err = libraryLoader.Values(valuesOverlays, currSchema)
 	if err != nil {
--- a/pkg/cmd/template/cmd.go
+++ b/pkg/cmd/template/cmd.go
@@ -121,6 +121,7 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
 		IgnoreUnknownComments:   o.IgnoreUnknownComments,
 		ImplicitMapKeyOverrides: o.ImplicitMapKeyOverrides,
 		StrictYAML:              o.StrictYAML,
+		SchemaEnabled:           o.SchemaEnabled,
 	})
 
 	libraryCtx := workspace.LibraryExecutionContext{Current: rootLibrary, Root: rootLibrary}
--- a/pkg/schema/schema.go
+++ b/pkg/schema/schema.go
@@ -73,10 +73,6 @@ func NewMapType(m *yamlmeta.Map) (*MapType, error) {
 		}
 		mapType.Items = append(mapType.Items, mapItemType)
 	}
-	annotations := template.NewAnnotations(m)
-	if _, nullable := annotations[AnnotationSchemaNullable]; nullable {
-		mapType.Items = nil
-	}
 
 	return mapType, nil
 }
--- a/pkg/schema/schema.go
+++ b/pkg/schema/schema.go
@@ -207,15 +207,15 @@ func (s *DocumentSchema) AssignType(typeable yamlmeta.Typeable) yamlmeta.TypeChe
 	return s.Allowed.AssignTypeTo(typeable)
 }
 
-func (as *AnySchema) AsDataValue() *yamlmeta.Document {
+func (as *AnySchema) DefaultDataValues() *yamlmeta.Document {
 	return nil
 }
 
-func (n NullSchema) AsDataValue() *yamlmeta.Document {
+func (n NullSchema) DefaultDataValues() *yamlmeta.Document {
 	return nil
 }
 
-func (s *DocumentSchema) AsDataValue() *yamlmeta.Document {
+func (s *DocumentSchema) DefaultDataValues() *yamlmeta.Document {
 	return s.defaultDVs
 }
 
--- a/pkg/workspace/data_values_pre_processing.go
+++ b/pkg/workspace/data_values_pre_processing.go
@@ -49,7 +49,7 @@ func (o DataValuesPreProcessing) apply(files []*FileInLibrary) (*DataValues, []*
 	var dataValuesDoc *yamlmeta.Document
 	for _, dv := range allDvs {
 		switch {
-		case dv.HasLib():
+		case dv.HasLibRef():
 			dvsForOtherLibraries = append(dvsForOtherLibraries, dv)
 		case dataValuesDoc == nil:
 			err := o.loader.schema.ValidateWithValues(1)
@@ -89,20 +89,9 @@ func (o DataValuesPreProcessing) apply(files []*FileInLibrary) (*DataValues, []*
 	return dataValues, dvsForOtherLibraries, nil
 }
 
-func (o DataValuesPreProcessing) typeAndCheck(dataValuesDoc *yamlmeta.Document) (chk yamlmeta.TypeCheck) {
-	chk = o.loader.schema.AssignType(dataValuesDoc)
-	if len(chk.Violations) > 0 {
-		return
-	}
-
-	typeCheck := dataValuesDoc.Check()
-	chk.Violations = append(chk.Violations, typeCheck.Violations...)
-	return
-}
-
 func (o DataValuesPreProcessing) collectDataValuesDocs(files []*FileInLibrary) ([]*DataValues, error) {
 	var allDvs []*DataValues
-	if defaults := o.loader.schema.AsDataValue(); defaults != nil {
+	if defaults := o.loader.schema.DefaultDataValues(); defaults != nil {
 		dv, err := NewDataValues(defaults)
 		if err != nil {
 			return nil, err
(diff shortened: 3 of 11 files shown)


pivotaljohn:
> When --enable-experiement-schema flag is not enabled, then if a schema file appears anywhere, the schema is IGNORED and a warning is given. This is because the schema flag is protecting the user against non-GA code.

I really like this choice. It gives the author the ability to support _both_ modes. Include both a `schema.yml` and `values/default.yml` with basically the same content. If I enable schemas, I get the schema checks. If I don't I get the overlay checks.

pkg/yttlibrary/overlay/array.go
```diff
@@ -43,6 +43,8 @@ func (o Op) mergeArrayItem(
 		}
 		if replace {
 			leftArray.Items[leftIdx].Value = newItem.Value
+			//todo: should we be setting the position elsewhere?
+			leftArray.Items[leftIdx].Position = newItem.Position
```
cppforlife: can we keep position related improvements separate from this PR. it's a bit hard to tell if these are the only places where we need to change this -- would love to see holistic evaluation within overlay module.

pkg/cmd/template/cmd.go
```diff
@@ -121,36 +120,20 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
...
 		ImplicitMapKeyOverrides: o.ImplicitMapKeyOverrides,
 		StrictYAML:              o.StrictYAML,
+		SchemaEnabled:           o.SchemaEnabled,
 	})
 
 	libraryCtx := workspace.LibraryExecutionContext{Current: rootLibrary, Root: rootLibrary}
 	libraryLoader := libraryExecutionFactory.New(libraryCtx)
 
-	schemaDocs, err := libraryLoader.Schemas()
+	currSchema, err := libraryLoader.Schemas()
```
cppforlife: is it worth renaming libraryLoader -> rootLibraryLoader

pkg/schema/schema.go
```diff
@@ -73,10 +73,6 @@ func NewMapType(m *yamlmeta.Map) (*MapType, error) {
 		}
 		mapType.Items = append(mapType.Items, mapItemType)
 	}
-	annotations := template.NewAnnotations(m)
```
cppforlife, line 76: note for self: removed because it's dead code.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -37,46 +38,89 @@ func (o DataValuesPreProcessing) Apply() (*DataValues, []*DataValues, error) {
...
 
 func (o DataValuesPreProcessing) apply(files []*FileInLibrary) (*DataValues, []*DataValues, error) {
-	values := o.loader.schema.AsDataValue()
-	var libraryValues []*DataValues
-	for _, fileInLib := range files {
-		valuesDocs, err := o.templateFile(fileInLib)
-		if err != nil {
-			return nil, nil, fmt.Errorf("Templating file '%s': %s", fileInLib.File.RelativePath(), err)
-		}
+	_, checkSchema := o.loader.schema.(*schema.DocumentSchema)
```
cppforlife: something to think about... why do we need a boolean here? why cant we just tell schema to type check against itself. all schemas should be able to do it succesfully.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -37,46 +38,89 @@ func (o DataValuesPreProcessing) Apply() (*DataValues, []*DataValues, error) {
...
 
-		for _, valuesDoc := range valuesDocs {
-			dv, err := NewDataValues(valuesDoc)
+	allDvs, err := o.collectDataValuesDocs(files)
```
cppforlife, line 41: nice little refactor!

pkg/workspace/data_values_pre_processing.go
```diff
@@ -37,46 +38,89 @@ func (o DataValuesPreProcessing) Apply() (*DataValues, []*DataValues, error) {
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// merge all Data Values YAML documents into one
+	var dvsForOtherLibraries []*DataValues
+	var dataValuesDoc *yamlmeta.Document
```
cppforlife: i wonder if naming this from dataValuesDoc -> resultDVsDoc would make it a bit more readable. (it's the result from this loop)

pkg/workspace/data_values_pre_processing.go
```diff
@@ -37,46 +38,89 @@ func (o DataValuesPreProcessing) Apply() (*DataValues, []*DataValues, error) {
...
+	for _, dv := range allDvs {
+		switch {
+		case dv.HasLibRef():
```
cppforlife: this switch statement could be made more readable as two if statements. note that type checking below switch statement would happen even if dvs are for a library ref with current impl.

```
if lib dvs {
   ...
   continue
}

if nil {
  ...
} else {
  ...
}
```

the fact that there is 2 if statements leads the reader to see that it's two unrelated pieces of functionality.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -37,46 +38,89 @@ func (o DataValuesPreProcessing) Apply() (*DataValues, []*DataValues, error) {
...
+			dvsForOtherLibraries = append(dvsForOtherLibraries, dv)
+		case dataValuesDoc == nil:
+			err := o.loader.schema.ValidateWithValues(1)
```
cppforlife: note to self: function goes away as soon as we kill experimental flag.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -37,46 +38,89 @@ func (o DataValuesPreProcessing) Apply() (*DataValues, []*DataValues, error) {
...
-			default:
-				var err error
-				values, err = o.overlay(values, dv.Doc)
-				if err != nil {
-					return nil, nil, err
-				}
+			dataValuesDoc = dv.Doc
+		default:
+			if checkSchema {
+				o.annotateWithDefaultMissingOk(dv.Doc)
```
cppforlife: let's pull this out as a separate PR. by adding missing_ok=True in certain cases we are affecting how overlay feature behaves under certain conditions. that's not desirable because we want users to know that same piece of code behaves same way regardless where it's being called from.

quick example that comes to mind: 
```
#@overlay/match by=overlay.subset({"name":"foo"})
- blah: something
```
will behave differently since since overlay module by default sets expects=1.

to start resolving this situtation, i think we need to figure out what happens in various interactions between schema and overlay. what errors would one expect to see when new keys are used, what errors do we expect to see when new keys are used with any type, etc.. let's have a document enumerating these cases and discuss them.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -37,46 +38,89 @@ func (o DataValuesPreProcessing) Apply() (*DataValues, []*DataValues, error) {
...
+			}
+			allDvs = append(allDvs, dv)
+		}
+	}
+	allDvs = append(allDvs, o.valuesOverlays...)
+	return allDvs, nil
+}
 
-	return dv, libraryValues, nil
+func (o DataValuesPreProcessing) typeAndCheck(dataValuesDoc *yamlmeta.Document) (chk yamlmeta.TypeCheck) {
```
cppforlife: we typically discourage use of named returns. dont think it's necessary in this case.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -37,46 +38,89 @@ func (o DataValuesPreProcessing) Apply() (*DataValues, []*DataValues, error) {
...
+	chk = o.loader.schema.AssignType(dataValuesDoc)
+	if len(chk.Violations) > 0 {
+		return
+	}
+
+	typeCheck := dataValuesDoc.Check()
+	chk.Violations = append(chk.Violations, typeCheck.Violations...)
```
cppforlife: no need to append to chk.Violations since it's guranteed to be 0.

pkg/workspace/data_values_pre_processing.go
cppforlife: accessing private data from loader got me curious... it seems that template loader does not use schema at all. it seems weird for template loader to know anything about schema since it's not dealing with data values. i would have expected DataValuesPreProcessing to be initialized with appropriate schema.

pkg/workspace/data_values_pre_processing.go
```diff
@@ -151,30 +178,23 @@ func (o DataValuesPreProcessing) overlay(valuesDoc, newValuesDoc *yamlmeta.Docum
...
-	for _, valuesOverlay := range o.valuesOverlays {
-		var err error
-
-		result, err = o.overlay(result, valuesOverlay.Doc)
-		if err != nil {
-			// TODO improve error message?
-			return nil, fmt.Errorf("Overlaying additional data values on top of "+
-				"data values from files (marked as @data/values): %s", err)
-		}
+func (o DataValuesPreProcessing) NewEmptyDataValuesDocument() *yamlmeta.Document {
```
cppforlife: seems like this method could be inlined where it's used, or at least made private.

pkg/workspace/library_loader.go
```diff
@@ -46,30 +46,36 @@ func NewLibraryLoader(libraryCtx LibraryExecutionContext,
 	}
 }
 
-func (ll *LibraryLoader) Schemas() ([]*yamlmeta.Document, error) {
+func (ll *LibraryLoader) Schemas() (Schema, error) {
```
cppforlife: if we are returning a single schema should this method be named Schema() as well?

pkg/workspace/library_loader.go
```diff
@@ -46,30 +46,36 @@ func NewLibraryLoader(libraryCtx LibraryExecutionContext,
...
-		}
+			_, resultDocSet, err := loader.EvalYAML(libraryCtx, schemaFiles[0].File)
+			if err != nil {
+				return nil, err
+			}
 
-		docs, _, err := DocExtractor{resultDocSet}.Extract(AnnotationSchemaMatch)
-		if err != nil {
-			return nil, err
+			docs, _, err := DocExtractor{resultDocSet}.Extract(AnnotationSchemaMatch)
```
cppforlife, line 65: if we get multiple schema docs i would expect that we would blow up for now since we do not allow merging schemas.

pkg/workspace/library_module.go
```diff
@@ -318,7 +317,8 @@ func (l *libraryValue) libraryValues(ll *LibraryLoader) (*DataValues, []*DataVal
 		}
 	}
 
-	dvs, foundChildDVss, err := ll.Values(append(dvss, afterLibModDVss...), &schema.AnySchema{})
+	schema, err := ll.Schemas()
```
cppforlife: missing err check.

pkg/workspace/library_module.go
```diff
@@ -318,7 +317,8 @@ func (l *libraryValue) libraryValues(ll *LibraryLoader) (*DataValues, []*DataVal
...
+	dvs, foundChildDVss, err := ll.Values(append(dvss, afterLibModDVss...), schema)
```
cppforlife: it seems that we are using ll.Schemas() and then passing it to ll.Values(). i would argue Values() shouldnt require schema as input and should just use ll.Schemas() internally.

pkg/yttlibrary/overlay/array.go
pivotaljohn: ... yup and that PR would precede this one — this fix addresses an actual bug: the wrong file is attributed a value.

pkg/workspace/data_values_pre_processing.go
pivotaljohn: ... in part because there's no need for any code outside this module to generate an empty Data Values.

pkg/workspace/data_values_pre_processing.go
pivotaljohn: ... and to clarify, it's not so much that we'd pull this out as a separate PR... but that we'd _either_:

- address this problem _within_ this PR, or,
- temporarily accept this intermediate state where it generally works but has undesired side-effects in some cases... but then _immediately_ address this problem in a subsequent PR.  (consider this change set "large enough" and address this as a separate unit of work)

But either way, we can't remove the `--enable-experiment-schemas` flag until we ensure that this works exact the way we want it.

pkg/workspace/data_values_pre_processing.go
pivotaljohn: ... and taking that suggestion a step further, if we use the same naming scheme across identifiers: `(qualifier)(noun)`, the eye parses identifiers using the same pattern matcher, reducing cognitive load.

Example:
- `allDvs`
- `otherLibraryDvs`
- `resultDvDoc`
- others?

pkg/workspace/library_loader.go
cari-lynn: Correct, we have a story coming up to allow multiple schema docs #294. Prior to this story we would also blow up if we get multiple docs, so we haven't added guards for that case.

pkg/workspace/library_module.go
cari-lynn: 😨

pkg/workspace/data_values_pre_processing.go
cari-lynn: We pulled out this to a separate PR where it is addressed, @cppforlife ready for review. The PR is #337 and is a prerequisite to this work so that PR must be merged in prior to this one.

pkg/yttlibrary/overlay/array.go
cari-lynn: These position changes are now necessary to report the position of the overlay in the schema error message, if it's acceptable, we want to keep the change made to the position of arrays and maps here in [this](https://github.com/vmware-tanzu/carvel-ytt/pull/337) PR, and then address this more comprehensively in the [Overlay Positions](https://github.com/vmware-tanzu/carvel-ytt/pull/341) PR.

PR #322, opened 2021-03-13
Allow null values to be overlayed

### LINKED ISSUE #108: [overlay] reconsider ytt behaviour of overlaying map on top of nil
currently overlay module will not allow to do so.

### LINKED ISSUE #296: Type check private library data values

pivotaljohn:
Prep to allow for @schema/nullable values that are subsequently
overlayed with non-null values.

Addresses: vmware-tanzu/carvel-ytt#108

### DIFF
--- a/pkg/yttlibrary/overlay/array.go
+++ b/pkg/yttlibrary/overlay/array.go
@@ -34,9 +34,12 @@ func (o Op) mergeArrayItem(
 	}
 
 	for _, leftIdx := range leftIdxs {
-		replace, err := o.apply(leftArray.Items[leftIdx].Value, newItem.Value, matchChildDefaults)
-		if err != nil {
-			return err
+		replace := true
+		if leftArray.Items[leftIdx].Value != nil {
+			replace, err = o.apply(leftArray.Items[leftIdx].Value, newItem.Value, matchChildDefaults)
+			if err != nil {
+				return err
+			}
 		}
 		if replace {
 			leftArray.Items[leftIdx].Value = newItem.Value
--- a/pkg/yamltemplate/filetests/ytt-library/overlay/merge-over-nil.yml
+++ b/pkg/yamltemplate/filetests/ytt-library/overlay/merge-over-nil.yml
@@ -1,0 +1,55 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:template", "template")
+
+#! test1: overlay a map over a null
+#! test2: overlay an array over a null
+#! test3: overlay a document over a null
+
+---
+#@ def null_map():
+foo: null
+#@ end
+---
+#@ def map():
+foo:
+  bar: 42
+#@ end
+---
+
+test1: #@ overlay.apply(null_map(), map())
+
+
+---
+#@ def null_array():
+- null
+#@ end
+---
+#@ def array():
+#@overlay/match by=overlay.all
+- - bar 42
+#@ end
+---
+test2: #@ overlay.apply(null_array(), array())
+
+#@ def null_doc():
+--- null
+#@ end
+
+#@ def doc():
+#@overlay/match by=lambda i,l,r: True
+---
+test3: foo bar
+#@ end
+
+--- #@ template.replace(overlay.apply(null_doc(), doc()))
+
++++
+
+test1:
+  foo:
+    bar: 42
+---
+test2:
+- - bar 42
+---
+test3: foo bar
--- a/pkg/yttlibrary/overlay/document.go
+++ b/pkg/yttlibrary/overlay/document.go
@@ -30,9 +30,12 @@ func (o Op) mergeDocument(
 	}
 
 	for _, leftIdx := range leftIdxs {
-		replace, err := o.apply(leftDocSets[leftIdx[0]].Items[leftIdx[1]].Value, newDoc.Value, matchChildDefaults)
-		if err != nil {
-			return err
+		replace := true
+		if leftDocSets[leftIdx[0]].Items[leftIdx[1]].Value != nil {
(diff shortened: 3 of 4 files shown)


pkg/yttlibrary/overlay/array.go
```diff
@@ -34,9 +34,12 @@ func (o Op) mergeArrayItem(
 	}
 
 	for _, leftIdx := range leftIdxs {
-		replace, err := o.apply(leftArray.Items[leftIdx].Value, newItem.Value, matchChildDefaults)
-		if err != nil {
-			return err
+		replace := true
+		if leftArray.Items[leftIdx].Value != nil {
+			replace, err = o.apply(leftArray.Items[leftIdx].Value, newItem.Value, matchChildDefaults)
```
cppforlife, line 39: it seems that apply is able to indicate when replace is necessary. any reason why not check nil value for left side within apply?

pivotaljohn: I had been thinking keeping this behavior specifically about the merge operation.

I'll take a second look.

pivotaljohn: I see. What was subtle to me was that:
- only the merge and assert operations continue the traversal
- only the merge operation is sensitive to `replace`

I like the simplicity of the null check in `apply()`.

👍🏼

cppforlife: this recursion does get a little mind bending. it's a little weird since inside apply we dont really look at left as a primary thing.

im starting to like your initial approach more since it doesnt raise a question inside apply why are we exiting early and acting on left instead of right.

pivotaljohn: > this recursion does get a little mind bending

Well, I'm glad to hear I'm not alone in that experience! 😌 

> im starting to like your initial approach more since it doesnt raise a question inside apply why are we exiting early and acting on left instead of right.

I find the first approach more readable and the second approach tighter.  In this case (and given that the probability of adding more operations to overlays is low, so reading is starting to become the overwhelmingly primary activity, here), I vote for the former.

cppforlife: 👍 let's roll with original approach.

PR #324, opened 2021-03-15
Skip processing plain YAML files marked as template

### LINKED ISSUE #232: Automatically detect when a YAML file is a template or plain YAML (obviating the `--ignore-unknown-comments` flag)
_Context: https://github.com/k14s/ytt/issues/14#issuecomment-726409172_

**I want to** provide supply a combination of plain YAML and ytt YAML templates
and have `ytt` automatically figure out which are `yaml-plain` and which are `yaml-template`
**so that** I can — without doing any extra work — a) easily include YAML from 3rd party sources; and b) get the full benefits of `ytt` ensuring my templates execute well

__

## Scenario: Mixed types of YAML files are correctly interpreted
**Given** at least one file that is plain YAML (e.g. includes standard YAML comments, `# `) and at least one file that is a valid `ytt` template (i.e. contains at least one annotation `#@`)
**When** I invoke `ytt` specifying those files with no additional flags
**Then** the output contains both the properly evaluated template output _and_ the plain YAML.

## Scenario: Using (now) deprecated flag
**When** I invoke `ytt` with the `--help` flag
**Then** I see that the `--ignore-unknown-comments` flag is deprecated.

**When** I invoke `ytt` with the `--ignore-unknown-comments`
**Then** I see a warning (on standard error) that informs me that this flag is deprecated.

**When** I search the docs with keywords like "yaml-plain", "yaml-template", "detection", or "comments" 
**Or** I navigate to sections that describe the "file mark" feature or that describe how templates are processed
**Then** I learn that `ytt` differentiates between "plain YAML" and "YAML templates", and how this behavior can be overiden.

pivotaljohn:
Addresses issue vmware-tanzu/carvel-ytt#232.

### DIFF
--- a/pkg/workspace/doc_extractor.go
+++ b/pkg/workspace/doc_extractor.go
@@ -17,8 +17,7 @@ const (
 )
 
 type DocExtractor struct {
-	DocSet    *yamlmeta.DocumentSet
-	MetasOpts yamltemplate.MetasOpts
+	DocSet *yamlmeta.DocumentSet
 }
 
 func (v DocExtractor) Extract(annName structmeta.AnnotationName) ([]*yamlmeta.Document,
@@ -37,6 +36,8 @@ func (v DocExtractor) Extract(annName structmeta.AnnotationName) ([]*yamlmeta.Do
 	return matchedDocs, nonMatchedDocs, nil
 }
 
+var ignoreUnknown = yamltemplate.MetasOpts{IgnoreUnknown: true}
+
 func (v DocExtractor) extract(docSet *yamlmeta.DocumentSet,
 	annName structmeta.AnnotationName) ([]*yamlmeta.Document, []*yamlmeta.Document, error) {
 
@@ -49,7 +50,7 @@ func (v DocExtractor) extract(docSet *yamlmeta.DocumentSet,
 		for _, meta := range doc.GetMetas() {
 			// TODO potentially use template.NewAnnotations(doc).Has(yttoverlay.AnnotationMatch)
 			// however if doc was not processed by the template, it wont have any annotations set
-			structMeta, err := yamltemplate.NewStructMetaFromMeta(meta, v.MetasOpts)
+			structMeta, err := yamltemplate.NewStructMetaFromMeta(meta, ignoreUnknown)
 			if err != nil {
 				return nil, nil, err
 			}
@@ -80,7 +81,7 @@ func (v DocExtractor) checkNonDocs(val interface{}, annName structmeta.Annotatio
 	}
 
 	for _, meta := range node.GetMetas() {
-		structMeta, err := yamltemplate.NewStructMetaFromMeta(meta, v.MetasOpts)
+		structMeta, err := yamltemplate.NewStructMetaFromMeta(meta, ignoreUnknown)
 		if err != nil {
 			return err
 		}
--- a/pkg/workspace/doc_extractor.go
+++ b/pkg/workspace/doc_extractor.go
@@ -36,8 +36,6 @@ func (v DocExtractor) Extract(annName structmeta.AnnotationName) ([]*yamlmeta.Do
 	return matchedDocs, nonMatchedDocs, nil
 }
 
-var ignoreUnknown = yamltemplate.MetasOpts{IgnoreUnknown: true}
-
 func (v DocExtractor) extract(docSet *yamlmeta.DocumentSet,
 	annName structmeta.AnnotationName) ([]*yamlmeta.Document, []*yamlmeta.Document, error) {
 
@@ -50,7 +48,7 @@ func (v DocExtractor) extract(docSet *yamlmeta.DocumentSet,
 		for _, meta := range doc.GetMetas() {
 			// TODO potentially use template.NewAnnotations(doc).Has(yttoverlay.AnnotationMatch)
 			// however if doc was not processed by the template, it wont have any annotations set
-			structMeta, err := yamltemplate.NewStructMetaFromMeta(meta, ignoreUnknown)
+			structMeta, err := yamltemplate.NewStructMetaFromMeta(meta, yamltemplate.MetasOpts{IgnoreUnknown: true})
 			if err != nil {
 				return nil, nil, err
 			}
@@ -81,7 +79,7 @@ func (v DocExtractor) checkNonDocs(val interface{}, annName structmeta.Annotatio
 	}
 
 	for _, meta := range node.GetMetas() {
-		structMeta, err := yamltemplate.NewStructMetaFromMeta(meta, ignoreUnknown)
+		structMeta, err := yamltemplate.NewStructMetaFromMeta(meta, yamltemplate.MetasOpts{IgnoreUnknown: true})
 		if err != nil {
 			return err
 		}
--- a/pkg/workspace/library_loader.go
+++ b/pkg/workspace/library_loader.go
@@ -14,7 +14,6 @@ import (
 	"github.com/k14s/ytt/pkg/schema"
 	"github.com/k14s/ytt/pkg/structmeta"
 	"github.com/k14s/ytt/pkg/yamlmeta"
-	"github.com/k14s/ytt/pkg/yamltemplate"
 )
 
 type LibraryLoader struct {
@@ -63,9 +62,7 @@ func (ll *LibraryLoader) Schemas() ([]*yamlmeta.Document, error) {
 			return nil, err
 		}
 
-		tplOpts := yamltemplate.MetasOpts{IgnoreUnknown: ll.templateLoaderOpts.IgnoreUnknownComments}
-
-		docs, _, err := DocExtractor{resultDocSet, tplOpts}.Extract(AnnotationSchemaMatch)
+		docs, _, err := DocExtractor{resultDocSet}.Extract(AnnotationSchemaMatch)
 		if err != nil {
 			return nil, err
 		}
@@ -116,10 +113,7 @@ func (ll *LibraryLoader) filesByAnnotation(annName structmeta.AnnotationName, lo
 				return nil, err
 			}
 
-			// we're only peeking at the document to determine its "type", no need to detect (yet) if contains unknowns.
-			tplOpts := yamltemplate.MetasOpts{IgnoreUnknown: true}
-
-			values, _, err := DocExtractor{docSet, tplOpts}.Extract(annName)
+			values, _, err := DocExtractor{docSet}.Extract(annName)
 			if err != nil {
 				return nil, err
 			}
--- a/pkg/workspace/library_loader.go
+++ b/pkg/workspace/library_loader.go
@@ -116,7 +116,8 @@ func (ll *LibraryLoader) filesByAnnotation(annName structmeta.AnnotationName, lo
 				return nil, err
 			}
 
-			tplOpts := yamltemplate.MetasOpts{IgnoreUnknown: ll.templateLoaderOpts.IgnoreUnknownComments}
+			// we're only peeking at the document to determine its "type", no need to detect (yet) if contains unknowns.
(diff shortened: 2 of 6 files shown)


pkg/files/file.go
```diff
@@ -225,14 +228,32 @@ func (r *File) IsTemplate() bool {
...
 
+func (r *File) hasTemplating() bool {
+	data, err := r.Bytes()
+	if err != nil {
+		panic(fmt.Sprintf("Unable to get contents of file %s: %s", r.OriginalRelativePath(), err))
+	}
+	switch r.Type() {
+	case TypeYAML:
+		for _, ann := range literal.AllAnnotations {
+			if bytes.Contains(data, ann) {
```
cppforlife: i think we are at the wrong abstraction level here. if somebody has a `foo: "#@overlay/match"` content this would pick this up as being a template, even though it is not.

pkg/files/file.go
```diff
@@ -225,14 +228,32 @@ func (r *File) IsTemplate() bool {
...
+				return true
+			}
+		}
+	case TypeText:
+		return bytes.Contains(data, literal.TextTemplateStartToken)
```
cppforlife: hmm... i am not sure we can do this for text (== always template by default) since there is no such thing as a comment in text.

in the line above, what if user only placed `@)` and screwed up opening tag?

pkg/files/file.go
```diff
@@ -4,11 +4,14 @@
...
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
 	"strings"
+
+	"github.com/k14s/ytt/pkg/literal"
```
cppforlife: feels a bit weird for the lowest layer to pick up annotations from the highest layers.

pkg/literal/annotations.go
```diff
@@ -0,0 +1,51 @@
...
+	SchemaMatchAnnotationName               = "schema/match"
+	SchemaNullableAnnotationName            = "schema/nullable"
+	TemplateCodeAnnotationName              = "template/code"
+	TemplateValueAnnotationName             = "template/value"
+	YamlMapKeyOverrideAnnotationName        = "yaml/map-key-override"
+	YamlTextTemplatedStringsAnnotationName  = "yaml/text-templated-strings"
+)
+
+var AllAnnotations = [...][]byte{
+	[]byte("#@" + CommentAnnotationShorthand),
```
cppforlife: #@ and (@ seems like details that are "owned" by other packages.

pkg/cmd/template/cmd.go
```diff
@@ -60,7 +60,7 @@ func NewCmd(o *Options) *cobra.Command {
 		RunE:    func(c *cobra.Command, args []string) error { return o.Run() },
 	}
 	cmd.Flags().BoolVar(&o.IgnoreUnknownComments, "ignore-unknown-comments", false,
-		"Configure whether unknown comments are considered as errors (comments that do not start with '#@' or '#!')")
+		"(deprecated: to explicitly disable processing, use `--file-mark <filename>:type=yaml-plain`) Configure whether unknown comments are considered as errors (comments that do not start with '#@' or '#!')")
```
cppforlife: this description is a bit confusing since it almost implies that this flag is deprecated and i should use file marks, even though the main change here is that folks dont need to do anything anymore. this also applies to the --implicit-map-key-overrides flag.

i wouldnt modify this description because it's still true as is for yaml templates.

pkg/files/file.go
pivotaljohn: I blame the beer. 🍺 

Yeah, this approach was way too literal (pun inadvertent) and subpar in pretty much every measure.

I see how I got fixated on the notion that `File.IsTemplate()` is _the_ source of truth about whether a file contains templating or not. Quite a bit of complexity flowed from that myoptia.

I see now that `files` is about user input: through a combination of filenames and file marks, indicating how they _intend_ a file to be processed.

Whether a given Document Set actually contains templating is squarely in the domain of `yamltemplate`. Leveraging that fact to adjust processing (i.e. skipping processing of files that contain nothing to process) is a decision well-made at the `workspace` level.

pkg/files/file.go
pivotaljohn: ... and there's no need to change how text files are processed. This wasn't asked for.

pkg/cmd/template/cmd.go
pivotaljohn: With this detection in place, what's the need for the `--ignore-unknown-comments` flag?

pkg/workspace/library_loader.go
```diff
@@ -116,7 +116,8 @@ func (ll *LibraryLoader) filesByAnnotation(annName structmeta.AnnotationName, lo
 				return nil, err
 			}
 
-			tplOpts := yamltemplate.MetasOpts{IgnoreUnknown: ll.templateLoaderOpts.IgnoreUnknownComments}
+			// we're only peeking at the document to determine its "type", no need to detect (yet) if contains unknowns.
+			tplOpts := yamltemplate.MetasOpts{IgnoreUnknown: true}
```
cppforlife: should we move this into DocExtractor?

pkg/workspace/template_loader.go
```diff
@@ -160,7 +160,7 @@ func (l *TemplateLoader) EvalYAML(libraryCtx LibraryExecutionContext, file *file
 	l.ui.Debugf("### ast\n")
 	docSet.Print(l.ui.DebugWriter())
 
-	if !file.IsTemplate() && !file.IsLibrary() {
+	if !(file.IsTemplate() && yamltemplate.IsTemplate(docSet)) && !file.IsLibrary() {
```
cppforlife: this is getting hard to understand. should we just add a separate if cond with return instead?

pkg/yamltemplate/template.go
```diff
@@ -34,6 +34,33 @@ type TemplateOpts struct {
...
 
+func IsTemplate(val interface{}) bool {
+	node, ok := val.(yamlmeta.Node)
+	if !ok {
+		return false
+	}
+
+	metaOpts := MetasOpts{IgnoreUnknown: true}
+	for _, meta := range node.GetMetas() {
+		structMeta, err := NewStructMetaFromMeta(meta, metaOpts)
```
cppforlife, line 49: it seems that contract right now for NewStructMetaFromMeta is to return errors for non-valid annotations. if we load it up with `MetasOpts{}` (== `IgnoreUnknown: false`) then we can just check on the error, right?

cppforlife: actually nm, this wouldnt work if there would be no comments, right...

cppforlife: i guess this loop looks a little "annoying" but i dont think we can do much about it.

pkg/yamltemplate/template.go
```diff
@@ -34,6 +34,33 @@ type TemplateOpts struct {
 	ImplicitMapKeyOverrides bool
 }
 
+func IsTemplate(val interface{}) bool {
```
cppforlife: it's a little weird to have `val interface{}`. may be have input as DocumentSet and have private isTemplate() function with interface? wdyt?

pivotaljohn: Yeah... I started with that exact structure... but it seemed a bit silly to have an exported function that's a one-liner that calls the unexported function that had all the logic.

Then on top, the first thing we do in the function is downcast that quietly terminates.

From a "when in Rome" perspective, in every other case I saw something like that, here, there was some kind of pre-condition or post-condition work that felt like it justified the separate functions of the same name. 

Perhaps given that it's an exported function is sufficient justification for having as typed of a signature as possible. I'll switch it back. Thanks for asking.

pkg/yamltemplate/template.go
pivotaljohn: > actually nm, this wouldnt work if there would be no comments, right...

I like your moxy. But, exactly.

> i guess this loop looks a little "annoying" but i dont think we can do much about it.

Yeah, it's also irritatingly identical to some linting logic we do in `DocExtractor`, but the duplication isn't tripping that threshold where the cost of the abstraction would be worth it, yet.

pkg/workspace/library_loader.go
pivotaljohn: Well played.

Yeah, DocExtractor being insensitive to "unknown comments" in exchange for a simpler interface there feels like a win.

pkg/workspace/template_loader.go
pivotaljohn: No so fast, mister... I think we can do it with a little boolean transformation and a more precise name for this new function...

```go
if !file.IsTemplate() && !file.IsLibrary() || !yamltemplate.HasTemplating(docSet) {
```

huh? huh?

pkg/cmd/template/cmd.go
cppforlife: this flag would still be used to ignore unknown comments within legit ytt templates if user wants that.

pkg/workspace/doc_extractor.go
```diff
@@ -37,6 +36,8 @@ func (v DocExtractor) Extract(annName structmeta.AnnotationName) ([]*yamlmeta.Do
 	return matchedDocs, nonMatchedDocs, nil
 }
 
+var ignoreUnknown = yamltemplate.MetasOpts{IgnoreUnknown: true}
```
cppforlife: should we just inline this? seems a bit weird to have workspace package own this little tidbit as top level var.

pkg/workspace/template_loader.go
```diff
@@ -160,7 +160,7 @@ func (l *TemplateLoader) EvalYAML(libraryCtx LibraryExecutionContext, file *file
 	l.ui.Debugf("### ast\n")
 	docSet.Print(l.ui.DebugWriter())
 
-	if !file.IsTemplate() && !file.IsLibrary() {
+	if !file.IsTemplate() && !file.IsLibrary() || !yamltemplate.HasTemplating(docSet) {
```
cppforlife, line 163: i was thinking this reads better:

```
if !file.IsTemplate() && !file.IsLibrary() {
    return nil, docSet, nil
}
if !yamltemplate.HasTemplating(docSet) {
    return nil, docSet, nil
}
```

pkg/workspace/doc_extractor.go
pivotaljohn: I mean, we can... I hear you about a package-level dangling var... I guess the fact that we're building the same value object again and again isn't too big a deal.

pkg/cmd/template/cmd.go
pivotaljohn: Why would they want this... like what's the use-case that says, "I authored this file, but I still want YAML comments"?

It invites failure states. What convenience does it afford?

pkg/workspace/template_loader.go
pivotaljohn: I respectfully disagree.

The way it's written in the PR, it's the same early exit for the same reason and it's all in one sentence:

"If the file is not a template... and it's not a library... Ooooor, it just doesn't have any templating in the first place... well, no need to 'evaluate' it."

pkg/cmd/template/cmd.go
cppforlife: i can think of one case: i have a third party yaml file with bunch of comments. i want to start using ytt templating in it so i add one ytt directive. without having to fix all the other comments _right this moment_ i want to start seeing ytt do its evaluation.

pkg/workspace/template_loader.go
cppforlife: ok lets keep it as is.

pkg/workspace/doc_extractor.go
cppforlife: your call. im ok either way with this.

> the fact that we're building the same value object again and again isn't too big a deal

yup, if we ever start optimizing ytt for perf i dont think this will show up on your radar.

PR #328, opened 2021-03-20
Add a video that explains the whole thing

adriens:
Add a video that explains the whole thing and its motivations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 # ytt
 
+- Discover `ytt` in [video](https://youtu.be/KbB5tI_g3bo)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,7 @@
 
 # ytt
 
+- Discover `ytt` in [video](https://youtu.be/KbB5tI_g3bo)
```
pivotaljohn: A real gem. @cppforlife looks so young in that video. (I kid)  Thanks for digging it up and sharing it here.

What do you think of pulling it just underneath the next line (keeping the "Play" the first item)? Or was your placing it at the top rather intentional?

adriens:
👍

PR #341, opened 2021-04-06
Update Left-side overlay positions

cari-lynn:
This work is to update the position (file name and line number) of left side node when a document is overlayed. It is needed to report right side schema error messages, and also shows up when using the `--output pos` flag.


gcheadle-vmware:
We've found that in general, starlark structs have unknown positions. The unknown position now surfaces (using `--output pos`) when overlaying starlark structs. Thinking about the scope of this PR, we've decided that changes to the positions of starlark struct do not belong in this PR. If this is something the team wants to add, we should write another story to be picked up. 

Starlark struct example:
```
#@ def return_starlark():
#@   return {"foo": "something"}
#@ end
```

cc: @cari-lynn

pivotaljohn (changes requested):
Reviewed together.

Notes:
- in `template_test.go` look tearing off an interface for the `AsBytes()` function; `interface{}` means we can't commit to a given interface.
- explore where the "[Single Point Of Truth](http://www.catb.org/~esr/writings/taoup/html/ch04s02.html#spot_rule)" could be for setting a value of a target node, given a replacing node.

cari-lynn:
Addressed @jtigger 's comments above.

We created a Node interface method to set the position. We did not update references to where we are setting the position directly (ie: map.Position) and did not make the field private since we noticed there is not a consistent pattern for other fields currently. Happy to hear opinions if you have any regarding it.

pkg/yamltemplate/filetests/ytt-library/overlay/array-output-position.yml
```diff
@@ -0,0 +1,74 @@
...
+test2
+--- #@ template.replace(overlay.apply(test2_left(), test2_right()))
+---
+test3
+--- #@ template.replace(overlay.apply(test3_left(), test3_right()))
+---
+
++++
+
+OUTPUT POSITION:          stdin:45 | [doc]
```
cppforlife: should we add a newline after OUTPUT POSITION? seems like it would read way better with it.

cari-lynn: Good idea 💥

pivotaljohn (approved):
Reviewed synchronously.

In terms of "Single Point of Truth" — thanks for looking into this. Seems fine to let this ride for a bit as we watch the schema feature settling in. The comments left there I think will help future readers understand the current intent and what might change that implementation.

PR #352, opened 2021-04-14
Support floats as a schema type

StevenLocke:


pkg/schema/type.go
```diff
@@ -107,6 +107,8 @@ func (a ArrayItemType) String() string {
 }
 func (m ScalarType) String() string {
 	switch m.Value.(type) {
+	case float64:
+		return "number"
```
cppforlife: any reason why not use `float`?

pkg/schema/type.go
```diff
@@ -177,11 +179,18 @@ func (m *ScalarType) CheckType(node yamlmeta.TypeWithValues) (chk yamlmeta.TypeC
...
-	case int:
-		if _, ok := m.Value.(int); !ok {
+	case float64:
+		if _, ok := m.Value.(float64); !ok {
 			chk.Violations = append(chk.Violations,
 				NewMismatchedTypeError(node, m))
 		}
+	case int:
+		if _, ok := m.Value.(int); !ok {
+			if _, ok = m.Value.(float64); !ok {
```
cppforlife, line 189: is this code saying that we accept float for type integer fields?

StevenLocke: The opposite, that we accept integers for float fields.

It looks a little weird, because we're switching on the type of the thing to check, not the schema. So we're saying if we got an int, make sure the schema is an int or a float.

pkg/schema/type.go
StevenLocke: Does go have a `float` type? I thought it only included float32 and float64. I'll check again :)

cppforlife: im talking about the string `"number"`, should probably be changed to `"float"` because ytt users do not know any concept called number.

StevenLocke: The objective with that wording was to make sure people would know that ints would be valid as well. 

Made a quick slack poll to see how folx feel:
https://kubernetes.slack.com/archives/CH8KCCKA5/p1618507762314100

cppforlife: i dont think there is major language out there that would not accept int value into a float type. e.g. in Go `var x float64 = 4` is totally valid.

pivotaljohn: I see the motivation to get to the most precise term and I appreciate that effort. Getting to precision is an important factor... and it needs to be contextualized in order to avoid a local optimization ...

So, stepping back, thinking of the overall user experience, if YAML has a type called "float" and Starlark has a type called "float", it is intuitive to call such a value a "float" here.

pkg/yamlmeta/node.go
```diff
@@ -37,6 +37,8 @@ func (s *Scalar) ValueTypeAsString() string       { return typeToString(s.Value)
 
 func typeToString(value interface{}) string {
 	switch value.(type) {
+	case float64:
+		return "number"
```
pivotaljohn: Ditto from: [comment above](https://github.com/vmware-tanzu/carvel-ytt/pull/352/files#r613647106).

pivotaljohn (changes requested):
The reviewer feedback is sufficiently strong, gonna mark this as "request changes" to make the state of this PR clear.

pivotaljohn (approved):
lgtm. Thank you, @StevenLocke!  A great way to send off!

PR #353, opened 2021-04-15
Type check command line data values

gcheadle-vmware:
Add tests to confirm that command line data values are being type checked against the schema.

### DIFF
--- a/pkg/cmd/template/schema_test.go
+++ b/pkg/cmd/template/schema_test.go
@@ -116,6 +116,51 @@ data_value: #@ data.values
 		assertSucceeds(t, filesToProcess, expected, opts)
 	})
 
+	t.Run("when a data value is passed using --data-value", func(t *testing.T) {
+		cmdOpts := cmdtpl.NewOptions()
+		cmdOpts.SchemaEnabled = true
+		schemaYAML := `#@schema/match data_values=True
+---
+foo: bar
+`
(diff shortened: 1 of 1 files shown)


pkg/cmd/template/schema_test.go
```diff
@@ -255,6 +300,31 @@ data_values.yml:6 |   - secure  #! expecting a map, got a string
 		assertFails(t, filesToProcess, expectedErr, opts)
 	})
 
+	t.Run("when a data value is passed using --data-value, but schema expects a non string", func(t *testing.T) {
```
cppforlife, line 303: i was hoping that part of this feature we would actually use schema's type to interpret the value. e.g.

if schema say field is of type int, then -v foo=3 should result in data.values.foo => 3 (integer).

pivotaljohn: I hear ya, but the story didn't request that. This story was to make sure that these inputs work properly with Schema, as is.

I see significant value in your expectation, but I feel like that's a chunk of work we might want to prioritize against other work...

Next step: write a story against the `--data-value...` flags that _does_ infer types when schema is present?

cppforlife: ok. do we have a separate issue for this then?

pivotaljohn: No. I'll create it now.

PR #363, opened 2021-04-19
Overlay multiple schema files

gcheadle-vmware:


cari-lynn:
Remaining items before the PR is ready to be reviewed:
- [x] extract duplicate code for dv and schema into another package
- [x] rename either package libraryref or type libraryref
- [x] Rename the types that were extracted from common dv and schema code
- [x] Does the 'common' dv and schema code belong in `template` package?
- [x] since we can use data values in text templates, we should reconsider not passing schema into `template_loader.go` for evalText and evalStarlark
- [x] 🆕  Support `#@ library/ref` annotation on schemas docs. 
  - [x] add tests
  - [x] if no schema file exists in a library, use the one annotated with `library/ref` as the base data values file.
- [x] Address TODO's from previous bullet point working on `library/ref`.
- [x] should the duplicated `UsedInLibrary` function have similar behaviour? Currently they do different things
- [x] in `library_loader.go` Values() takes “overlays” because of command-line flags. Does Schemas() need to take “overlays”?
- [x] does it make sense to pass in `nil` to `schemas()` function in root library? Or is there a better way to do that?
- [x] Do we need to have two interfaces for schema and for data values?

pkg/cmd/template/schema_test.go
```diff
@@ -933,6 +933,162 @@ foo: #@ data.values.foo
...
+	t.Run("when additional schema file is scoped to root library", func(t *testing.T) {
+		schemaYAML1 := `#@schema/match data_values=True
+---
+db_conn:
+- hostname: ""
+`
+
+		schemaYAML2 := `#@ load("@ytt:overlay", "overlay")
+#@schema/match data_values=True
+#@overlay/match by=overlay.all
```
cppforlife: something to think about... why do we need `#@overlay/match by=overlay.all` on schemas but not on data/values?

pkg/schema/schema.go
```diff
@@ -22,6 +22,9 @@ type DocumentSchema struct {
 	Source     *yamlmeta.Document
 	defaultDVs *yamlmeta.Document
 	Allowed    *DocumentType
+
+	originalLibRef []template.LibRefPiece
+	libRef         []template.LibRefPiece
```
cppforlife: is LibRefPiece really related to templating?

pkg/cmd/template/schema_test.go
cari-lynn: Good catch.. we don't need this annotation. This must've been a copy/paste mistake. Fixed.

pkg/schema/schema.go
pivotaljohn: This also came up in an in-person review we did, yesterday.  👍🏼

Finding/creating the right home for this (and related bits) is on the to-do list, in this issue/PR.

DennisDenuto: We agreed it felt 'off' being in template. We landed on `workspace/ref`.  There was sufficient logic in how libref's are parsed / handled.  We treat this new 'ref' package as a 'go' library used by schemas and datavalues.

DennisDenuto: resolving since this change is pretty outdated. `LibRefPiece` was also renamed to `LibRef`

pivotaljohn (approved):
LGTM!

PR #367, opened 2021-04-21
Add methods in url module to parse and trim credentials

### LINKED ISSUE #366: Add function to extract username and password from credential embedded URL in url module
[A description of the current challenge that you are experiencing.]
When handling a credential embedded URL in the form of `https://username:password@URL`, we want to parse the credential information like `username` and `password`, store them into a k8s secret. Also we want to get a URL that doesn't contain the credential so we can store it elsewhere.

[A clear and concise description of what you want to happen. If applicable a visual representation of the UX (ex: new CLI argument name, or resulting yaml).]

Involve following methods in `url` module:
1. `extract_username`: takes in a URL string, returns the embedded username as a string. If there's no credential, return `""`.
2. `extract_password`: takes in a URL string, returns the embedded password as a string. If there's no credential, return `""`.
3. `trim_credentials`: takes in a URL string, returns the URL that removes the credentials.

_(see also: thread in [#carvel](https://kubernetes.slack.com/archives/CH8KCCKA5/p1619026581416200))_

---
Vote on this request

hanlins:
Signed-off-by: Hanlin Shi <shihanlin9@gmail.com>

Introduce three methods in `url` module:
1. `extract_username`: to extract username in URL.
2. `extract_password`: to extract password in URL.
3. `trim_credentials`: to remove credentials in URL.

Closes #366.

### DIFF
--- a/pkg/yamltemplate/filetests/ytt-library/url.yml
+++ b/pkg/yamltemplate/filetests/ytt-library/url.yml
@@ -19,6 +19,16 @@ query_params:
   test3: #@ url.query_params_encode({"w":[]})
   test4: #@ url.query_params_decode("")
 
+parse_credentials:
+  test1: #@ url.extract_username("https://username:password@url.domain:8080")
+  test2: #@ url.extract_password("https://username:password@url.domain:8080")
+  test3: #@ url.trim_credentials("https://username:password@url.domain:8080")
+  test4: #@ url.extract_username("https://url.domain:8080")
+  test5: #@ url.extract_password("https://url.domain:8080")
+  test6: #@ url.trim_credentials("https://url.domain:8080")
+  test7: #@ url.extract_username("https://username@url.domain:8080")
+  test8: #@ url.extract_password("https://username@url.domain:8080")
+  test9: #@ url.trim_credentials("https://username@url.domain:8080")
 +++
 
 path_segment:
@@ -43,3 +53,13 @@ query_params:
     - ""
   test3: ""
   test4: {}
+parse_credentials:
+  test1: username
+  test2: password
+  test3: https://url.domain:8080
+  test4: ""
+  test5: ""
+  test6: https://url.domain:8080
+  test7: username
+  test8: ""
+  test9: https://url.domain:8080
--- a/pkg/yttlibrary/url.go
+++ b/pkg/yttlibrary/url.go
@@ -27,6 +27,10 @@ var (
 
 				"query_params_encode": starlark.NewBuiltin("url.query_params_encode", core.ErrWrapper(urlModule{}.QueryParamsEncode)),
 				"query_params_decode": starlark.NewBuiltin("url.query_params_decode", core.ErrWrapper(urlModule{}.QueryParamsDecode)),
+
+				"extract_username": starlark.NewBuiltin("url.extract_username", core.ErrWrapper(urlModule{}.ExtractUsername)),
+				"extract_password": starlark.NewBuiltin("url.extract_password", core.ErrWrapper(urlModule{}.ExtractPassword)),
+				"trim_credentials": starlark.NewBuiltin("url.trim_credentials", core.ErrWrapper(urlModule{}.TrimCredentials)),
 			},
 		},
 	}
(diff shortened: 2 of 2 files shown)


pivotaljohn:
> /hold on the discussion of the url module interface in the issue.

i.e. _this_ discussion: https://kubernetes.slack.com/archives/CH8KCCKA5/p1619033548435100?thread_ts=1619026581.416200&cid=CH8KCCKA5

pivotaljohn:
Thanks for the thorough start here, @hanlins.

I'm thinking about an API that would be generalized, compact, and intuitive.

Looking around for how others have solved this same need, I see a number of standard libraries using the technique of parsing a string into an object that is then interrogated for the desired part(s). For example: https://golang.org/pkg/net/url/#Parse

FWIW, there's prior art within `ytt` for introducing new value types into the system: https://github.com/vmware-tanzu/carvel-ytt/blob/bc8782e60d1089a4781060d36811fea5bbbb608b/pkg/workspace/library_module.go#L128

Thoughts?

pivotaljohn (changes requested):
Marking the PR as "Request Changes" to clearly indicate we're having a design discussion, right now.

hanlins:
Hi @pivotaljohn, thanks for your replies! So do you mean instead of adding new methods ad-hoc into the `url` module, we should introduce a `url` type instead?

pivotaljohn:
> Hi @pivotaljohn, thanks for your replies! So do you mean instead of adding new methods ad-hoc into the `url` module, we should introduce a `url` type instead?

That is what I'm suggesting, yes. 👍🏼

... and I am intending this to be a discussion.

In the process I see something like this:

1. identify a want/need ✔️ 
1. think big picture about what new/extended capability we could add that would not only solve for this use case, but the class of needs, generally.
1. converge on a solution that results in the most resilient design:
    - keeps interface and implementation complexity to only what's essential
    - keeps the API compact and features as orthogonal as possible
1. run with it, collaborating on the lower-level design decisions along the way

So, I'm jumping into step 2, there. Bringing up an idea and trying to "show my work" by articulating a rationale for that approach.

At this point, there may be other ideas. _Anyone_ can chime in here. And we can bat around other thoughts.

hanlins:
Hi @pivotaljohn, this seems interesting, it seems to introduce a module-local type instead of a base type, is that correct? I think for a starting point, golang provides a nice implementation for URL already (https://github.com/golang/go/blob/e6ac2df2b198f583780277a7cf96e3b0b61fe0a1/src/net/url/url.go#L340-L369), maybe we parse a string to the golang URL and add methods to support existing use cases and add the new use cases we want to support.

pivotaljohn:
> Hi @pivotaljohn, this seems interesting, it seems to introduce a module-local type instead of a base type, is that correct? I think for a starting point, golang provides a nice implementation for URL already (https://github.com/golang/go/blob/e6ac2df2b198f583780277a7cf96e3b0b61fe0a1/src/net/url/url.go#L340-L369), maybe we parse a string to the golang URL and add methods to support existing use cases and add the new use cases we want to support.

That sounds really enticing: less code for us to maintain (and make sure we get right). So long as the contract with Go's `url.URL` doesn't break, that should be just fine... right?


... another factor we need to keep in mind is ensuring that we maintain the sandbox.
`ytt` very consciously prevents any access to the outside world from within templates: no network, no disk, no time, no RNG, no encryption.

We do this for two reasons:
1. so that the tool is safe to use — sensitive data likely flows through here. Last think you need as an operator is for the deployment tool chain to exfiltrate secrets.
2. so that when the same inputs are given, the _exact_ same output is rendered — having this guarantee allows operators to gate potentially expensive operations so that they only execute if there was actually a change.

So, we'll want to make sure we don't inadvertently provide access to the outside world.

I'm not saying that I've spotted something like that yet, just putting it on your radar.

pivotaljohn:
Are there any other viable (and meaningfully different) approaches we should consider?

hanlins:
Hi @pivotaljohn, I'm trying to wrap these ideas in my mind so correct me if I'm wrong, I'm trying to understand the https://github.com/vmware-tanzu/carvel-ytt/blob/bc8782e60d1089a4781060d36811fea5bbbb608b/pkg/workspace/library_module.go#L128 example you used above.
So, what we've been talking about is to introduce a new type namely `urlValue` in the `url` module. In the `url` module, we can have a method namely `parse` to parse the URL, and what it returns is a starlark value, which is of `Module` type. And we still need to selectively expose the methods that this object, is that correct?

hanlins:
Hi @pivotaljohn, I created https://github.com/vmware-tanzu/carvel-ytt/pull/372 which aggregates the methods into a `url` type, PTAL. If that is our direction then I'll close this one.

hanlins:
Closing this changeset in favor of https://github.com/vmware-tanzu/carvel-ytt/pull/372.

PR #372, opened 2021-04-27
Add url type in the url module

### LINKED ISSUE #366: Add function to extract username and password from credential embedded URL in url module

hanlins:
Signed-off-by: Hanlin Shi <shihanlin9@gmail.com>

Introduce type `url` in `url` module, which has three methods:

1. user.name: to extract username in URL.
2. user.password: to extract password in URL.
3. user: return the colon separated username+password tuple. (if password is not present, then just return username).
4. without_user(): generate a URL object that removes user credential information.
5. string(): returns the string representation of the URL object.

Closes #366.

### DIFF
--- a/pkg/template/core/starlark_value.go
+++ b/pkg/template/core/starlark_value.go
@@ -11,7 +11,7 @@ import (
 )
 
 type StarlarkValueToGoValueConversion interface {
-	AsGoValue() (interface{}, error)
+	AsGoValue() interface{}
 }
 
 type StarlarkValue struct {
@@ -22,8 +22,8 @@ func NewStarlarkValue(val starlark.Value) StarlarkValue {
 	return StarlarkValue{val}
 }
 
-func (e StarlarkValue) AsGoValue() (interface{}, error) {
-	return e.asInterface(e.val), nil
+func (e StarlarkValue) AsGoValue() interface{} {
+	return e.asInterface(e.val)
 }
 
 func (e StarlarkValue) AsString() (string, error) {
@@ -53,11 +53,7 @@ func (e StarlarkValue) AsInt64() (int64, error) {
 
 func (e StarlarkValue) asInterface(val starlark.Value) interface{} {
 	if obj, ok := val.(StarlarkValueToGoValueConversion); ok {
-		res, err := obj.AsGoValue()
-		if err != nil {
-			panic(err)
-		}
-		return res
+		return obj.AsGoValue()
 	}
 
 	switch typedVal := val.(type) {
@@ -116,6 +112,13 @@ func (e StarlarkValue) dictAsInterface(val *starlark.Dict) interface{} {
 }
 
 func (e StarlarkValue) structAsInterface(val *StarlarkStruct) interface{} {
+	if val.represent != nil {
+		repVal, err := val.represent()
+		if err != nil {
+			panic(err)
+		}
+		return e.asInterface(repVal)
+	}
 	// TODO accessing privates
 	result := orderedmap.NewMap()
 	val.data.Iterate(func(k, v interface{}) {
--- a/pkg/template/core/starlark_value.go
+++ b/pkg/template/core/starlark_value.go
@@ -11,7 +11,7 @@ import (
 )
 
 type StarlarkValueToGoValueConversion interface {
-	AsGoValue() interface{}
+	AsGoValue() (interface{}, error)
 }
 
 type StarlarkValue struct {
@@ -22,8 +22,8 @@ func NewStarlarkValue(val starlark.Value) StarlarkValue {
 	return StarlarkValue{val}
 }
 
-func (e StarlarkValue) AsGoValue() interface{} {
-	return e.asInterface(e.val)
+func (e StarlarkValue) AsGoValue() (interface{}, error) {
+	return e.asInterface(e.val), nil
 }
 
 func (e StarlarkValue) AsString() (string, error) {
@@ -53,7 +53,11 @@ func (e StarlarkValue) AsInt64() (int64, error) {
 
 func (e StarlarkValue) asInterface(val starlark.Value) interface{} {
 	if obj, ok := val.(StarlarkValueToGoValueConversion); ok {
-		return obj.AsGoValue()
+		res, err := obj.AsGoValue()
+		if err != nil {
+			panic(err)
+		}
+		return res
 	}
 
 	switch typedVal := val.(type) {
@@ -112,13 +116,6 @@ func (e StarlarkValue) dictAsInterface(val *starlark.Dict) interface{} {
 }
 
 func (e StarlarkValue) structAsInterface(val *StarlarkStruct) interface{} {
-	if val.represent != nil {
-		repVal, err := val.represent()
-		if err != nil {
-			panic(err)
-		}
-		return e.asInterface(repVal)
-	}
 	// TODO accessing privates
 	result := orderedmap.NewMap()
 	val.data.Iterate(func(k, v interface{}) {
--- a/pkg/template/core/starlark_value.go
+++ b/pkg/template/core/starlark_value.go
@@ -119,13 +119,6 @@ func (e StarlarkValue) dictAsInterface(val *starlark.Dict) interface{} {
 }
 
 func (e StarlarkValue) structAsInterface(val *StarlarkStruct) interface{} {
-	if val.represent != nil {
-		repVal, err := val.represent()
-		if err != nil {
-			panic(err)
-		}
-		return e.asInterface(repVal)
-	}
 	// TODO accessing privates
 	result := orderedmap.NewMap()
 	val.data.Iterate(func(k, v interface{}) {
--- a/pkg/template/core/starlark_value.go
+++ b/pkg/template/core/starlark_value.go
@@ -14,6 +14,10 @@ type StarlarkValueToGoValueConversion interface {
 	AsGoValue() interface{}
 }
 
+type UnconvertableStarlarkValue interface {
+	ConversionHint() string
+}
+
 type StarlarkValue struct {
 	val starlark.Value
 }
@@ -52,6 +56,9 @@ func (e StarlarkValue) AsInt64() (int64, error) {
 }
 
 func (e StarlarkValue) asInterface(val starlark.Value) interface{} {
+	if obj, ok := val.(UnconvertableStarlarkValue); ok {
+		panic(obj.ConversionHint())
+	}
 	if obj, ok := val.(StarlarkValueToGoValueConversion); ok {
 		return obj.AsGoValue()
 	}
--- a/pkg/template/core/starlark_value.go
+++ b/pkg/template/core/starlark_value.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 
 	"github.com/k14s/starlark-go/starlark"
-	"github.com/k14s/starlark-go/starlarkstruct"
 	"github.com/k14s/ytt/pkg/orderedmap"
 )
 
@@ -96,9 +95,6 @@ func (e StarlarkValue) asInterface(val starlark.Value) interface{} {
 	case *starlark.Set:
 		return e.itearableAsInterface(typedVal)
 
-	case *starlarkstruct.Struct:
-		return e.asInterface(typedVal.Constructor())
-
 	default:
 		panic(fmt.Sprintf("unknown type %T for conversion to go value", val))
 	}
@@ -116,6 +112,13 @@ func (e StarlarkValue) dictAsInterface(val *starlark.Dict) interface{} {
 }
 
 func (e StarlarkValue) structAsInterface(val *StarlarkStruct) interface{} {
+	if val.represent != nil {
+		repVal, err := val.represent()
+		if err != nil {
+			panic(err)
+		}
+		return e.asInterface(repVal)
+	}
 	// TODO accessing privates
 	result := orderedmap.NewMap()
 	val.data.Iterate(func(k, v interface{}) {
--- a/pkg/template/core/starlark_value.go
+++ b/pkg/template/core/starlark_value.go
@@ -97,7 +97,7 @@ func (e StarlarkValue) asInterface(val starlark.Value) interface{} {
 		return e.itearableAsInterface(typedVal)
 
 	case *starlarkstruct.Struct:
-		return e.nativeStructAsInterface(typedVal)
+		return e.asInterface(typedVal.Constructor())
 
 	default:
 		panic(fmt.Sprintf("unknown type %T for conversion to go value", val))
@@ -111,19 +111,6 @@ func (e StarlarkValue) dictAsInterface(val *starlark.Dict) interface{} {
 			panic("dict item is not KV")
 		}
 		result.Set(e.asInterface(item.Index(0)), e.asInterface(item.Index(1)))
-	}
-	return result
-}
-
-func (e StarlarkValue) nativeStructAsInterface(val *starlarkstruct.Struct) interface{} {
-	// struct's ToStringDict uses map, hence ordering is not deterministic
-	result := orderedmap.NewMap()
-	for _, key := range val.AttrNames() {
-		v, err := val.Attr(key)
-		if err != nil {
-			panic("expected Attr() to succeed for *starlarkstruct.Struct")
-		}
-		result.Set(key, e.asInterface(v))
 	}
 	return result
 }
--- a/pkg/yamltemplate/filetests/ytt-library/url-value.yml
+++ b/pkg/yamltemplate/filetests/ytt-library/url-value.yml
@@ -5,13 +5,13 @@ url_value:
   #@ testURL2 = testURL1.without_user()
   #@ testURL3 = url.parse("http://admin@url.domain:8080")
   test1: #@ testURL1.string()
-  test2: #@ testURL1.user.string()
+  test2: #@ testURL1.user
   test3: #@ testURL1.user.name
   test4: #@ testURL1.user.password
   test5: #@ testURL2.string()
   test6: #@ testURL2.user
   test7: #@ testURL3.string()
-  test8: #@ testURL3.user.string()
+  test8: #@ testURL3.user
   test9: #@ testURL3.user.name
   test10: #@ testURL3.user.password
 
--- a/pkg/yamltemplate/filetests/ytt-library/url-value.yml
+++ b/pkg/yamltemplate/filetests/ytt-library/url-value.yml
@@ -1,0 +1,30 @@
+#@ load("@ytt:url", "url")
+
+url_value:
+  #@ testURL1 = url.parse("http://admin:password@url.domain:8080")
+  #@ testURL2 = testURL1.without_user()
+  #@ testURL3 = url.parse("http://admin@url.domain:8080")
+  test1: #@ testURL1.string()
+  test2: #@ testURL1.user
+  test3: #@ testURL1.user.name
+  test4: #@ testURL1.user.password
+  test5: #@ testURL2.string()
+  test6: #@ testURL2.user
+  test7: #@ testURL3.string()
+  test8: #@ testURL3.user
+  test9: #@ testURL3.user.name
+  test10: #@ testURL3.user.password
+
++++
+
+url_value:
+  test1: http://admin:password@url.domain:8080
+  test2: admin:password
+  test3: admin
+  test4: password
+  test5: http://url.domain:8080
+  test6: null
+  test7: http://admin@url.domain:8080
+  test8: admin
+  test9: admin
+  test10: null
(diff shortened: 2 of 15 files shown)


pkg/yttlibrary/url.go
```diff
@@ -178,3 +185,58 @@ func (b urlModule) sortedKeys(vals url.Values) []string {
...
+	return (&urlValue{parsedURL}).AsStarlarkValue(), nil
+}
+
+func (uv *urlValue) AsStarlarkValue() starlark.Value {
+	return &starlarkstruct.Module{
+		Name: "url",
+		Members: starlark.StringDict{
+			"username":         starlark.NewBuiltin("url.username", core.ErrWrapper(uv.Username)),
+			"password":         starlark.NewBuiltin("url.password", core.ErrWrapper(uv.Password)),
+			"trim_credentials": starlark.NewBuiltin("url.trim_credentials", core.ErrWrapper(uv.TrimCredentials)),
```
pivotaljohn: Where there are domain-specific canonical terms for things, it can be efficient to reuse those terms rather than coin our own.

I see the corresponding spec — [RFC 2396 § 3.2.2](https://www.rfc-editor.org/rfc/rfc2396.html#section-3.2.2) — the name for this combination is: "userinfo". The Go implementation exposes a field called `User`.

What might motivate us to do something different?

Suggestion: `.without_user()`

pkg/yttlibrary/url.go
```diff
@@ -178,3 +185,59 @@ func (b urlModule) sortedKeys(vals url.Values) []string {
...
+	parsedURL, err := url.Parse(urlStr)
+	if err != nil {
+		return starlark.None, err
+	}
+
+	return (&urlValue{parsedURL}).AsStarlarkValue(), nil
+}
+
+func (uv *urlValue) AsStarlarkValue() starlark.Value {
+	return &starlarkstruct.Module{
```
pivotaljohn: This value feels to me more like a `starlarkstruct.Struct` than a `starlarkstruct.Module`: `urlValue` _is_ a data structure.

I do see that `libraryValue` has been implemented as a `module`...
https://github.com/vmware-tanzu/carvel-ytt/blob/bc8782e60d1089a4781060d36811fea5bbbb608b/pkg/workspace/library_module.go#L143-L153

Yet, while it's all methods, it too is a value, not a result of an evaluated Starlark Module.

That `TODO` on line 143 is legit: this isn't a pattern we want to propagate.

...

There's a "constructor" for `Structs`:
https://github.com/vmware-tanzu/carvel-ytt/blob/bc8782e60d1089a4781060d36811fea5bbbb608b/vendor/github.com/k14s/starlark-go/starlarkstruct/struct.go#L50-L52

And `Builtin`s (i.e. a function implemented in Go) can be bound to any Starlark value:
https://github.com/vmware-tanzu/carvel-ytt/blob/bc8782e60d1089a4781060d36811fea5bbbb608b/vendor/github.com/k14s/starlark-go/starlark/value.go#L663-L676

pkg/yttlibrary/url.go
```diff
@@ -178,3 +185,59 @@ func (b urlModule) sortedKeys(vals url.Values) []string {
...
+		passwd = ""
+	}
+	return starlark.String(passwd), nil
+}
+
+func (uv urlValue) TrimCredentials(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() != 0 {
+		return starlark.None, fmt.Errorf("expected no argument")
+	}
+	urlVar := *uv.URL
```
pivotaljohn: Appreciate the defensive copy! 👍🏼  Values are more reliable when immutable.

pkg/yttlibrary/url.go
```diff
@@ -178,3 +185,59 @@ func (b urlModule) sortedKeys(vals url.Values) []string {
...
+	urlVar.User = nil
+	return starlark.String(urlVar.String()), nil
```
pivotaljohn: What do you think about returning a `urlValue`, here, instead?

It keeps open the possibility of a fluent interface.
It would likely be useful to generally provide a `String()` for `urlValue`.

pivotaljohn (changes requested):
This is a significant leveling-up from #367. Thank you for putting in the extra effort to work out how to provide a value.

Unfortunately, the prior art around that work within the `ytt` code is slightly off from what I believe we want here: a value rather than a code module (see comments, below). So, there's a little adjustment to the actual Starlark data structures we'd use to present these `url` values.

pkg/template/core/starlark_value.go
```diff
@@ -115,19 +115,6 @@ func (e StarlarkValue) dictAsInterface(val *starlark.Dict) interface{} {
...
-	for _, key := range val.AttrNames() {
-		v, err := val.Attr(key)
-		if err != nil {
-			panic("expected Attr() to succeed for *starlarkstruct.Struct")
-		}
-		result.Set(key, e.asInterface(v))
-	}
-	return result
-}
-
```
pivotaljohn, line 130: Yes, this is legitimately dead code. Let's keep this deletion. 👍🏼 

- `starlarkstruct.Struct` was used in the very beginning.
- but then the maintainers found it lacking in important ways — mostly because it's not iterable, but others as well... something that `ytt` really needs.
- it was an oversight that this `case` was not pruned.

pkg/yttlibrary/url.go
```diff
@@ -178,3 +185,55 @@ func (b urlModule) sortedKeys(vals url.Values) []string {
...
+		return starlark.None
+	}
+	return starlark.String(passwd)
+}
+
+func (uv *urlValue) withoutUser(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() != 0 {
+		return starlark.None, fmt.Errorf("expected no argument")
+	}
+	urlVar := *uv.URL
```
pivotaljohn:
```suggestion
      // sufficient "deep copy" because we set the only pointer in the structure to `nil`
	urlVar := *uv.URL
```

pkg/yamltemplate/filetests/ytt-library/url.yml
```diff
@@ -19,6 +19,21 @@ query_params:
...
+  test2: #@ testURL1.user
+  test3: #@ testURL1.user.name
+  test4: #@ testURL1.user.password
+  test5: #@ testURL2
+  test6: #@ testURL2.user
+  test7: #@ testURL3
+  test8: #@ testURL3.user
+  test9: #@ testURL3.user.name
+  test10: #@ testURL3.user.password
+
```
pivotaljohn:
```suggestion
url_type:
  #@ testURL1 = url.parse("http://admin:password@url.domain:8080")
  #@ testURL2 = testURL1.without_user()
  #@ testURL3 = url.parse("http://admin@url.domain:8080")
  test1: #@ testURL1.string()
  test2: #@ testURL1.user
  test3: #@ testURL1.user.name
  test4: #@ testURL1.user.password
  test5: #@ testURL2.string()
  test6: #@ testURL2.user
  test7: #@ testURL3.string()
  test8: #@ testURL3.user
  test9: #@ testURL3.user.name
  test10: #@ testURL3.user.password

```

As noted in the summary comment, let's follow your initial idea of _not_ inferring a string encoding from a "naked" instance.

This means that there's a cryptic error message complaining "`...unknown type *starlark.Builtin for conversion to go value...`" 
... it's _this_ situation that makes me think we'd get to where we want to go faster and with the needed control if we "rolled our own" out the gate.

pivotaljohn (commented):
Seems like a design conversation is in order... but here are some notes...

## The Tale of Two Structs

So, today I learned:
- `ytt` has not been using `starlarkstruct.Struct`... despite that spot of dead code (the very same you expunged in this PR)
- in fact, we've rolled our own: `core.StarlarkStruct` ... for reasons.
- given the set of specific needs, I'm starting to think that starting out with a custom, yet congruent `struct` impl. might be the best move, here...?

and, it would be a bad idea to have two implementations of `struct` floating around in the runtime... it's just asking for strange behavior.

Apologies for inadvertently misleading you to the defunct implementation. 🙏🏼 


## To String or Not To String

`urlValue` is actually the first time we're hatching a "higher-level" type within `ytt`'s flavor of Starlark. All the other types are either primitives or generic data structures.

I got a little over zealous in my desire for a "great UX" when I suggested that a naked `urlValue` instance should imply the stringified form.
- part of the problem here is that for other similar higher-level types, this may not be the right behavior.
- ... and in other (albiet rarer) use-cases for `urlValue`, the structure form might be what's desired, anyway.

So, better to defer making that inference to a time when we can make a better informed decision (i.e. after some use and ideally a couple more higher-level types). 

Concretely, yes, let's include a `.string()` as you had originally designed.

_(details, below)_

pkg/yamltemplate/filetests/ytt-library/url.yml
hanlins: I think we want to make the following changes after having a discussion last Friday:
1. Instead of using the legacy `starlarkstruct.Struct`, we want to converge to `StarlarkStruct`.
2. Instead of returning a confusing error message `...unknown type *starlark.Builtin for conversion to go value...` when referencing `testURL` directly, we want to see things like `testURL cannot be directly referenced, please use testURL.string()`, something more instructive like that.
3. As @jtigger mentioned, involve a new method `string()` to return the string representation of the URL. We do this   1). to accommodate the potential use case that we might need to return an encoded string that contains Unicode characters, 2). we might introduce more module-level types later on, and we might not always want to depend on the struct constructor. We need to have some flexibility for those types of representations.

I'll try and see if it's feasible to address all these requirements in a single changeset.

pkg/yamltemplate/filetests/ytt-library/url.yml
```diff
@@ -19,6 +19,21 @@ query_params:
   test3: #@ url.query_params_encode({"w":[]})
   test4: #@ url.query_params_decode("")
 
+url_type:
```
pivotaljohn: (nit) the test failure message is getting kinda hard to see what went wrong; would you mind pulling this into a separate test file, please? (e.g. `url-value.yml`)?

pkg/yttlibrary/url.go
```diff
@@ -178,3 +185,76 @@ func (b urlModule) sortedKeys(vals url.Values) []string {
...
+	m.Set("user", uv.user())
+	m.Set("without_user", starlark.NewBuiltin("url.without_user", core.ErrWrapper(uv.withoutUser)))
+	m.Set("string", starlark.NewBuiltin("url.string", core.ErrWrapper(uv.string)))
+	s := core.NewStarlarkStruct(m)
+	s.SetRepresentation(urlRepresent)
+	return s
+}
+
+func urlRepresent() (starlark.Value, error) {
+	return nil, fmt.Errorf("urlValue cannot be directly referenced, please use urlValue.string()")
```
pivotaljohn: (nit)

```suggestion
	return nil, fmt.Errorf("urlValue: cannot coerce to a string; use .string()")
```

Attempting to follow the error message formatting I'm seeing elsewhere in Starlark/ytt.

pivotaljohn (commented):
This is fantastic, Hanlin!

I love the move of keeping all of what's common for custom type values while providing the extension point for what will vary. 👍🏼 

I need to reflect on the co-existence of the `represent` feature alongside the `StarlarkValueToGoValueConversion`. I _think_ it's okay, but I want to sleep on it.

Other than that, I'm really liking where this is landing.

hanlins:
> This is fantastic, Hanlin!
> 
> I love the move of keeping all of what's common for custom type values while providing the extension point for what will vary. 👍🏼
> 
> I need to reflect on the co-existence of the `represent` feature alongside the `StarlarkValueToGoValueConversion`. I _think_ it's okay, but I want to sleep on it.
> 
> Other than that, I'm really liking where this is landing.

Hi @jtigger, thanks for the feedbacks! I think the tricky part for `StarlarkValueToGoValueConversion` is, as for now, it doesn't provide a way for people to emit errors. So is it ok if I refactor `StarlarkValueToGoValueConversion` the interface to allow it return error? If so, then I can make it return error such as `"urlValue: cannot coerce to a string; use .string()"`.

pkg/yamltemplate/filetests/ytt-library/url.yml
hanlins: There are no negative cases here. I can move the test cases to a separate file but I don't know how to add negative test cases.

pivotaljohn: Thanks for making that move, Hanlin.

The only negative case I can see is the representation error message. All the other cases are covered by a combination of tests I expect around `url.URL` and other ways we've confirmed that errors occurring through this code path make their way to the user.

FWIW, there _is_ a way to specify an error message in this parameterized test setup. The check is made here:
https://github.com/vmware-tanzu/carvel-ytt/blob/a1fa9e9bf578879c932ef360457c411f43ec2535/pkg/yamltemplate/template_test.go#L75-L82

And here's an example of it in use:
https://github.com/vmware-tanzu/carvel-ytt/blob/e850b3716926fae85418c652e50f4dcc42d0ca0d/pkg/yamltemplate/filetests/datatypes-err.yml#L9-L14

That said, I'd accept and merge this PR _without_ a specific test for this case.

pivotaljohn (approved):
LGTM. Thank you for all your work, here, Hanlin. We're not just adding a new feature, but setup the codebase to accept other custom types. 👍🏼

pkg/template/core/starlark_value.go
```diff
@@ -54,7 +53,11 @@ func (e StarlarkValue) AsInt64() (int64, error) {
 
 func (e StarlarkValue) asInterface(val starlark.Value) interface{} {
 	if obj, ok := val.(StarlarkValueToGoValueConversion); ok {
-		return obj.AsGoValue()
+		res, err := obj.AsGoValue()
+		if err != nil {
+			panic(err)
```
pivotaljohn: Regardless of how you decide that question "`represent` vs. `StarlarkValueToGoValueConversion` vs. ....?" (summarized above), seems like we can provide a significantly better experience to the end user if we return an `error` rather than panic out of `StarlarkValue.asInterface()`.

We'd love to preserve the panic for situations where we truly need to abort rather than as an error handling technique.

`AsGoValue()` already returns an `error`, so I think most of the plumbing is in place from there.

Doing so means that the internal Go stack trace will just go away, leaving the user with a short (and easier to read) error message.

pkg/yamltemplate/filetests/ytt-library/url-value.yml
```diff
@@ -0,0 +1,30 @@
...
+  test1: #@ testURL1.string()
+  test2: #@ testURL1.user
+  test3: #@ testURL1.user.name
+  test4: #@ testURL1.user.password
+  test5: #@ testURL2.string()
+  test6: #@ testURL2.user
+  test7: #@ testURL3.string()
+  test8: #@ testURL3.user
+  test9: #@ testURL3.user.name
+  test10: #@ testURL3.user.password
```
pivotaljohn:
```suggestion
  test1: #@ testURL1.string()
  test2: #@ testURL1.user.string()
  test3: #@ testURL1.user.name
  test4: #@ testURL1.user.password
  test5: #@ testURL2.string()
  test6: #@ testURL2.user.string()
  test7: #@ testURL3.string()
  test8: #@ testURL3.user.string()
  test9: #@ testURL3.user.name
  test10: #@ testURL3.user.password
```

I created an inconsistency, here. The rationale that we came to for wanting to require that `urlValue` to no assume an encoding... well, not sure why it didn't occur to me that the same is true for `urlValue.user` 😬 .

pkg/yttlibrary/url.go
```diff
@@ -27,13 +27,23 @@ var (
...
+				"parse": starlark.NewBuiltin("url.parse", core.ErrWrapper(urlModule{}.ParseURL)),
 			},
 		},
 	}
 )
 
 type urlModule struct{}
 
+// urlValue stores a parsed URL
+type urlValue struct {
```
pivotaljohn: _(optional)_ in this project, we're aiming to being careful in our API design _and_ defaulting to exporting primary types.  Given that the `urlValue` is effectively part of the `ytt` Starlark programming context, it's generally useful to export the same type in the Go code.

Suggestion: export `urlValue`

pkg/yttlibrary/url.go
```diff
@@ -27,13 +27,23 @@ var (
...
+	url *url.URL
+	*core.StarlarkStruct
+}
+
+type urlUser urlValue
```
pivotaljohn: _(optional)_

is `urlUser` _actually_ a whole `urlValue`?
Would it make sense for it to hold a `*url.UserInfo`?

pivotaljohn (changes requested):
## `represent` vs. `StarlarkValueToGoValueConversion` vs. ....?

Talked about this with @cppforlife and we arrived at what I thought was a cool compromise:

Given that the only error we'd be "throwing" is that the type is not convertible, what if we introduced an interface that specifically indicates the type is not convertible?

Something like:
```go
type UnconvertableStarlarkValue interface {
	ConversionHint() string
}
```

So, instead of muddying up `StarlarkValueToGoValueConversion` with an `error` (and all the error checks), a type would instead implement the `UnconvertableStarlarkValue` interface. `StarlarkValue.asInterface()` detects this interface and returns an `error` with part of the error message coming from `ConversionHint()`.

Personally, I'm digging how this makes the code more explicit about what's going on: that there are some types that are just not convertible... and that _those_ types ought to give contextualized hints/help in _how_ they might be converted.

Thoughts?

...

A few more comments in-line, below.

Note: a few thing I've marked as "_(optional)_". I'm sharing these as observations; I'm gonna leave it up to you whether you agree and want to include that change. _It's okay if you don't._

pkg/yttlibrary/url.go
hanlins: I see. So apart of `UrlValue`, should we also export `urlUser` which is also a starlark type?

pkg/yttlibrary/url.go
hanlins: Sure, fixed.

hanlins:
> ## `represent` vs. `StarlarkValueToGoValueConversion` vs. ....?
> 
> Talked about this with @cppforlife and we arrived at what I thought was a cool compromise:
> 
> Given that the only error we'd be "throwing" is that the type is not convertible, what if we introduced an interface that specifically indicates the type is not convertible?
> 
> Something like:
> 
> ```go
> type UnconvertableStarlarkValue interface {
> 	ConversionHint() string
> }
> ```
> 
> So, instead of muddying up `StarlarkValueToGoValueConversion` with an `error` (and all the error checks), a type would instead implement the `UnconvertableStarlarkValue` interface. `StarlarkValue.asInterface()` detects this interface and returns an `error` with part of the error message coming from `ConversionHint()`.
> 
> Personally, I'm digging how this makes the code more explicit about what's going on: that there are some types that are just not convertible... and that _those_ types ought to give contextualized hints/help in _how_ they might be converted.
> 
> Thoughts?
> 
> ...
> 
> A few more comments in-line, below.
> 
> Note: a few thing I've marked as "_(optional)_". I'm sharing these as observations; I'm gonna leave it up to you whether you agree and want to include that change. _It's okay if you don't._

Thanks for the suggestions. I think it would make sense to have a separate changeset to cover this optional refactoring, as it's something invisible for consumers.

pkg/yamltemplate/filetests/ytt-library/url-value.yml
hanlins: Small correction here, in `test6: #@ testURL2.user.string()`, we actually should just keep it `test6: #@ testURL2.user` as `testURL2` doesn't contain user information, so its `user` member will be a `None` object.

pivotaljohn (approved):
This has been a journey in introducing a new kind of design element to `ytt`: a domain-specific Starlark type. You've been a terrific leading partner in this journey, @hanlins. Thank you for your efforts and commitment to getting us to this point. 🙏🏼

Issue #344, opened 2021-04-07
Report Schema Error when Data Values Overlay Fails


cari-lynn:
## Why / Context
Previously, there was some work done to report a schema error for when an overlay fails to overlay a data value onto a schema document, but the work was moved to this issue. Because of this, we have some tests that do not pass and are being skipped. Currently, instead of a schema error, we see an overlay error message.

Narrative:
**As a** configuration author
**I want** see a helpful error message when my data values file does not conform to my schema
**so that** I can easily find where the offending items are in my document and fix it

---

## 🟢  Report a Schema Error when Overlaying fails
**Given** I have a schema 
**And** a data value that is **not** in my schema
**When** I run `ytt` that includes that schema and data value
**Then** I see a error that my data value did not conform to my schema

```yaml
#@schema/match data_values=True
---
db_conn:
  port: 0
```

```yaml
#@ load("@ytt:overlay", "overlay")
#@data/values
---
db_conn:
  #@overlay/match by=overlay.subset({"key_doesnt_exist": True})
  new_key: i should not be here
```

```yaml
#! template.yml
#@ load("@ytt:data", "data")
---
db_conn: #@ data.values.db_conn
```

Then the result of `ytt -f schema.yml -f dv.yml -f template.yml --enable-experiment-schema` is:
```bash
data_values.yml:4 | new_key: i should not be here
                  |
                  | UNEXPECTED KEY - the key of this item was not found in the schema's corresponding map:
                  |      found: new_key
                  |   expected: (a key defined in map) (by schema.yml:3)
                  |   (hint: declare data values in schema and override them in a data values document)```
```

pivotaljohn:
With the rise of the "Plain YAML as Data Values" (#381) feature comes an approach to providing Data Values that is designed to _not_ result in overlay errors.

When _that_ method becomes the _primary_ way to supply data values, this work becomes far less valuable.

For that reason, we've decided to _at least_ defer this work until later: moved to #323 for this reason.

pivotaljohn:
We've recently clarified further the criteria for each pipeline in our backlog.
The probability that this bug will become "won't fix" has increased to the point it shouldn't be considered "carvel accepted" any longer. 

For these reasons, moving to "New Issues".

PR #374, opened 2021-05-03
Annotate node in schema to allow any type

gcheadle-vmware:


pkg/cmd/template/schema_test.go
```diff
@@ -646,6 +646,198 @@ overriden:
...
+  a: 1
+`
+
+		filesToProcess := files.NewSortedFiles([]*files.File{
+			files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(schemaYAML))),
+			files.MustNewFileFromSource(files.NewBytesSource("template.yml", []byte(templateYAML))),
+		})
+
+		assertSucceeds(t, filesToProcess, expected, opts)
+	})
```
cari-lynn: I believe that this case where data values aren't given is the same as this test: `TestSchema_Provides_default_values`. So I think we can remove this test.

pkg/cmd/template/schema_test.go
```diff
@@ -646,6 +646,198 @@ overriden:
...
+             |
+             | INVALID SCHEMA ANNOTATION - expected 'schema/type' annotation value in keyword argument 'any' to be a boolean, but was '1'
+`
+		filesToProcess := files.NewSortedFiles([]*files.File{
+			files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(schemaYAML))),
+		})
+
+		assertFails(t, filesToProcess, expectedErr, opts)
+	})
+}
```
cari-lynn: This tests may fit under `TestSchema_When_invalid_reports_error` since that's a kind of catch-all for schema invalid errors.

pkg/cmd/template/schema_test.go
```diff
@@ -646,6 +646,198 @@ overriden:
...
+`
+
+		filesToProcess := files.NewSortedFiles([]*files.File{
+			files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(schemaYAML))),
+			files.MustNewFileFromSource(files.NewBytesSource("dataValues.yml", []byte(dataValuesYAML))),
+			files.MustNewFileFromSource(files.NewBytesSource("template.yml", []byte(templateYAML))),
+		})
+
+		assertSucceeds(t, filesToProcess, expected, opts)
+	})
```
cari-lynn: I think narrowing the scope of this test to only test map items, and creating a separate test to test array items would fit with the pattern introduced with `TestSchema_Allows_null_values_via_nullable_annotation`. It would also make this test simpler and more focused which may make the tests easier to modify and add to in the future.

pkg/cmd/template/schema_test.go
```diff
@@ -646,6 +646,198 @@ overriden:
...
+bar:
+#@schema/type any=True
+- 0
+#@schema/type any=True
+baz:
+  a: 1
+`
+		dataValuesYAML := `#@data/values
+---
+#@overlay/replace
```
cari-lynn, line 827: I think I'm missing something here, I don't understand why the `#@overlay/replace` is necessary. Shouldn't this test work without it?

pkg/schema/schema.go
```diff
@@ -96,6 +97,13 @@ func NewMapItemType(item *yamlmeta.MapItem) (*MapItemType, error) {
 			"null value is not allowed in schema (no type can be inferred from it)",
 			"to default to null, specify a value of the desired type and annotate with @schema/nullable")
 	}
+	anyType, err := processTypeAnnotation(templateAnnotations)
+	if err != nil {
+		return nil, NewInvalidSchemaTypeAnnotationError(item, err.Error())
+	} else if anyType {
+		valueType = AnyType{Position: item.Position}
```
cari-lynn: The `else` can be removed here since the previous if returns and we're guaranteed to be in the 'else' case. (The same goes for the code in `NewArrayItemType` line 148).

pkg/schema/schema.go
```diff
@@ -167,6 +181,30 @@ func newCollectionItemValueType(collectionItemValue interface{}, position *filep
 	return nil, fmt.Errorf("Collection item type did not match any known types")
 }
 
+func processTypeAnnotation(templateAnnotations template.NodeAnnotations) (bool, error) {
+	if ann, anyType := templateAnnotations[AnnotationSchemaType]; anyType {
+		kwargName, err := core.NewStarlarkValue(ann.Kwargs[0][0]).AsString()
```
cari-lynn: We should do some validation that `ann.Kwargs[0][0]` exists. If someone accidentally types `#@schema/type any` instead of `#@schema/type any=True` this panics.

pkg/schema/schema.go
```diff
@@ -167,6 +181,30 @@ func newCollectionItemValueType(collectionItemValue interface{}, position *filep
...
+		if err != nil {
+			return false, err
+		}
+		if kwargName != "any" {
```
cari-lynn: "any" could be a constant.

pkg/schema/schema.go
```diff
@@ -167,6 +181,30 @@ func newCollectionItemValueType(collectionItemValue interface{}, position *filep
...
+			return false, fmt.Errorf("unknown '%v' annotation keyword argument '%v'. Supported kwargs are 'any'", AnnotationSchemaType, kwargName)
+		}
+		kwargBool, err := core.NewStarlarkValue(ann.Kwargs[0][1]).AsBool()
+		if kwargBool == false && err != nil {
+			kwargBool := core.NewStarlarkValue(ann.Kwargs[0][1]).AsGoValue()
```
cari-lynn: As a stylistic thing I would probably assign this to a new variable rather than reassign it to `kwargBool ` since they are different types and `kwargBool` is used later on as well.

pkg/schema/schema.go
```diff
@@ -167,6 +181,30 @@ func newCollectionItemValueType(collectionItemValue interface{}, position *filep
 	return nil, fmt.Errorf("Collection item type did not match any known types")
 }
 
+func processTypeAnnotation(templateAnnotations template.NodeAnnotations) (bool, error) {
```
cari-lynn: Since this function returns a bool, maybe rename it to indicate that. `hasAnyTypeAnnotation()`?

pkg/schema/schema.go
pivotaljohn: +1.

While we're talkin' names, here. At this point, I think we're certain of the _role_ of the second kwarg: it's whether `any` type is allowed. Surfacing the role into the name (a form of "revealing intention") can make the code more readable. 👍🏼

pkg/schema/schema.go
```diff
@@ -129,11 +137,17 @@ func NewArrayItemType(item *yamlmeta.ArrayItem) (*ArrayItemType, error) {
...
 	}
 
-	annotations := template.NewAnnotations(item)
+	templateAnnotations := template.NewAnnotations(item)
 
-	if _, found := annotations[AnnotationSchemaNullable]; found {
+	if _, found := templateAnnotations[AnnotationSchemaNullable]; found {
 		return nil, NewInvalidSchemaError(item, fmt.Sprintf("@%s is not supported on array items", AnnotationSchemaNullable), "")
 	}
+	anyType, err := processTypeAnnotation(templateAnnotations)
```
pivotaljohn: Can we consolidate the ways we check for the "various" (😁  there's two right now) annotations? Not sure if it fits, but I have an impulse to use a `switch` here, for example.

pkg/cmd/template/schema_test.go
gcheadle-vmware: This relates to #344. The overlay/replace allows us to get past the overlaying of data values without erroring. Without the overlay/replace, we get overlay errors about expecting a string but found a map. In this test, the string is the schema value of a node that is annotated as anyType, thus replacing the string with any value should be allowed by the schema (which it is).

pkg/schema/schema.go
gcheadle-vmware: I found that that panic when providing `#@schema/type any` comes from `ann.Kwargs[0]` being out of range, thus I think it is sufficient enough to check the length of `ann.Kwargs`.

However, one thing that I discovered that I could not figure out how to avoid  were these scenerios: 
- `#@schema/type any=something` gives this unhelpful error:
```
- undefined: something
    schema.yml:3 | #@schema/type any=something
````
however, we get an actual schema error message is `something` is in quotes.
- `#@schema/type any True` gives this unhelpful error:
```
- got identifier, want ','
    schema.yml:3 | #@schema/type any True
```
which seem to error during the EvalYaml call.
It doesn't feel like this issue is in scope due to the fact that I have seen this error before when using other annotations.

Comment (cari-lynn) [reply pkg/schema/schema.go]: Yeah I agree, that seems out of scope since those are similar errors to what we see in other areas of code now. 👍

File: pkg/schema/schema.go (hunk 10 above)
Comment (gcheadle-vmware) [reply pkg/schema/schema.go]: with `schema/nullable` and `schema/type`, is it possible to annotate a node with both annotations? If not, I agree that a switch statement would be the way to go.

File: pkg/cmd/template/schema_test.go (hunk 4 above)
Comment (gcheadle-vmware) [reply pkg/cmd/template/schema_test.go:827]: I was able to remove the annotation from the array item, and it did not error. So I'm going to look into this more to see if there is a way to make it possible for maps.

Comment (gcheadle-vmware) [reply pkg/cmd/template/schema_test.go:827]: The array item does not need the `#@overlay/replace` due to the schema's default data values. When we construct the default dvs for an array, we set the items to be empty: `typedNode.Items = []*yamlmeta.ArrayItem{}`. Since we can overlay onto `nil`, it overlays without the replace annotation. However, arrays are a special case in schemas since they do not have default values.

File: pkg/schema/schema.go (hunk 10 above)
Comment (gcheadle-vmware) [reply pkg/schema/schema.go]: is this a valid schema definition?
```
---
#@schema/type any=True
foo: null
```
just as this is a valid schama:
```
---
#@schema/nullable
foo: null
```

File: pkg/cmd/template/schema_test.go (hunk 4 above)
Comment (cari-lynn) [reply pkg/cmd/template/schema_test.go:827]: That makes sense why it doesn't work now. IIRC, this is the only place where we require an overlay annotation on the first data values document, so it's a divergent experience. Curious if you have thoughts on whether it would be possible to tackle here or if it makes sense to separate out?

File: pkg/schema/schema.go (hunk 10 above)
Comment (cari-lynn) [reply pkg/schema/schema.go]: This may have already been resolved offline, but my take here is yes this is valid

Comment (pivotaljohn) [reply pkg/schema/schema.go]: +1 @cari-lynn.

We _did_ talk about it offline. Meant to summarize here. Will do now.

That:
- "null" is the one value of the type "null" and so it is a legitimate value to have in a schema
- `type any=True` implies _all_ types, which includes "null" type

And:
- while `@schema/nullable` over a "null" is technically valid, it's also useless (a variable that can't vary?) but not seeing a need to attempt to prevent it, right now.
- one use-case of an author using `type any=True` is a required key whose contents are merely passed along: "null" seems like a reasonable default, here.

Did I miss anything @gcheadle-vmware ?

File: pkg/schema/schema.go [hunk 11]
```diff
@@ -88,14 +90,22 @@ func NewMapItemType(item *yamlmeta.MapItem) (*MapItemType, error) {
 		defaultValue = &yamlmeta.Array{}
 	}
 
-	templateAnnotations := template.NewAnnotations(item)
-	if _, nullable := templateAnnotations[AnnotationSchemaNullable]; nullable {
-		defaultValue = nil
-	} else if valueType == nil {
+	templateAnnotations, ann, err := processSchemaNodeAnnotations(item)
```
pivotaljohn: I'm finding it challenging to extract the general algorithm to deduce the type of an item.
All the ingredients are here... but they are squeezed together... this is making it hard for me to tease apart the various responsibilities, which is how I try to make sense of a chunk of code.
I think we should attempt to create some space between the various responsibilities.

### Taking Inventory of Responsibilities
What are these "responsibilities"?

Starting at `NewMapItemType()`, I see these responsibilities:
- extracting values from the kwargs of annotations (including converting from Starlark to Go)
- validating those values
- determining whether two types can co-exist (said another way, whether they can combine into a single type)
- "calculating" the default value of the type

How might we figure out how to "better" organize these?

### Sketching

I did two things:
1. wrote pseudo code of an "algorithm" — the minimum set of steps, involving the smallest number of things.
2. drew a picture of what could be constituent parts — what are all the things? what categories do they fall into? what are the responsibilities of each category?

#### Pseudo Code
Here's that pseudo code:

```
determine the type:
  get the inferred type from the item's value. 
  get the explicit type from the item's annotations.
  if neither yielded a valid type  (e.g. _only_ "null" type)
    err
  if the explicit type can combine the inferred type,
    type = (explicit or inferred)
  else
    type = (explicit)

determine the default value:
  get the inferred default value from the item's value
  get the explicit default value from the item's annotations
  if explicit value exists
    default = explicit
  else
    default = implicit

return &item{valueType: type, defaultValue: default}
```

#### A "Class" Diagram

And I used Miro for that picture: [(Miro Board) "ytt boxes and lines : Elements of Schema"](https://miro.com/app/board/o9J_kopV1_U=/?moveToWidget=3074457358481979842&cot=14)
_(It's way more than the scope of what we're looking at, but I was trying to get perspective, so I zoomed out a bit — thinking about the bigger picture.)_

FWIW, it felt like the pseudo code came out of "top-down" thinking and the diagram came out of "bottom-up" thinking.

### Reimagining `NewMapItemType()`

With these in hand, what if we reimagined `NewMapItemType()` so that it:
- delegates:
  - getting the `Type` from any annotations on the `MapItem` (and all that's involved)
  - inferring the `Type` from the value of the `MapItem`
  - getting the default value from the `MapItem`
- takes responsibility for:
  - given the inferred and explicit `Type`s determine the final `Type`
  - given final Type, default Value, etc... build the `MapItemType`

Thoughts?

pkg/schema/type.go
```diff
@@ -338,6 +360,39 @@ func (m *ScalarType) AssignTypeTo(typeable yamlmeta.Typeable) (chk yamlmeta.Type
...
+		typeable.SetType(&MapItemType{Key: typedItem.Key, ValueType: a, Position: a.Position})
+		typeableValue, ok := typedItem.Value.(yamlmeta.Typeable)
+		if ok {
+			a.AssignTypeTo(typeableValue)
+		}
+	}
+
+	return
+}
+
```
pivotaljohn: Random thought: what if the "zero value" of `Type` _is_ `AnyType`? That's another way of satisfying the "all Nodes have a Type" invariant...

Could be:
- all nodes come out of a constructor function
- that all nodes start life _with_ a valid for `Type`
- and that value is `AnyType`

...or...
- move all access to `Type` an accessor `GetType()`
- define that accessor to return `AnyType` when the underlying field is `nil`

Seems like it would make it easier to retain the invariant _and_ would obviate this whole function.

pkg/schema/type.go
```diff
@@ -117,6 +129,9 @@ func (m ScalarType) String() string {
 		return fmt.Sprintf("%T", m.Value)
 	}
 }
+func (a AnyType) String() string {
+	return "any type"
```
pivotaljohn:
```suggestion
	return "any"
```

Based on, [Schemas — Part 3: Supported Types](https://hackmd.io/pODV3wzbT56MbQTxbQOOKQ?view#Part-3-Supported-Types) there exists _some_ semblance of a "convention" for this type namespace.

Until we GA the feature, this namespace (and frankly all others) are open to debate ... and then it becomes part of the published "API".

pkg/schema/type.go
```diff
@@ -13,13 +13,16 @@ import (
 
 const (
 	AnnotationSchemaNullable structmeta.AnnotationName = "schema/nullable"
+	AnnotationSchemaType     structmeta.AnnotationName = "schema/type"
+	SchemaTypeAny            string                    = "any"
```
pivotaljohn:
```suggestion
	TypeAnnotationKwargAny   string                    = "any"
```

Prior art:
https://github.com/vmware-tanzu/carvel-ytt/blob/b4a5e57e0b7b6d2d27edd857f8c8d3907172133b/pkg/yttlibrary/overlay/match_annotation_expects_kwarg.go#L16-L21

pkg/schema/type.go
```diff
@@ -13,13 +13,16 @@ import (
 
 const (
 	AnnotationSchemaNullable structmeta.AnnotationName = "schema/nullable"
+	AnnotationSchemaType     structmeta.AnnotationName = "schema/type"
```
pivotaljohn:
```suggestion
	AnnotationNullable structmeta.AnnotationName = "schema/nullable"
	AnnotationType     structmeta.AnnotationName = "schema/type"
```

Prior art:
https://github.com/vmware-tanzu/carvel-ytt/blob/b4a5e57e0b7b6d2d27edd857f8c8d3907172133b/pkg/yttlibrary/overlay/annotations.go#L14-L26

Also, subtle: when thinking about `type`s and `const`s, (especially exported ones), do include the package name in your consideration. Here, this means there's no need to (again) qualify the constant name. 👍🏼

pkg/schema/error.go
```diff
@@ -137,6 +144,23 @@ func (t unexpectedKeyError) Error() string {
 	return msg
 }
 
+type invalidSchemaTypeAnnotationError struct {
```
pivotaljohn:
```suggestion
type invalidTypeAnnotationError struct {
```

For brevity, can be shortened given that we're in the `schema` package.

pivotaljohn: More fundamental question: whether we need a new error type here.

Is there a significant difference between an `invalidSchemaError` and a `invalidTypeAnnotationError`?  I personally haven't thought deeply about the `schema` package's errors.

I _think_:
- keep the number of `error` types to as few as possible (managing complexity)
- without losing meaningful distinctions (e.g. different data captured in each situation; different error message format desired; potentially different handling by callers...)

pkg/schema/type.go
```diff
@@ -145,6 +160,9 @@ func (t *MapItemType) CheckType(node yamlmeta.TypeWithValues) (chk yamlmeta.Type
 		// A Map must've yielded a non-MapItem which is not valid YAML
 		panic(fmt.Sprintf("MapItem type check was called on a non-MapItem: %#v", node))
 	}
+	if _, anyType := t.ValueType.(AnyType); anyType {
+		return
+	}
```
pivotaljohn: Before the `AnyType` came along, we made "null" a special case (i.e. wrote explicit checks for null like the `if` just below here.

One consequence of that design choice is that it forced checking the _Value_ of a MapItem to be split:
- _sometimes_ occur in `MapItem.Check()` and,
-  _other times_ occur in the `{MapType|ArrayType|ScalarType}.CheckType()`

If we implement `@schema/nullable` as a `Type` then all the checks can occur in that second spot.
This leave the `MapItem.Check()` can be a simple structure check like it is in `ArrayItemType.CheckType()`.

So, while this wasn't a design choice made within _this_ story, we're forced to make matters worse... and this starts a downward spiral unless we jump off, right here.

Nothing is sacred and we're _almost always_ better informed today than we were last go around: let's make the `NullType` so that there's no need to do _any_ explicit check here.

pivotaljohn (changes requested):
Apologies for the delay in this review, @gcheadle-vmware.

Taking a full step back, `@schema/type any=True` is a profound add. It isn't just that a new annotation got added (and so, now there are two), or that it's the first schema annotation with kwargs...
... it critically highlights where certain parts of our design are wanting for a little more future proofing.

When navigating, sometimes we discover that what we _genuinely_ thought was a good course turns out to include trade-offs we — it turns out — don't want to make.

This is true for this story.

Here's the summary:

1. we didn't realize `@schema/nullable` as a `NullType`. That move resulted in unavoidable accidental complexity for this story. You stemmed the bleeding, but this is only going to get worse if we leave it, as is. (details in `schema/type.go`)
2. you clearly made a yeoman's effort in wrangling all the parts that reach from a given `Type` down through to the `@schema/type` annotation. Given that we're going to only expand our footprint, in this part of the code, we'd have more approachable code, but more extensible if we teased apart all the things into layers. I seeded that thinking with some pseudo code and a diagram. (see `schema/schema.go` for more).
3. the rest are some suggestions for nips and tucks, here and there.

pkg/cmd/template/schema_test.go
gcheadle-vmware: Since #344 has been unprioritized, John and I discussed where this will be addressed in our future work. The conclusion that we came to was that we will not need the overlay annotations once the data values are passed using the `--data-values-file` flag. When the `--data-values-file` feature is implemented, this test, along with ['when map item's key is not among those declared in schema'](https://github.com/vmware-tanzu/carvel-ytt/blob/91cd2f10285ee2a456005bd5a3d9fad3cdf14baa/pkg/cmd/template/schema_test.go#L183), and ['checks after every data values document is processed and stops if there was a violation'](https://github.com/vmware-tanzu/carvel-ytt/blob/91cd2f10285ee2a456005bd5a3d9fad3cdf14baa/pkg/cmd/template/schema_test.go#L380) should pass without any overlay annotations.

pkg/schema/type.go
gcheadle-vmware: `if node.Type == nil {return AnyType{}}`

pivotaljohn (approved):
Yeomen's work, folks! Not only do we have support for the "any" type, but extended the type system to include a new "breed" — meta types that combine well with the existing primitive types.

🙏🏼

PR #385, opened 2021-05-10
Report attempts to encode URLValue as errors

pivotaljohn:
... to reserve panics for unexpected/exceptional conditions (of which
    this is not).

Also:
- give custom types module-namespaced name
- evolving conversion hints for URLValue and URLUser (staring at it
    longer and think these are more readable).


pkg/template/core/starlark_struct.go
```diff
@@ -49,7 +49,11 @@ func (s *StarlarkStruct) AttrNames() []string {
 }
 
 func (s *StarlarkStruct) Get(key starlark.Value) (val starlark.Value, found bool, err error) {
-	if attr, ok := NewStarlarkValue(key).AsGoValue().(string); ok {
+	goValue, err := NewStarlarkValue(key).AsGoValue()
```
cppforlife: i would name this attrVal or keyVal

cppforlife: we got AsStringValue which seems appropriate to use here. that should eliminate custom error on the next line.

pkg/template/core/starlark_value.go
```diff
@@ -4,16 +4,19 @@
...
 	"github.com/k14s/starlark-go/starlark"
 	"github.com/k14s/ytt/pkg/orderedmap"
 )
 
 type StarlarkValueToGoValueConversion interface {
-	AsGoValue() interface{}
+	AsGoValue() (interface{}, error)
 }
 
+var _ = []StarlarkValueToGoValueConversion{&StarlarkValue{}}
```
cppforlife: it's better written as

var _ StarlarkValueToGoValueConversion = &StarlarkValue{}

pkg/template/core/starlark_value.go
```diff
@@ -55,37 +58,37 @@ func (e StarlarkValue) AsInt64() (int64, error) {
 	return 0, fmt.Errorf("expected starlark.Int")
 }
 
-func (e StarlarkValue) asInterface(val starlark.Value) interface{} {
+func (e StarlarkValue) asInterface(val starlark.Value) (interface{}, error) {
 	if obj, ok := val.(UnconvertableStarlarkValue); ok {
-		panic(obj.ConversionHint())
+		return nil, errors.New(obj.ConversionHint())
```
cppforlife: i think we should have something like `fmt.Errorf("xxx: %s")` where `xxx` is some generic text that indicates that value is unconvertable.

pkg/template/core/starlark_value.go
```diff
@@ -107,34 +110,54 @@ func (e StarlarkValue) asInterface(val starlark.Value) interface{} {
...
 }
 
-func (e StarlarkValue) structAsInterface(val *StarlarkStruct) interface{} {
+func (e StarlarkValue) structAsInterface(val *StarlarkStruct) (interface{}, error) {
 	// TODO accessing privates
 	result := orderedmap.NewMap()
+	var err error
 	val.data.Iterate(func(k, v interface{}) {
-		result.Set(k, e.asInterface(v.(starlark.Value)))
+		if err == nil {
```
cppforlife, line 136: should we add IterateWithErr?

pkg/template/evaluation_ctx.go
```diff
@@ -65,7 +68,11 @@ func (e *EvaluationCtx) TplReplace(thread *starlark.Thread, _ *starlark.Builtin,
 func (e *EvaluationCtx) TplSetNode(thread *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
 	if args.Len() > 1 {
 		if _, noop := args.Index(1).(*core.StarlarkNoop); !noop {
-			err := e.parentNodes[len(e.parentNodes)-1].SetValue(core.NewStarlarkValue(args.Index(1)).AsGoValue())
+			goValue, err := core.NewStarlarkValue(args.Index(1)).AsGoValue()
```
cppforlife: should we just name it `val`?

pkg/yttlibrary/url.go
```diff
@@ -221,7 +226,7 @@ func (uv *URLValue) AsStarlarkValue() starlark.Value {
 }
 
 func (uv *URLValue) ConversionHint() string {
-	return "URLValue: cannot coerce to a string; use .string()"
+	return "URLValue: does not automatically encode; use .string()"
```
cppforlife: instead of using semi-colon, can we use either `--` or, similar to some other messages `(hint: use .string())`

pkg/yttlibrary/url.go
```diff
@@ -245,8 +250,10 @@ func (uv *URLValue) User() starlark.Value {
 	return uu
 }
 
+func (uu *URLUser) Type() string { return "@ytt:url.user" }
```
cppforlife, line 253: since we are getting into "first classing" type, should we also be adding Len(), Truthy()?

pivotaljohn: Yes. There's another flight of changes I want to make in this area (the nub of which is the `TODO` at the embedding of `StarlarkStruct` into these types).

_This PR_ is focused on getting the error messages to be more user-friendly.

Let me attempt to address the other feedback, here, and see if I can't get to this too...

pkg/template/core/starlark_struct.go
pivotaljohn: 👍🏼  ...also pulled type information from the `starlark.Value`.

pkg/template/core/starlark_value.go
pivotaljohn: fixed.

pkg/template/core/starlark_value.go
pivotaljohn: yup, makes sense.

fixed.

pkg/yttlibrary/url.go
pivotaljohn: Yeah, elsewhere we're using the `(hint: %s)` format... in an effort to be consistent (and I find the potential corrective action easier to find that way).

Fixed.

pkg/template/core/starlark_value.go
pivotaljohn: No. That would just make too much sense.

Fixed.

pkg/template/evaluation_ctx.go
pivotaljohn: Yeah, that's perfect.

Fixed.

pkg/orderedmap/map.go
```diff
@@ -72,6 +72,16 @@ func (m *Map) Iterate(iterFunc func(k, v interface{})) {
 	}
 }
 
+func (m *Map) IterateWithErr(iterFunc func(k, v interface{}) error) (err error) {
```
cppforlife: looks like below IterateErr does the right thing already.

pkg/template/core/starlark_struct.go
```diff
@@ -49,14 +49,15 @@ func (s *StarlarkStruct) AttrNames() []string {
...
 func (s *StarlarkStruct) Get(key starlark.Value) (val starlark.Value, found bool, err error) {
-	if attr, ok := NewStarlarkValue(key).AsGoValue().(string); ok {
-		val, found := s.data.Get(attr)
-		if found {
-			return val.(starlark.Value), true, nil
-		}
-		return starlark.None, false, nil
+	attrName, err := NewStarlarkValue(key).AsString()
+	if err != nil {
+		return starlark.None, false, fmt.Errorf("expected key `%s` to be a string but is a %s", key, key.Type())
```
cppforlife: do you think we still need to have this error message here? (assumption here is that https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/template/core/starlark_value.go#L37 should be good enough, if not can we improve it there?) if we do want to keep this message here, then we should start it with capital letter (to be consistent with other error messages) and use single quotes instead of ticks (ticks arent used in any carvel tools).

pkg/template/core/starlark_value.go
```diff
@@ -55,37 +57,37 @@ func (e StarlarkValue) AsInt64() (int64, error) {
 	return 0, fmt.Errorf("expected starlark.Int")
 }
 
-func (e StarlarkValue) asInterface(val starlark.Value) interface{} {
+func (e StarlarkValue) asInterface(val starlark.Value) (interface{}, error) {
 	if obj, ok := val.(UnconvertableStarlarkValue); ok {
-		panic(obj.ConversionHint())
+		return nil, fmt.Errorf("unable to convert value: %s", obj.ConversionHint())
```
cppforlife: nit: capitalize first letter in error message

pkg/orderedmap/map.go
pivotaljohn: _literally_ right below.

Fixed.

pkg/template/core/starlark_struct.go
pivotaljohn: Ah, yeah, I see how I inadvertently "shadowed" the underlying error message.

Also, I see how this is a separate case from the other type errors we're handling here. In _this_ case, the value _must_ be a `string`, not just coerce to one. 👍🏼 

Fixed.

pkg/template/core/starlark_value.go
pivotaljohn: fixed.

hanlins:
Closes https://github.com/vmware-tanzu/carvel-ytt/issues/366.

Issue #381, opened 2021-05-06
Accept Data Values as a plain YAML file


pivotaljohn:
Per [ytt-002 Data Values as Plain YAML](https://github.com/vmware-tanzu/carvel-community/blob/f5213952f6c0b1e36e321236a74bccc1e9fc088f/proposals/ytt/002-raw-data-values/README.md) proposal.

This feature serves two primary classes of users:

**As an** Integrator
**I want** to be able to present _my_ users with a simple YAML-based configuration interface, devoid of any `ytt` adornments (e.g. annotations, Starlark code, etC)
**So that** my users need not know I've integrated with `ytt` nor learn any of its syntax
**And** I don't have to do anything special to make that integration "just work" (e.g. prepending a file with `@data/values` or scan/filter for code when I don't want to provide the option for it, etc)

**As a** Configuration Consumer
**I want** to pick up and use someone's `ytt` library, including providing my own configuration values (apparently called "Data Values") in a simple YAML format
**So that** I don't have to learn `ytt` syntax or learn advanced `ytt` features

In short:

Add a flag to `ytt` — `--data-values-file` that accepts _only_ plain YAML (i.e. with no `ytt` directives).

Stories:
- [x] #376 
- [x] #377
- [x] #378 
- [x] #379 
- [x] #380 
- [x] #387

pivotaljohn:
@cppforlife please review the connected stories for acceptance criteria around documentation.

james-pollard-leapyear:
Hey y'all,

I'm trying to track what came in with the latest release and I'm having trouble following how #398 implements this proposal, while there are still the 6 unclosed referenced in the description for this issue.

Some clarification would be greatly appreciated!

pivotaljohn:
The feature is implemented and included as a "pre-release" in v0.34.0. However, these stories have unmet acceptance criteria around updating documentation. Once that work is done, we can consider the feature released.

We didn't make that clear in the release notes. Sorry for the ambiguity.

The `ytt` fanatic will note that there's scope identified in the proposal required to push this more user-friendly approach to supplying Data Values to the fore-front. That scope is about properly adjusting documentation and supporting this new flag as the primary way to supply Data Values (e.g. Playground doesn't support setting flags).

In the meantime, this feature is, in fact, ready for use.

james-pollard-leapyear:
[...]

PR #398, opened 2021-05-24
Data values files

cppforlife:
still in draft.

implements this proposal: https://github.com/vmware-tanzu/carvel-community/tree/develop/proposals/ytt/002-raw-data-values for https://github.com/vmware-tanzu/carvel-ytt/issues/381

this PR also:
- adds `or_add=True` to `@overlay/replace` so that when map key does not match, replace would fallback to just adding map item
- relaxes data values given over cmd line (--data-values-* family) to not require existing values in the data values overlays

### DIFF
--- a/pkg/cmd/template/cmd_data_values_file_test.go
+++ b/pkg/cmd/template/cmd_data_values_file_test.go
@@ -1,0 +1,151 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package template_test
+
+import (
+	"fmt"
+	"testing"
+
+	cmdtpl "github.com/k14s/ytt/pkg/cmd/template"
+	"github.com/k14s/ytt/pkg/cmd/ui"
+	"github.com/k14s/ytt/pkg/files"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDataValuesWithDataValuesFileFlags(t *testing.T) {
+	yamlTplData := []byte(`
+#@ load("@ytt:data", "data")
+values: #@ data.values`)
+
+	builtinDVs := []byte(`
+#@data/values
+---
+predefined: true`)
+
+	// Ensure various non-annotation YAML comments
+	// are allowed and do not affect parsed content
+	dvs1 := []byte(`
+# top comment
+int: 123
+str: str
+boolean: false
+nested:
+  #comment without space
+  # comment with space
+  value: not-str
+  ### some other unknown comment
+  nested:
+    #! ytt comment1
+    #! ytt comment2
+    subnested: true
+another:
+  nested:
+    map: {"a": 123}
+array:
+- 123
+- str
+
+# bottom comment`)
+
+	dvs2 := []byte(`
+int: 123
+str: str
+boolean: true
+nested:
+  value: not-str
+  nested: true
+another:
+  nested:
+    map: {"a": 123}
+# Multiple documents in one file
+---
+array:
+- str`)
+
+	// Ensure file with only comments (no structures) is allowed
+	dvs3 := []byte(`# value: 1
+# value: 2`)
+
+	expectedYAMLTplData := `values:
+  predefined: true
+  int: 123
+  str: str
+  boolean: true
+  nested:
+    value: not-str
+    nested: true
+  another:
+    nested:
+      map:
+        a: 123
+  array:
+  - str
+`
+
+	filesToProcess := files.NewSortedFiles([]*files.File{
+		files.MustNewFileFromSource(files.NewBytesSource("tpl.yml", yamlTplData)),
+		files.MustNewFileFromSource(files.NewBytesSource("values.yml", builtinDVs)),
+	})
+
+	ui := ui.NewTTY(false)
+	opts := cmdtpl.NewOptions()
+
+	opts.DataValuesFlags = cmdtpl.DataValuesFlags{
+		FromFiles: []string{"dvs1.yml", "dvs2.yml", "dvs3.yml"},
+		ReadFileFunc: func(path string) ([]byte, error) {
+			switch path {
+			case "dvs1.yml":
+				return dvs1, nil
+			case "dvs2.yml":
+				return dvs2, nil
+			case "dvs3.yml":
+				return dvs3, nil
+			default:
+				return nil, fmt.Errorf("Unknown file '%s'", path)
+			}
+		},
+	}
+
+	out := opts.RunWithFiles(cmdtpl.Input{Files: filesToProcess}, ui)
+	require.NoError(t, out.Err)
+	require.Len(t, out.Files, 1, "unexpected number of output files")
+
+	file := out.Files[0]
+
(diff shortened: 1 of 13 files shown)


pkg/cmd/template/data_values_flags.go
```diff
@@ -242,23 +296,26 @@ func (s *DataValuesFlags) buildOverlay(keyPieces []string, value interface{}, de
...
 	for _, piece := range keyPieces {
-		newMap := &yamlmeta.Map{}
-		nodeAnns := template.NodeAnnotations{}
-
-		if strings.HasSuffix(piece, missingOkSuffix) {
+		// In previous versions of ytt '+' key suffix was meant to indicate
+		// "@overlay/match missing_ok=True". It's no longer needed as all keys
+		// are considered as such. Data values schemas should be enough to
+		// provide key checking/validations.
+		// (Stripped for backwards compatibility.)
```
pivotaljohn: This feature was not documented; few if any users knew about it.

What do you think of simply removing it, reducing code complexity?

cppforlife: i dont have a strong feeling. could go either way. any preference?

pivotaljohn: Personally, I'd delete it (and we'll note the breaking change) except: keep the line about "schemas should be enough to provide key checking/validations." as a kind of justification as to why these inputs are so permissive.

I don't see real value in keeping the code or reference to how things "were."

I _do_ see value in making explicit an assumption one part of the code holds about another's responsibilities to help tie together how we expect the overall experience to unfold.

pkg/cmd/template/cmd_data_values_file_test.go
```diff
@@ -0,0 +1,150 @@
...
+	yamlTplData := []byte(`
+#@ load("@ytt:data", "data")
+values: #@ data.values`)
+
+	builtinDVs := []byte(`
+#@data/values
+---
+predefined: true`)
+
+	// Comments are added on purpose
```
pivotaljohn: can we succinctly state that purpose?

pkg/cmd/template/cmd_data_values_file_test.go
```diff
@@ -0,0 +1,150 @@
...
+	"testing"
+
+	cmdtpl "github.com/k14s/ytt/pkg/cmd/template"
+	"github.com/k14s/ytt/pkg/cmd/ui"
+	"github.com/k14s/ytt/pkg/files"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDataValuesWithDataValuesFileFlags(t *testing.T) {
```
pivotaljohn, line 17: I think I see three distinct test cases in one, here.

Any reason not to split this out?

pkg/cmd/template/cmd_data_values_file_test.go
```diff
@@ -0,0 +1,150 @@
...
+	opts.DataValuesFlags = cmdtpl.DataValuesFlags{
+		FromFiles: []string{"dvs1.yml"},
+		ReadFileFunc: func(path string) ([]byte, error) {
+			switch path {
+			case "dvs1.yml":
+				return dvs1, nil
+			default:
+				return nil, fmt.Errorf("Unknown file '%s'", path)
+			}
+		},
```
pivotaljohn, line 146: That the contents of the files are read in through a function like this seems a bit noisy to me.

What matters at _this_ interface is that a file named X will yield content Y.

What if we wrote a little helper to produce the function, here?

It would make the test that much more readable because the reader wouldn't read over those (irrelevant/test harness implementation) details.

pkg/cmd/template/data_values_file.go
```diff
@@ -0,0 +1,93 @@
...
+		if valNode, ok := val.(yamlmeta.Node); ok {
+			err := f.disallowAnyAnnotations(valNode)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
```
pivotaljohn: Any reason not to find a way to export 
https://github.com/vmware-tanzu/carvel-ytt/blob/553ef493d768bcd419c60776f3622c88c0b180bb/pkg/yamltemplate/template.go#L41-L66

from `yamltemplate` and use that here?

I reads to me like this is not just inadvertently the same code, but almost literally the same for the same purpose.

pkg/cmd/template/data_values_flags.go
```diff
@@ -67,6 +72,17 @@ func (s *DataValuesFlags) AsOverlays(strict bool) ([]*workspace.DataValues, []*w
 
 	var result []*workspace.DataValues
 
+	// Files go first
```
pivotaljohn, line 75: Yeah, like what was said in the group session, these "headers" really work for me, too. 👍🏻

pkg/cmd/template/cmd_data_values_file_test.go
cppforlife: my thinking is that it's a lot less noise to deal with. imho this test even though it covers several cases is still pretty lightweight and easy enough to understand.

pkg/cmd/template/cmd_data_values_file_test.go
cppforlife: got suggestion how syntactically represent that. all readable variants take multiple lines and all of them would have to return a function from a function. do throw in an example of what you are thinking here.

pkg/cmd/template/data_values_file.go
cppforlife: havent seen that function. sounds good.

pkg/cmd/template/data_values_flags.go
cppforlife: ok.

pkg/cmd/template/cmd_data_values_file_test.go
pivotaljohn: Checking your perspective from another angle: had _you_ come across this as a small set of tests instead of one medium test, would you have "encouraged" them to be combined?

pkg/cmd/template/data_values_file.go
cppforlife: done.

pkg/cmd/template/cmd_data_values_file_test.go
cppforlife: done.

pkg/cmd/template/data_values_flags.go
cppforlife: done.

pkg/cmd/template/cmd_data_values_file_test.go
cppforlife: i havent closely looked at the code base recently, so nothing comes to mind.

cppforlife:
updated HasTemplating() to accept yamlmeta.Node. looks better!

pivotaljohn:
This looks great to me.

The conversation about organizing tests really ought to take into account _all_ `--data-value...` flag functionality; we'll peel that off for another time.

Thank you for your contribution, Mr. Kalinin.

Issue #390, opened 2021-05-17
Report clearly which kind of document(s) were unused during an evaluation


cari-lynn:
_(significantly edited from original as the code evolved since the issue was first created. However, the overall scope has not changed, just the details.)_

### 🟡  Scenario: Library Reference on Schema does not match any libraries

**given** I have a Schema document targeted to a specific library
**and** no library exists by that name
**when** I run `ytt`
**then** I expect to see an error message that makes clear that there were _Schema_ documents that weren't used by any libraries
**instead of** an error message that says there are "... library data values documents..."

### 🟡  Scenario: Library Reference on Data Values does not match any libraries

**given** I have a Data Value document targeted to a specific library
**and** no library exists by that name
**and** there weren't any "unused" Schema documents
**when** I run `ytt`
**then** I expect to see an error message that makes clear that there were Data Value _documents_ that weren't used by any libraries

gcheadle-vmware:
Closing due to v0.36.0 release.

PR #399, opened 2021-05-25
Extract document schema

pivotaljohn:
- Introduces a refactor to hopefully communicate the difference between schema's used by the 'current' library and a schema intended for a 'child' library.

Or put another way:
Splitting the entity of a document schema, and the 'usage' of a document schema, in this case locating the correct library and marking it used or not by said child library.

- Slight improvement in the error message when a data value is not used. Changed "DataValue" to "Data Value"


pivotaljohn:
The metaphor is really working for me. 👍 Names are HARD!

Inspires me to want to make the mirror change to DataValues; but that is for another time. 😉 

We're going to completely address the "not used" error message in #390.

PR #401, opened 2021-05-26
Overlay data values int mismatch

### LINKED ISSUE #394: Integer Data Value from overlay'ed Starlark expression type checks successfully
## Acceptance Criteria

**Given** Schema declaring a Data Values as an `int`
**And** a Data Value overlay containing a Starlark integer expression

```console
.
├── _ytt_lib
│   └── libby
│       └── schema.yml
└── config.yml
```

`_ytt_lib/libby/schema.yml`
```yaml
---
foo: 4
```

`config.yml`
```yaml

#@ def values():
foo: #@ 4
#@ end

--- #@ template.replace(library.get("libby").with_data_values(values()).eval())
```

**When** evaluated

```console
```

Should succeed.

---
`ytt` built at `acb019d`  (circa v0.33.0 + `overlay-schemas` branch) would produce this output:

```console
ytt: Error:
- library.eval: Evaluating library 'libby': Overlaying data values (in following order: additional data values):
    in <toplevel>
      config.yml:8 | --- #@ template.replace(library.get("libby").with_data_values(values()).eval())

    reason:
     config.yml:5 | foo: #@ 4
                  |
                  |      found: int64
                  |   expected: integer (by _ytt_lib/libby/schema.yml:3)
```

## Context

During exploratory testing, found that this scenario is more likely than we first believed. In fact, it's a blocker for any project with the following setup:

When a root library sets data values on a child library (via `library.with_data_values()`;
and one or more of those data values' value is a Starlark expression;
then, the subsequent `eval()` fails (because of the gap described in this issue).

For example:

Using `ytt` built from `acb019d` (i.e. late-stage `overlay-schema`), when rendering config for `cf-for-k8s`, this error results:

```
ytt: Error:
- library.eval: Evaluating library 'capi-k8s-release': Overlaying data values (in following order: capi/_ytt_lib/capi-k8s-release/config/values/_defaults.yml, capi/_ytt_lib/capi-k8s-release/config/values/images.yml, additional data values):
    in <toplevel>
      capi/capi.yml:97 | --- #@ template.replace(capi.with_data_values(capi_values()).eval())

    reason:
     capi/capi.yml:46 |   port: #@ data.values.capi.database.port
                      |
                      |      found: int64
                      |   expected: integer (by values/capi_schema.yml:26)
```

DennisDenuto:
Fixes #394 

- Introduce fuzz testing to hopefully help drive out issues like #394 in the future.
- Also ensure that the error message returned during schema type checking are consistent between int, int64, and uint64. i.e. when a type mismatch occurs, refer to the above types as 'integer'.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/aws/aws-lambda-go v1.8.2
 	github.com/cppforlife/cobrautil v0.0.0-20200514214827-bb86e6965d72
 	github.com/cppforlife/go-cli-ui v0.0.0-20200505234325-512793797f05
+	github.com/google/gofuzz v1.2.0
 	github.com/hashicorp/go-version v1.2.0
 	github.com/k14s/difflib v0.0.0-20201117154628-0c031775bf57
 	github.com/k14s/starlark-go v0.0.0-20200720175618-3a5c849cc368
--- a/go.sum
+++ b/go.sum
@@ -54,6 +54,8 @@ github.com/google/go-cmp v0.2.0/go.mod h1:oXzfMopK8JAjlY9xF4vHSVASa0yLyX7SntLO5a
 github.com/google/go-cmp v0.3.0/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.3.1/go.mod h1:8QqcDgzrUqlUb/G2PQTWiueGozuR1884gddMywk6iLU=
 github.com/google/go-cmp v0.4.0/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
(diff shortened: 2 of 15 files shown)


DennisDenuto:
Seed number: 1622072430578435000

When YTT tries to renders float64: 0.00006598648894952906
It is output as scientific notation: 6.598648894952906e-05 

It hasn't lost any precision, and are equivalent. However it isn't *exactly* what the 'user' has input either....

`strconv.FormatFloat(float64(0.00006598648894952906), 'f', -1, 64)` is able to convert that float64 and return it as 
 the full decimal value correctly.

This feels like an edge case. I can't think of a practical use case and I think I will update the fuzz test to exclude float numbers below a certain threshold for now, documented around a comment as to why that limitation around floats exist.

TL;DR:

ytt uses scientific notation when there are 4 or more leading zeros. Fuzz test will be changed to not provide input values with 4 more of zeros.


EDIT:

Decided to go with a slightly different approach. Generate float64 numbers, with as many zeros as it generates with. Instead make the assertion flexible, to also allow the scientific notation as a valid output too.

pivotaljohn (approved):
LGTM!

Appreciate the extra credit work in introducing fuzz testing, here.

Issue #151, opened 2020-06-08
unknown file location for nodes added via template.replace


cppforlife:
repro: https://get-ytt.io/#gist:https://gist.github.com/cppforlife/d83f925e9f02b605205afd95afdae29d

pivotaljohn:
Q1: Is there any reason to not establish and maintain the invariant that the AST _always_ be a composite of `Node`?

Q2: what are the consequences of defining the "source line number" of such values to be the line where the `template.replace()` itself injects the new node(s)?

pivotaljohn:
A little backlog grooming: the "discussion" was had in comments on #443.

We'll consider this issue addressed, for now. There remains the question of how feasible it is to more precisely identify the source of a given value when it originates from within Starlark execution. If that need arises, a separate issue can be created to capture it.

PR #403, opened 2021-05-26
Mismatched Type Error: include "found" portion of error message when position is unknown

gcheadle-vmware:
"found" portion in error message is now included when position is unknown. 

In some investigation, I discovered that the position could also be `nil` if inserted using `template.replace()`. To prevent a nil dereference panic, if the inserted items are from starlark, then they are given an unknown position.


pivotaljohn:
This forced our hand a little to (partially) address #151.

We'll use _that_ issue to discuss the deeper issue of injecting non-AST into the tree.

PR #406, opened 2021-05-28
Migrate e2e from bash to golang

DennisDenuto:
As an ongoing effort to improve test output, convert the e2e tests (written in bash) to golang tests

- Will make reporting failures for units and e2e tests easier in the future
- Should make test failures consistent between units and e2e
- e2e now follows a hierarchical test layout, with the hope it is easier to keep inventory of what is being tested

Also...
Changed hack/test-all.sh to just run 'all the golang tests' (units and e2e). 
Removed `hack/test-unit.sh` `hack/test-e2e.sh`. Rationale being that e2e tests takes a second to run. So having a helper script to run a separate suite of tests didn't seem to make much sense.


pivotaljohn (approved):
This gives the flexibility to run test either through the shell script(s) _or_ through the test runner. 👍🏻  🙏🏻

PR #411, opened 2021-06-01
Polish schema errors

### LINKED ISSUE #396: Polish schema error messages
## Summary

The four (4) error messages around schema could benefit from:
- wording that is more accessible to a casual user
- formatting that separates and leads the eye to the more important information, first
so that they are easy to read.

## Details

Inspired by: 

During exploratory testing — with an eye towards [error message design work from the Rust project](https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html), many of the current schema error messages could be _significantly_ improved.

These include:
- more important information in a more prominent place
    1. what is the error? — the source line where the error likely occurred
    2. what's wrong? — the exact expectation/invariant that was broken
    3. where abouts? — roughly where about within the source file (i.e. line number)
    4. where _exactly_? — the name of the file
    5. how can I fix it? — hints of what to do next
- make clear what's literal source and what's commentary.

The format of these error messages need polish.

In summary, the changes are:
- keep source line dead center (where the eye typically goes first) with even more surrounding whitespace
- split filename and line number
  - reduce required eye movement to locate source line
  - makes output display more predictable (left-side is _always_ 8 characters wide)
- use `=` in the gutter to indicate line is commentary, not source
- bring description of actual and expected as just below the source line (where eye goes next)
- ALL CAPS reads like yelling; instead pull error message to the top-line (relatively low prominence) to get proper separation from other data.
- whitespace to separate "expectations" and "hints"
- allow for multiple hints (e.g. split on `\n`?) to be able to provide more than one piece of advice without it getting too complicated.
- use the verb "define" (instead of "declare") consistently throughout error messages.

## Scenarios

The following are the four (4) cases we currently report errors. Each

### Scenario 1: Invalid Schema

`null` provided as a value in Schema (which is both useless and likely a misguided attempt at making the default value `null` — when the author _should_ use `@schema/nullable`)
                       
**Before**
```
ytt: Error:
values/00-values.yml:4 | system_domain:
                       |
                       | INVALID SCHEMA - null value is not allowed in schema (no type can be inferred from it)
                       |   (hint: to default to null, specify a value of the desired type and annotate with @schema/nullable)
```

**After**

```
ytt: Error: Invalid schema — null value not allowed here

 values/00-values.yml:
       |
     4 | system_domain:
       |
       = found: null value
       = expected: non-null value
       =
       = hint: in YAML, omitting a value implies null.
       = hint: to set the default value to null, annotate with @schema/nullable.
       = hint: to allow any value, annotate with @schema/type any=True.
```

### Scenario 2: Invalid Array Definition

Author defined an array, but not the type-setting array item inside it.
Initially, this will be a very common message while authors are converting from Data Values to Schema.

**Before**
```
ytt: Error:
values/00-values.yml:7 | app_domains: []
                       |
                       | INVALID ARRAY DEFINITION IN SCHEMA - unable to determine the desired type
                       |      found: 0 array items
                       |   expected: exactly 1 array item, of the desired type
                       |   (hint: in a schema, the item of an array defines the type of its elements; its default value is an empty list)
```                    

**After**
```
ytt: Error: Invalid schema — wrong number of items in array definition

  values/00-values.yml:
       |
     7 | app_domains: []
       |
       = found: 0 array items
       = expected: exactly 1 array item, of the desired type
       =
       = hint: in schema, the one item of the array implies the type of its elements.
       = hint: in schema, the default value for an array is always an empty list.
       = hint: default values can be overridden via a data values overlay.
```

### Scenario 3: Mismatched Type

The value given in a Data Values does not match the type defined in the schema.

```
ytt: Error: Overlaying data values (in following order: values.yml):
values.yml:3 | foo: ""
             |
             |      found: string
             |   expected: integer (by schema.yml:3)
```

**After**
```
ytt: Error: Value is of wrong type

  values.yml:
       |
     3 | foo: ""
       |
       = found: string
       = expected: integer (by schema.yml:3)
```

### Scenario 4: Undefined Map Key

A map item in a data value has a key that does not match any of those in the map defined in the schema.

**Before**
```
ytt: Error: Overlaying data values (in following order: values/10-images.yml, values/20-secrets-config-values.yml, values/30-experimental-values.yml, cf-values.yml):
values/10-images.yml:4 | images:
                       |
                       | UNEXPECTED KEY - the key of this item was not found in the schema's corresponding map:
                       |      found: images
                       |   expected: (a key defined in map) (by values/00-values.yml:2)
                       |   (hint: declare data values in schema and override them in a data values document)
```

**After**
```
ytt: Error: Overlaying data values (in following order: values/10-images.yml, values/20-secrets-config-values.yml, values/30-experimental-values.yml, cf-values.yml): 
  Undefined key in map:

  values/10-images.yml:
       |
     4 | images:
       |
       = found: images
       = expected: (a key from the map defined at values/00-values.yml:2)
       =
       = hint: is the key spelled correctly?
       = hint: to extend existing schema, define new keys in a schema overlay.
```

DennisDenuto:
fixes #396 

Introduces a `SchemaError` that accepts errors with the goal of producing a consistent error message. Kinda like a report of all the errors it knows about.

`SchemaError` will always print a 'summary' when `Error` is called.

If `SchemaError` is given an error of type `schemaAssertionError` it will report those errors in a 'diff' format. with the line, file name, and expected / actual value formatted in a nice readable way.

For all other type of errors given to `SchemaError`, it will delegate to their `Error()`  and append those messages to the end of the `SchemaError` 'report'

### DIFF
--- a/pkg/schema/error.go
+++ b/pkg/schema/error.go
@@ -4,18 +4,47 @@
 package schema
 
 import (
+	"bytes"
 	"fmt"
+	"log"
 	"strings"
+	"text/template"
 
 	"github.com/k14s/ytt/pkg/filepos"
 	"github.com/k14s/ytt/pkg/yamlmeta"
 )
 
-func NewInvalidSchemaError(found yamlmeta.Node, message string, hints []string) error {
+const invalidSchemaTemplate = `
+{{.Title}}
+
+{{.FileName}}:
+{{pad "|" false}}
+{{pad "|" true}} {{.Diff}}
+{{pad "|" false}}
+
+{{with .Found}}{{pad "=" false}} found: {{.}}{{end}}
+{{with .Expected}}{{pad "=" false}} expected: {{.}}{{end}}
+{{range .Hints}}{{pad "=" false}} hint: {{.}}
+{{end}}`
+
+func NewInvalidSchemaError(err error, node yamlmeta.Node) error {
+	if schemaErrorInfo, ok := err.(schemaAssertionError); ok {
+		return &invalidSchemaError{
+			Title:    fmt.Sprintf("Invalid schema — %s", schemaErrorInfo.description),
+			FileName: node.GetPosition().AsCompactString(),
+			filePos:  node.GetPosition().AsIntString(),
+			Diff:     node.GetPosition().GetLine(),
+			Expected: schemaErrorInfo.expected,
+			Found:    schemaErrorInfo.found,
+			Hints:    schemaErrorInfo.hints,
+		}
+	}
+
 	return &invalidSchemaError{
-		Found:   found,
-		message: message,
-		hints:   hints,
+		Title:    "Invalid schema",
+		FileName: node.GetPosition().AsCompactString(),
+		filePos:  node.GetPosition().AsIntString(),
+		Diff:     node.GetPosition().GetLine(),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


DennisDenuto:
As per the instructions: Closing to re-sign the CLA and will re-open this PR

pkg/schema/error.go
```diff
@@ -4,25 +4,47 @@
...
+
+{{.FileName}}:
+{{pad "|" false}}
+{{pad "|" true}} {{.Diff}}
+{{pad "|" false}}
+
+{{with .Found}}{{pad "=" false}} found: {{.}}{{end}}
+{{with .Expected}}{{pad "=" false}} expected: {{.}}{{end}}
+{{range .Hints}}{{pad "=" false}} hint: {{.}}
+{{end}}`
```
pivotaljohn: I dig it:
- 👍🏻  delegate formatting to stdlib
- 👍🏻  declarative, concise, and easy to edit.

pkg/schema/error.go
```diff
@@ -40,47 +62,49 @@ func NewUnexpectedKeyError(found *yamlmeta.MapItem, definition *filepos.Position
...
+	hints       []string
+}
+
 type invalidSchemaError struct {
-	Message string
-	Found   yamlmeta.Node
-	Hint    string
+	Title    string
+	FileName string
+	Diff     string
```
pivotaljohn: If I'm reading the usage right, I think `Diff` is the original source.... 

```suggestion
	Source   string
```

?

DennisDenuto: Good callout. done!

pivotaljohn:
@DennisDenuto

Regarding the markdown style header line...
... it didn't land for me at first.
... but then you shared with me the intent of looking at this output as a kind of report. That really made a shift in my head.

I want to see this experiment get some air.

Thanks for your thoughtful approach to this story.

pivotaljohn (approved):

PR #414, opened 2021-06-02
Add examples for Using Schema

cari-lynn:
In order to provide examples to get users bootstrapped using ytt Schemas in an experimental fashion, this PR will provide two examples:

- A 'simple' example showing how to declare and use a schema file in a template, and how to override data values using command line flags.
- An array example based on simplified pipeline configuration from [cf-for-k8s](https://github.com/cloudfoundry/cf-for-k8s/blob/45aff2b9dbf87a609116c99055a5c19f358fad91/ci/templates/cf-for-k8s-contributions.yml)


examples/schema-arrays/config.yml
```diff
@@ -0,0 +1,3 @@
+#@ load("@ytt:data", "data")
+---
+jobs: #@ data.values.releases
```
pivotaljohn: nit: newline

examples/schema-arrays/values.yml
```diff
@@ -0,0 +1,10 @@
+#@data/values
+---
+releases:
+  - name: capi-k8s-release
+    github_release: false
+    vendir_github_release: false
+  - name: eirini-release
+    github_org: cloudfoundry-incubator
+    build_dir: "build/eirini"
+  - name: metric-proxy
```
pivotaljohn: nit: newline

test/e2e/e2e_test.go
```diff
@@ -258,6 +259,115 @@ float: 123.123
...
+			{"--enable-experiment-schema": ""},
+		}
+		actualOutput := runYtt(t, testInputFiles{"../../examples/schema-arrays/config.yml", "../../examples/schema-arrays/schema.yml"}, "", flags, nil)
+		expectedOutput := `jobs: []
+`
+
+		require.Equal(t, expectedOutput, actualOutput)
+	})
+
+	t.Run("--data-value-yaml flag", func(t *testing.T) {
```
pivotaljohn, line 298: I'm thinking... we start losing returns in testing that multiple types are checked. (or am I missing something?)

I thought I might see a test here that shows that through all the different ways that Data Values are set (i.e. the full complement of `--data-value...` flags) all override schema defaults.

**Suggestion:** combine this test and the following and include the other `--data-value...` flags for a single test that puts the whole set of mechanisms through their paces.

test/e2e/e2e_test.go
```diff
@@ -258,6 +259,115 @@ float: 123.123
...
+any:
+- 1
+- 2
+- 4
+`
+
+		require.Equal(t, expectedOutput, actualOutput)
+	})
+
+	t.Run("array with data values file", func(t *testing.T) {
```
pivotaljohn:
```suggestion
	t.Run("defaults missing map items", func(t *testing.T) {
```

**rationale:** what's being exercised here seems to me to be more about the defaulting behavior is happening than that the overriding values are coming from a Data Values file.

examples/schema-arrays/expected.txt
```diff
@@ -0,0 +1,18 @@
...
+- name: eirini-release
+  github_org: cloudfoundry-incubator
+  build_dir: build/eirini
+  github_release: true
+  vendir_github_release: true
+- name: metric-proxy
+  github_release: true
+  github_org: cloudfoundry
+  vendir_github_release: true
+  build_dir: ""
```
pivotaljohn: Is this file being used somewhere? I can't seem to find that usage.

examples/schema/expected.txt
```diff
@@ -0,0 +1,26 @@
...
+***
+nothing: a new string
+string: str
+bool: true
+int: 123
+float: 123.123
+any:
+- 1
+- 2
+- 4
```
pivotaljohn: Similar to question above: in other example directories, we have `expected.txt` but I believe those are each referenced in a test in some way.

test/e2e/e2e_test.go
cari-lynn: I modified `TestSchema()` to test all the different data values flags, similar to the `TestDataValues()` test works. This seems like the right thing to do since we aren't testing all these cases in [schema_test.go](https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/cmd/template/schema_test.go#L132-L175), and since in this examples [README.md](https://github.com/vmware-tanzu/carvel-ytt/pull/414/files#diff-d8bb9e365c5466d96d492fa351191026e428d1fda2502ce04c3d19073fc54efa) suggests using data values flags we have confidence that the example will work.

test/e2e/e2e_test.go
cari-lynn: Removed this test all together in favor of using [`expected.txt`](https://github.com/vmware-tanzu/carvel-ytt/pull/414/files#diff-7484cbd9a3d77b5a1209d32b5dc32729da0515dd6e0cc0ab3e2d74002f772169) to assert on the examples output.

examples/schema-arrays/expected.txt
cari-lynn: The `expected.txt` files are now being used [here](https://github.com/vmware-tanzu/carvel-ytt/pull/414/files#diff-d5d12b6ea07b0a78c9ad1987e1486c565ecf16839f77a93e91b09475cf6c0e99R146-R161)

examples/schema/expected.txt
cari-lynn: same as above

cari-lynn:
The two failing tests are expected since we renamed the `@schema/data-values` annotation. We should merge this after #425 which will make the failing tests pass.

pivotaljohn:
Manually rebased on `develop`.

pivotaljohn (approved):
LGTM

PR #425, opened 2021-06-10
Rename @schema/match data_values=True to @data/values-schema

gcheadle-vmware:


cari-lynn (approved):
Looks like all mention and usage of the annotation was updated 👍

pivotaljohn:
How about how _few_ changes there were?

I read this lack of shotgun surgery to be as an indication of proper isolation. 👍🏻

PR #428, opened 2021-06-11
Automate creating github release

DennisDenuto:
- Uses goreleaser to upload and generate draft release notes with assets attached, and preliminary release notes generated

Authored-by: Dennis Leon <leonde@vmware.com>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,100 @@
+name: goreleaser
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    # Set permissions of github token. See https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.0
+
+      - name: Run GoReleaser
+        # GoReleaser v2.5.0
+        uses: goreleaser/goreleaser-action@5e15885530fb01d81d1f24e8a6f54ebbd0fed7eb
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          version: 0.162.0
+          args: release --rm-dist --debug
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: actions/github-script@v4
+        id: get-checksums-from-draft-release
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          result-encoding: string
+          script: |
+            var crypto = require('crypto');
+            const { owner, repo } = context.repo;
+
+            // https://docs.github.com/en/rest/reference/repos#list-releases
+            // https://octokit.github.io/rest.js/v18#repos-list-releases
+            var releases = await github.repos.listReleases({
+              owner: owner,
+              repo: repo
+            });
+
+            var crypto = require('crypto')
+            var fs = require('fs')
+            const url = require('url');
+            const https = require('https');
+
+            checksums = {}
+            for (const r of releases["data"]) {
+              if (r.draft && `refs/tags/${r.tag_name}` == "${{ github.ref }}") {
+                for (const asset of r.assets) {
+                  var release_asset = await github.repos.getReleaseAsset({ headers: {accept: `application/octet-stream`}, accept: `application/octet-stream`, owner: owner, repo: repo, asset_id: asset.id });
+                  const hash = crypto.createHash('sha256');
+
+                  let http_promise = new Promise((resolve, reject) => {
+                    https.get(release_asset.url, (stream) => {
+                      stream.on('data', function (data) {
+                        hash.update(data);
+                      });
+                      stream.on('end', function () {
+                        checksums[asset.name]= hash.digest('hex');
+                        resolve(`${asset.name}`);
+                      });
+                    });
+                  });
+                  await http_promise;
+                }
+              }
+            }
+            console.log(checksums)
+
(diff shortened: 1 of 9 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+    tags:
+      - 'v*'
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      -
```
pivotaljohn: Just a question. This is legit YAML, but it's less common to put a newline between the "array item" and the contained "map".

I'm down for trying something different to see how it looks/feels.

At this moment, I'm finding it _less_ readable as the dashes (`-`)  seem slightly detached from the content of their step, by my eye... making it distracting for me.

It's one of those strange nits... has zero affect on the correctness, but some kind of meaningful impact to _this_ human.

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          result-encoding: string
+          script: |
+            var crypto = require('crypto');
+            const { owner, repo } = context.repo;
+
+            // https://docs.github.com/en/rest/reference/repos#list-releases
+            // https://octokit.github.io/rest.js/v18#repos-list-releases
```
pivotaljohn, line 46: 🙏🏻  — really appreciate it when folks point to the API docs.

.goreleaser.yml
```diff
@@ -0,0 +1,78 @@
+# This is an example .goreleaser.yml file with some sane defaults.
```
pivotaljohn, line 1: Thoughts on pruning configuration that (re)sets to defaults?

.goreleaser.yml
```diff
@@ -0,0 +1,78 @@
...
+  # Default is empty
+  sort: asc
+
+  filters:
+    # Commit messages matching the regexp listed here will be removed from
+    # the changelog
+    # Default is empty
+    exclude:
+      - '^docs:'
+      - typo
```
pivotaljohn: nit: end-of-file newline?

examples/version-constraint/config.yml
```diff
@@ -0,0 +1,4 @@
+#@ load("@ytt:version", "version")
+#@ load("@ytt:overlay", "overlay")
```
pivotaljohn: superfluous. 
```suggestion
```

examples/version-constraint/config.yml
```diff
@@ -0,0 +1,4 @@
...
+
+#@ version.require_at_least("0.0.0")
```
pivotaljohn: nit: end-of-line newline?

pkg/version/version.go
```diff
@@ -3,6 +3,4 @@
 
 package version
 
-const (
-	Version = "0.34.0"
-)
+var Version = "develop"
```
pivotaljohn, line 6: Question to understand: what does setting this to non-semver do for us?

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+        with:
+          fetch-depth: 0
+      -
+        name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16.0
+      -
+        name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@5e15885530fb01d81d1f24e8a6f54ebbd0fed7eb
```
pivotaljohn, line 27: yes: immutable SHA. 👍🏻 

Perhaps mention intended version?

Also, 2.6.1 is out; should we grab that upgrade right now? 🤷🏻

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          version: 0.162.0
+          args: release --rm-dist --debug
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
pivotaljohn, line 33: What are the permissions/scopes of this token?

pivotaljohn (commented):
In general, looks good.

Left some questions to speed a synchronous review.

Thanks for putting this together!!!

.github/workflows/release.yml
DennisDenuto: Yeah thats fair. I've changed it to not have a new line after the array item.

.github/workflows/release.yml
DennisDenuto: I've added a comment to specify which version it belongs to. I haven't bumped it yet. I would like to test that version bump first although reading the release notes it should be really safe (they seem like chore related changes). Assuming everything still works, i'll bump all the tools to keep goreleaser action in sync across the tools.

I'd like to submit that bump as a separate PR. I plan on working on it next dev day.

.github/workflows/release.yml
DennisDenuto: They are above on line 12. The token permissions are set to allow 'write' on 'content' resource. I added a comment to refer to the docs around setting permissions.

.github/workflows/release.yml
DennisDenuto: 🙇

.goreleaser.yml
DennisDenuto: I have found it re-assuring / valuable when reading this file, to see the different configuration options set (even if they are set to their default).

i.e. seeing `draft: true` in this file. 
If I were to remove this line, I think I would check the docs every time to make sure that the default is to publish a draft.

pkg/version/version.go
DennisDenuto: This was primarily to make it consistent with feedback received for imgpkg.

I am not sure it does technically much for us. It acts more as a placeholder that the version comes from 'somewhere else'. And that somewhere else is our build tooling (via go's LDD build flags).

DennisDenuto:
Thanks for the feedback. I've addressed most of it (i think)

Any feedback given here, I've also ported over to kapp and kbld. i'll eventually get it over to imgpkg too.

examples/version-constraint/config.yml
DennisDenuto: done

PR #431, opened 2021-06-14
Bug fix: Empty data values is struct instead of None

cari-lynn:
In v.0.31.0 we changed the type of an empty data values to be None, this PR changes it back to what it was previously: a struct.


pivotaljohn (approved):
LGTM.

Nice lagniappe: collocating `newEmptyDataValuesDocument()` alongside `NewEmptyDataValues()`.

Issue #430, opened 2021-06-11
ParseIP and ParseCIDR


tylerschultz:
**Describe the problem/challenge you have**

I would like to assert on the ipv4ness or ipv6ness of IPs and CIDRs so that an IPv4 CIDR or address is not passed where an IPv6 address/CIDR is required.

**Describe the solution you'd like**

We propose a `net` module that starts off with `net.parse_ip` and `net.parse_cidr`

net.parse_ip("192.168.0.2", return_error=True) => (ip or None, err_msg_string or None)
ip.is_v4() => True
print(ip) => "192.168.0.2"
ip.is_v6() => False
err_msg_string  => "<error message>"

net.parse_cidr("192.168.0.1/24") => (ip or None, ip_net or None, err_msg_string or None)
ip.is_v4() => True
print(ip) => "192.168.0.1"
ip.is_v6() => False
print(ip_net) => "192.168.0.0/24"
err_msg_string => "string is not a valid CIDR"

**Anything else you would like to add:**

We acknowledge that this interface is nothing like python's parsing functions. We've proposed this interface knowing that it would be implemented with golang.

This is a far from complete proposal, but feels like a starting point that allows for extending later.

This proposal was discussed in this [slack thread](https://kubernetes.slack.com/archives/CH8KCCKA5/p1623436787481700).

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

tylerschultz:
On the topic of `net.parse_ip("192.168.0.2", return_error=True)`, it was proposed that it return a Tuple.
Does anyone have an opinion on what the function would return if `return_error=False`? Would it force callers to declare the error variable even though they'd prefer it panic?

pivotaljohn:
Thoughts on names (just tossing alternatives out there)...
- instead of `net`, perhaps call the module `ip` — rationale: `net` implies to me network connectivity; `ip` is purely about manipulating Internet Protocol data objects.
  - `ip.parse_addr()`
  - `ip.parse_cidr()`
- individual types: `ipaddress`, `ipnetwork` — a CIDR is the notation that describes a sub-network.

In terms of parse failures...
Another angle on this could be for this module to attempt to parse the string in various ways... and only panic when nothing can be parsed from it...

[...]

PR #433, opened 2021-06-15
module for handling Internet Protocol data

tylerschultz:
- implements parse_cidr and parse_ip
- cidrs and ips know if they are ipv4 or ipv6
- 'panics' when invalid input is given

This PR is an initial iteration on issue:
#430

It is acknowledged that the issue is still open for discussion and is not approved for implemenation.

Co-authored-by: Tyler Schultz <tschultz@vmware.com>
Co-authored-by: Edwin Xie <exie@vmware.com>
Co-authored-by: Christian Ang <angc@vmware.com>


pkg/yamltemplate/filetests/ytt-library/net/parse-cidr-err.yml
```diff
@@ -0,0 +1,11 @@
+#@ load("@ytt:net", "net")
+
+ipv4:
+  cidr: #@ net.parse_cidr("not-a-valid-cidr")
+
++++
+
+ERR: 
+- net.parse_cidr: invalid CIDR address: not-a-valid-cidr
```
mcwumbly: Nice! I didn't know it was possible to validate errors like this in these tests.

tylerschultz:
We have added a commit that causes `net.parse_ip` to return empty string when an invalid ip string is passed instead of 'panicing'. We want to be able to check the ipv6ness of a url hostname. If the hostname is a name, then we don't want to panic if it does not parse as an ip. This is where the proposed `net.parse_ip("192.168.0.2", return_error=True)` would be appealing.

pkg/yttlibrary/all.go
```diff
@@ -31,6 +31,7 @@ func NewAPI(replaceNodeFunc tplcore.StarlarkFunc, dataMod DataModule,
 		"json":   JSONAPI,
 		"yaml":   YAMLAPI,
 		"url":    URLAPI,
+		"net":    NETAPI,
```
cppforlife: nit: NETAPI -> NetAPI

cppforlife:
> We want to be able to check the ipv6ness of a url hostname. If the hostname is a name, then we don't want to panic if it does not parse as an ip. This is where the proposed net.parse_ip("192.168.0.2", return_error=True) would be appealing.

@tylerschultz i think we should implement it since it's a bit weird that we got an ip that isnt real ip that's answering questions whether it is or not. thinking of following signature:

```
net.parse_ip("192.168.0.2", return_error=True) => (<ip>, <err>)

net.parse_ip("foo", return_error=True)         => (None, "some error")
net.parse_ip("192.168.0.2", return_error=True) => (ip, None)

net.parse_ip("foo")         => panic
net.parse_ip("192.168.0.2") => ip
```

good thing about tuples as return value is that they can be splatted into vars: `(ip, err) = net.parse_ip(...)`

tylerschultz:
To aid in discussions, here's something like what were trying to accomplish:

```
#@ load("@ytt:assert", "assert")
#@ load("@ytt:data", "data")
#@ load("@ytt:net", "net")
#@ load("@ytt:url", "url")

#@ if data.values.IP_FAMILY == "ipv6":
#@   if data.values.CLUSTER_CIDR && !net.parse_cidr(data.values.CLUSTER_CIDR).is_ipv6():
#@     assert.fail('invalid CLUSTER_CIDR "' + data.values.CLUSTER_CIDR + '", expected to be a CIDR of type "ipv6"'
#@   end
#@   if data.values.HTTP_PROXY:
#@     hostname = url.parse_url(data.values.HTTP_PROXY).hostname() # could be foo.bar or 192.168.0.1 or [fe80::1]
#@     ip, err = net.parse_ip(hostname, return_error=True)
#@     if !err && !ip.is_ipv6():
#@       assert.fail('invalid HTTP_PROXY "' + data.values.HTTP_PROXY + '", expected to be an address of type "ipv6"')
#@     end
#@   end
#@ end
```

@cppforlife I didn't catch the problem that we'd be returning a garbage IP object if the IP were to be a name. That makes returning a Tuple a nice way out.

tylerschultz:
One point of clarification, you are suggesting that when the kwarg `return_error=False` the function returns only an ip or panics, and when `return_error=True`, the function returns a Tuple? Different return values based on what is requested?

cppforlife:
modifying above snippet a little bit. assuming that when error is returned, ip/cidr would be None. (actual error doesnt seem to be needed for your use case.)

```
if data.values.IP_FAMILY == "ipv6":
  cidr, _ = net.parse_cidr(data.values.CLUSTER_CIDR or "", return_error=True)
  if cidr and not cidr.ip_ipv6():
    assert.fail('invalid CLUSTER_CIDR "' + data.values.CLUSTER_CIDR + '", expected to be a CIDR of type "ipv6"')
  end

  if data.values.HTTP_PROXY:
    hostname = url.parse_url(data.values.HTTP_PROXY).hostname() # could be foo.bar or 192.168.0.1 or [fe80::1]
    ip, _ = net.parse_ip(hostname, return_error=True)
    if ip and not ip.is_ipv6():
      assert.fail('invalid HTTP_PROXY "' + data.values.HTTP_PROXY + '", expected to be an address of type "ipv6"')
    end
  end
end
```

cppforlife:
> I didn't catch the problem that we'd be returning a garbage IP object if the IP were to be a name.

i was just pointing out that getting an ip object filled with something-invalid/nothing is not great imho (if ip object is constructed successfully it should be filled with ipv4 or ipv6). i know Go does this (return empty typed byte array) in this case and that api isnt great either.

> One point of clarification, you are suggesting that when the kwarg return_error=False the function returns only an ip or panics, and when return_error=True, the function returns a Tuple?

yeah, since `return_error=False` would be equivalent to not specifying it at all.

> Different return values based on what is requested?

yup, generally i wouldnt recommend that, but it might be a good sugar for ytt. alternatively we could introduce another top level function like `net.parse_ip_err(...)` which would always return tuples, and never panic. definitely makes it easier to describe function contract (especially if we add static typing down the road). what do yall think?

we should also probably think about how this pattern would interact with upcoming e.g. `@schema/validate net.parse_ip` (i guess when running validators, ytt should catch panics or accept functions that return tuples with an assumption that [1] is error) or other modules (may be something in overlays?).

pkg/yttlibrary/all.go
```diff
@@ -31,6 +31,7 @@ func NewAPI(replaceNodeFunc tplcore.StarlarkFunc, dataMod DataModule,
 		"json":   JSONAPI,
 		"yaml":   YAMLAPI,
 		"url":    URLAPI,
+		"net":    NetAPI,
```
pivotaljohn: naming nit: `net` conveys networking. Something like `ip` would more properly convey that this module provides decoding, calculations, and encoding of Internet Protocol "objects" (addresses, networks/CIDRs, ... maybe interfaces...)

For example:
```
ip.parse_addr("192.0.0.1")
ip.parse_cidr("192.0.2.1/24")
```

pivotaljohn:
> we should also probably think about how this pattern would interact with upcoming e.g. @schema/validate net.parse_ip (i guess when running validators, ytt should catch panics or accept functions that return tuples with an assumption that [1] is error) or other modules (may be something in overlays?).

I like reaching over to the schema feature to seek out any design pressure... 

yet, a function that parses feels a bit awkward, at first blush:

```
#@schema/validate ip.parse_cidr
CLUSTER_CIDR: ""
```

I do see where if I _just_ want to make sure that the node contains a string that is a well-formed CIDR, just being able to drop the parse function on there is nice... 🤔 

But, I could want to validate a number of things about a given value:
- there's whether it's an address or a network of addresses (i.e. CIDR)
- whether it's IPv4 or IPv6
- whether a given address is within a particular CIDR
- ...

I note that other modules will "panic" if they are unable to parse/decode.

Another angle is to provide a check type function that predicts whether the subsequent parse will succeed:

```
network = ip.parse_cidr(value) if ip.is_cidr(value) else None
if not ip.is_addr(value):
  assert.fail("expected a valid IP address, but got " + value ".")
end
```

cppforlife:
> Another angle is to provide a check type function that predicts whether the subsequent parse will succeed:

validating should do the parsing so we are effecctively doing work twice. it also comes with a downside that there is no way to get meaningful explanation of whats wrong (why was it invalid)

cppforlife:
assuming that we are catching "panics" via starlark's err return value (we should double check that) from lambda following variations should work:

> there's whether it's an address or a network of addresses (i.e. CIDR)

`via=lambda v: ip.parse_addr_err(v)[0] or ip.parse_cidr_err(v)[0]`
cannot do this: `via=lambda v: ip.parse_addr(v) or ip.parse_cidr(v)` since it would always panic on parse_addr if it's cidr. is that good? it does consistently align with url.parse, json.decode, etc.

> whether it's IPv4 or IPv6

`via=lambda v: ip.parse_addr(v).is_v4()`

> whether a given address is within a particular CIDR

`via=lambda v: ip.parse_cidr("10.0.0.0/16").includes(ip.parse_addr(v))`

pivotaljohn:
**re: cost of parsing twice**
> validating should do the parsing so we are effecctively doing work twice. 

🤷🏻, I'm valuing the UX of the API far more than any wasted cycles in our world, right now (remembering [`ytt` is currently very fast](https://kubernetes.slack.com/archives/CH8KCCKA5/p1616176816066100?thread_ts=1616169648.057800&cid=CH8KCCKA5)).

--

**re: using error messages**
> it also comes with a downside that there is no way to get meaningful explanation of whats wrong (why was it invalid)

It starts to make me wonder what a `ytt` "program" would do with such an error message...
I have a hard time imagining it doing _anything_ except fail in order to report it (which is what you get with a "panic").

For example, would one want to be able to take a string, attempt to parse an IP address out of it, realize there's a subnet mask, and then recover by parsing using CIDR notation?

That said, I'm not 100% (yet) ready to say that we wouldn't — in general — want the ability to capture and use a return message _within_ templating (e.g. emit a comment in the output that describes why a given value is "zero" for a non-critical configuration datum).

--

**re: readability**
The `..._err()` suffix and reaching for the 0th element from the tuple is less widely intuitive than a function with the `is_...()` prefix: the former requires the reader to learn than a tuple is returned, whereas the latter is broadly understood to return a `bool` without any other context.

--

**re: consistent behavior with similar library functions**
> cannot do this: via=lambda v: ip.parse_addr(v) or ip.parse_cidr(v) since it would always panic on parse_addr if it's cidr. is that good? it does consistently align with url.parse, json.decode, etc

Yes, I think that consistency is critical. 👍🏻  Otherwise, it would be frustrating to have to learn when a standard module function "panics" on fail or not.

--

**In Use**

- "Is it a IPv4 value?" (e.g. when no IPv6 stack is available)
     ```
    via=lambda v: (ip.is_addr(v) and ip.parse_addr(v).is_v4()) or (ip.is_cidr(v) and ip.parse_cidr(v).is_v4())
    ```

Back to your example, @tylerschultz ...

```
load("@ytt:assert", "assert")
load("@ytt:data", "data")
load("@ytt:ip", "ip")
load("@ytt:url", "url")

if data.values.IP_FAMILY == "ipv6":
  cidr = data.values.CLUSTER_CIDR
  if cidr and not (ip.is_cidr(cidr) and ip.parse_cidr(cidr).is_ipv6()):
    assert.fail('invalid CLUSTER_CIDR "' + cidr + '", expected to be a CIDR of type "ipv6"'
  end
  if data.values.HTTP_PROXY:
    hostname = url.parse_url(data.values.HTTP_PROXY).hostname() # could be foo.bar or 192.168.0.1 or [fe80::1]
    if hostname and not (ip.is_addr(hostname) and ip.parse_addr(hostname).is_v6()):
      assert.fail('invalid HTTP_PROXY "' + data.values.HTTP_PROXY + '", expected to be an address of type "ipv6"')
    end
  end
end

```

mcwumbly:
@pivotaljohn - should this line be removed in the example above? `ip, err = ip.parse_ip(hostname)`?

pivotaljohn:
> @pivotaljohn - should this line be removed in the example above? `ip, err = ip.parse_ip(hostname)`?

Yes, sir. Updated. Thank you.

cppforlife:
> 🤷🏻, I'm valuing the UX of the API far more than any wasted cycles in our world, right now (remembering ytt is currently very fast).

i dont have a concern that it's going to be slow (it's not). i have a concern that conceptually we are having to guard invocations that themselves should be able to tell us if something is wrong or not. as an exercise, how would you implement parse_ip and is_addr yourself in starlark (not go; for this thought exercise assume that all ip addresses are just single digit integer)? (answers a question if this pattern is something our users can embrace for their own functions).

tylerschultz:
Hi friends, I'm catching up on this thread here. I'll attempt to consolidate, with some of the proposals from the slack thread too.

There is the rename of net module to ip module. I think this makes sense and will make the rename if y'all would like to go that way.

Parsing - there are several approaches:
1) DK's - `cidr, e = ip.parse_cidr("<a cidr>", return_error=True)` and `cidr = ip.parse_cidr("<a cidr>")` 
2) JR's - `ip.is_cidr(cidr) and ip.parse_cidr(cidr).is_ipv6()`
3) `ip = ip.parse_ip("<the ip>")` and `ip, e = ip.parse_ip_err("<the ip>")`
4) General upcoming schema compatibility concerns

1 seems clever, but perhaps not what a programmer would expect. It would solve my use case.
2 seems straight forward, but it loses a little preference because of the called twice problem. It would solve my use case.
3 seems the least controversial, but makes the api noisier. It would solve my use case.
4 I don't know enough to have an opinion.

Are there other approaches I've missed? Are any approaches gaining favor?

mcwumbly:
FWIW, my preference is for option (2), and I think with the example provided it reads the best (per https://github.com/vmware-tanzu/carvel-ytt/pull/433#issuecomment-867669013).

flawedmatrix:
Just adding my two cents on this discussion:

I'm inclined to lean towards option (1) actually. I agree that the API in (3) is not intuitive and the `return_error` kwarg is a decent workaround. I also agree on the point that maybe we should not be guarding invocations that are themselves already guarded. The panic that results from parsing an invalid input should have helpful output, and this seems to be consistent with the pattern that the other modules have. If there needs to be more context added to this error, then the `return_error` API should suffice.

tylerschultz:
FYI - We've implemented & pushed option 1.

pivotaljohn:
_(Sorry for the delay... been rather heads-down focused on getting Schemas finished.)_

@tylerschultz summarized:
> Parsing - there are several approaches:
> 
> 1. DK's - `cidr, e = ip.parse_cidr("<a cidr>", return_error=True)` and `cidr = ip.parse_cidr("<a cidr>")`
> 2. JR's - `ip.is_cidr(cidr) and ip.parse_cidr(cidr).is_ipv6()`
> 3. `ip = ip.parse_ip("<the ip>")` and `ip, e = ip.parse_ip_err("<the ip>")`
> 4. General upcoming schema compatibility concerns
> 
> 1 seems clever, but perhaps not what a programmer would expect. It would solve my use case.
> 2 seems straight forward, but it loses a little preference because of the called twice problem. It would solve my use case.
> 3 seems the least controversial, but makes the api noisier. It would solve my use case.
> 4 I don't know enough to have an opinion.
> 
> Are there other approaches I've missed? Are any approaches gaining favor?

Thanks for the summarization to drive toward a conclusion, @tylerschultz. 👍🏻 

With regards to the API choices ...
- I think it's critical to consider not just what seems nice/easy/cool to me, but rather consider what would help make as many _other_ `ytt` users productive as quickly as possible.
- over my tenure on this project, I've gotten a lot of feedback from a sector of users that `ytt` has lots of surprises and requires a lot of learning to get out of "the suck" zone and into "the productive" zone. So, some of the intensity in my advocacy stems from hearing and empathizing with that feedback.

I'd rank the options:
- 2 — it is both the most obvious and least surprising... which means that it would be accessible to the widest range of users.
- 3 — given that the method name hints at the difference... it's not a huge thing to learn
- 1 — modifying the intended _shape_ of the return based on an optional parameter is not something I've run into ... this seems like it would be the most surprising API.

Also, I'm struggling to see the value of attempting to capture the error (i.e. return `err`).

What would help me get there is if we could come up some use cases where a template would _use_ the contents (and not just the presence) of an `err` value.

pivotaljohn:
Update:

I'll be out for some time, starting very soon.

I believe that I've articulated this perspective the best I can. That perspective values approachability (i.e. maximizing the number of users that could look at a chunk of code and successfully guess what it did) over seemingly most other concerns. There are many valid perspectives, this is one.

@cppforlife will be the one to make the final call, here.

---

@tylerschultz et al: I appreciate your patience and diligence as we wrestle through this API design work. In a project like this, it's particularly valuable to avoid breaking changes... so we're front-loading the effort by attempting to make our genuinely best guess on _this_ side of it. API design is hard — there are lots of factors (and some of which are mutually influencing) which makes predicting the consequences of a decision sometimes quite difficult to predict.  Thank you for your contribution.

flawedmatrix:
Thanks @pivotaljohn for your perspective! This does look like a pretty tough API design decision to make, so I think you or @cppforlife can make a more informed decision than I or @tylerschultz can at this point.

tylerschultz:
@pivotaljohn I was out the last few days myself. Thanks for the updates. I'm interested in doing what's best for the project, so I'm (we're?) happy to continue iterating on this PR. Working on this has been fun!

@cppforlife We'll update this PR when you find some time to weigh in.

Stay cool

cppforlife:
@tylerschultz just got back from vacay. will think/chat more this week, and will report here.

tylerschultz:
Hi Friends, I'm commenting here to bump this PR. 

@cppforlife had wanted to wait for @jtigger to return to the office for more API discussion. I think we're all back now. LMK if I can be of any help. I'm always happy to jump on a call.

pivotaljohn:
Summarizing the design work @cppforlife and I did on the 9th:

## Design Considerations
What the options, above, amount to is: the way that failures are provided as "errors" rather than "exceptions" (i.e. returning a value rather than ending the program).

On reflection, we realized there is a generic need for this in `ytt`. Use cases include:
- I want to test a function in Starlark that `fail()`s in certain conditions, how would I write a test that doesn't end the suite if such a condition would occur?
- I am writing a function for [`@schema/validate`](https://hackmd.io/pODV3wzbT56MbQTxbQOOKQ?view#schemavalidate); part of my validation logic wants to attempt to parse to determine if the value is well-formed... but if the parse function `fail()`s on ill-formed input, how do I capture that error? (note that we will want _all_ validations to execute so that clients can get a full report, rather than stop at the first one).
- aaaand, the use-case we've been entertaining in _this_ work: I want to check to see if a given input is of a certain kind (here, whether a hostname is a domain name or an IP address)... and it's _not_ a fatal condition if that parse is unsuccessful.

...

A two design principles lighting up strongly, here:
1. keep it simple
2. keep features orthogonal

Being guided by these helps reduce the risk of bugs and increases the expressiveness/power/usefulness of the tool.

## Suggested Approach
The simplest strategies we came up with is to provide the ability to "try" to invoke a function that might `fail()`s. If it does `fail()`, return an `err` value, instead of terminating the program.

That approach:
- _composes_ into all of the use cases enumerated, above.
- avoids requiring modules to provide both a version of a method that `fail()`s and another that returns an `err` (i.e. they need only provide one that `fail()`s, and the caller can decide if they want to catch it)
- employs a widely understood mechanism: that an "exception" (here, `fail()`) can be "caught", allowing the program to continue.

We note that it comes with risks:
- this is yet another language feature readers have to understand in order to grok template code;
- exception systems are commonly abused and creates accidental complexity (e.g. using exceptions for flow control);

We believe the upsides are worth those risks. We also believe we can mitigate some of it (through examples and documentation).

## Approach in Action

in terms of our shared example, we'd have:

```
...

#@ if data.values.IP_FAMILY == "ipv6":
#@   if data.values.HTTP_PROXY:
#@     hostname = url.parse_url(data.values.HTTP_PROXY).hostname() # could be foo.bar or 192.168.0.1 or [fe80::1]
#@     (ip, err) = assert.try_to(lambda: ip.parse_addr(hostname))
#@     if not err and not ip.is_ipv6():
#@       assert.fail('invalid HTTP_PROXY "' + data.values.HTTP_PROXY + '", expected to be an address of type "ipv6"')
#@     end
#@   end
#@ end
```
where:
- the invocation of `ip.parse_addr()` is deferred by wrapping it in a lambda (acting like a closure)
- a new builtin that takes a function, invokes it, catching any `fail()`, returning a tuple of the result and error (`assert.try_to(...)`).

---

## Next Steps

What we're proposing is to create a separate pull request in order to introduce this exception handling capability and make _this_ PR dependent on that work.

The new PR would:
1. add a new builtin to the `assert` module named `try_to` (see endnote 2)
2. that builtin takes one argument: a callable.
3. if the callable fails (i.e. returns an non-nil (Go) `error`), `try_to()` converts that `error` into its string equivalent and returns the tuple `(None, err.Error())
4. if the callable succeeds (i.e. returns a nil `error`), `try_to()` converts that into the tuple `(result, starlark.None)`

Thoughts?

---

_Endnote 1:_ There's prior art around custom handling of errors:
https://github.com/vmware-tanzu/carvel-ytt/blob/01e2421ff04a2fa22c634b7cb3eedf04e054d7aa/pkg/template/core/errs.go#L15-L35

_Endnote 2:_ ​`try` [is a reserved keyword in Starlark](https://github.com/google/starlark-go/blob/master/doc/spec.md#lexical-elements). We're also considered `catch`. `try_to` is semantically closer to Python's `try`: attempt this block of code.

tylerschultz:
This seems like a valid and versatile approach. Thanks for the detailed write up.

We are interested in PRing the `assert.try_to` functionality. Is this a PR we can/should work on?

pivotaljohn:
@tylerschultz asked:
> Is this a PR we can/should work on?

If you're game, that would be fantastic!

tylerschultz:
We've removed the `return-error=True` feature from this PR. This PR should be mergable at this point. As discussed above, we'll work on a separate PR to address the `assert.try_to` functionality.

pkg/yamltemplate/filetests/ytt-library/ip/parse-addr.yml
```diff
@@ -0,0 +1,22 @@
+#@ load("@ytt:ip", "ip")
+
+ipv4:
+  ip: #@ ip.parse_addr("192.0.2.1")
```
pivotaljohn:
```suggestion
  ip: #@ ip.parse_addr("192.0.2.1").string()
```

We established in #372 the convention of _not_ assuming how to encode the given value. In the current, we're presuming to encode as a string.
For now, we're preserving that decision by making custom types non-convertible by default, but also provide the string encoding (i.e. `string()` method).

Getting there means `AddrValue` satisfies:
https://github.com/vmware-tanzu/carvel-ytt/blob/214258be2519cdb66953c8ab2e3e234f879af8ca/pkg/template/core/starlark_value.go#L19-L21

... and does _not_ satisfy:

https://github.com/vmware-tanzu/carvel-ytt/blob/214258be2519cdb66953c8ab2e3e234f879af8ca/pkg/template/core/starlark_value.go#L13-L15

... and then providing a `string()` method on it.

(see `URLValue` and `URLUser`: https://github.com/vmware-tanzu/carvel-ytt/pull/372/files?file-filters%5B%5D=.go#diff-526a88c3f59e3e4dc6e4124408bee316ab1ddbb86001953e222fcd01e603280b )

pkg/yamltemplate/filetests/ytt-library/ip/parse-addr.yml
```diff
@@ -0,0 +1,22 @@
...
+  is_ipv4: #@ ip.parse_addr("192.0.2.1").is_ipv4()
+  is_ipv6: #@ ip.parse_addr("192.0.2.1").is_ipv6()
+ipv6:
+  ip: #@ ip.parse_addr("2001:db8::1")
```
pivotaljohn:
```suggestion
  ip: #@ ip.parse_addr("2001:db8::1").string()
```

(see note above)

pkg/yttlibrary/ip.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	return (&AddrValue{parsedIP, nil}).AsStarlarkValue(), nil
+}
+
+func (av *AddrValue) Type() string { return "@ytt:ip.addr" }
+
+func (av *AddrValue) AsStarlarkValue() starlark.Value {
+	m := orderedmap.NewMap()
+	m.Set("is_ipv4", starlark.NewBuiltin("addr.is_ipv4", core.ErrWrapper(av.IsIPv4)))
+	m.Set("is_ipv6", starlark.NewBuiltin("addr.is_ipv6", core.ErrWrapper(av.IsIPv6)))
```
pivotaljohn:
```suggestion
	m.Set("is_ipv4", starlark.NewBuiltin("ip.addr.is_ipv4", core.ErrWrapper(av.IsIPv4)))
	m.Set("is_ipv6", starlark.NewBuiltin("ip.addr.is_ipv6", core.ErrWrapper(av.IsIPv6)))
```

Let's namespace the methods to modules to ensure no inadvertent collisions.  (In retrospect, we should have done the same for the methods on `@ytt:url.url`).

pkg/yttlibrary/ip.go
```diff
@@ -0,0 +1,183 @@
...
+				"parse_cidr": starlark.NewBuiltin("ip.parse_cidr", core.ErrWrapper(ipModule{}.ParseCIDR)),
+			},
+		},
+	}
+)
+
+type ipModule struct{}
+
+// AddrValue stores a parsed IP
+type AddrValue struct {
```
pivotaljohn:
```suggestion
type IPAddrValue struct {
```

Rationale: the whole slew of types used in the ytt standard library live in the same Go package: `yttlibrary`. Use the type name namespace to make unambiguous (e.g. against a future AddrValue type from another ytt standard library module).

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr.yml
```diff
@@ -0,0 +1,37 @@
...
+  cidr_addr_is_ipv4: #@ ip.parse_cidr("192.0.2.1/24").addr.is_ipv4()
+  cidr_addr_is_ipv6: #@ ip.parse_cidr("192.0.2.1/24").addr.is_ipv6()
+ipv6:
+  cidr: #@ ip.parse_cidr("2001:db8::1/96")
+  cidr_net: #@ ip.parse_cidr("2001:db8::1/96").net
+  cidr_addr: #@ ip.parse_cidr("2001:db8::1/96").addr
+  cidr_is_ipv4: #@ ip.parse_cidr("2001:db8::1/96").is_ipv4()
+  cidr_is_ipv6: #@ ip.parse_cidr("2001:db8::1/96").is_ipv6()
+  cidr_addr_is_ipv4: #@ ip.parse_cidr("2001:db8::1/96").addr.is_ipv4()
+  cidr_addr_is_ipv6: #@ ip.parse_cidr("2001:db8::1/96").addr.is_ipv6()
```
pivotaljohn:
```suggestion
#@ addr, net = ip.parse_cidr("192.0.2.1/24")
ipv4:
  addr: #@ addr.string()
  addr_is_ipv4: #@ addr.is_ipv4()
  addr_is_ipv6: #@ addr.is_ipv6()
  net: #@ net.string()
  net_addr: #@ net.addr.string()
  net_is_ipv4: #@ net.addr.is_ipv4()
  net_is_ipv6: #@ net.addr.is_ipv6()
#@ addr, net = ip.parse_cidr("2001:db8::1/96")
ipv6:
  addr: #@ addr.string()
  addr_is_ipv4: #@ addr.is_ipv4()
  addr_is_ipv6: #@ addr.is_ipv6()
  net: #@ net.string()
  net_addr: #@ net.addr.string()
  net_is_ipv4: #@ net.addr.is_ipv4()
  net_is_ipv6: #@ net.addr.is_ipv6()
```

... this, after considerable deliberation ...

- the original lines 4 & 12 make it easy to validate and use a CIDR value in a single line.
- calculated network values usually include zero values for masked bits (e.g. in the given example, the configured network would typically be `"192.0.2.0/24"` (i.e. note the last octet is zero).
- the Go's `net.ParseCIDR()` yields not a CIDR object, but a tuple of IP address and network values; Python's `ipaddress.ip_network()` throws an exception when the supplied CIDR has host bits set.
- original types `CIDRValue` and `NetValue` significantly overlap in concept and API.
- authors are encouraged to separate validation of values from their use (when schema validations are implemented, they execute _prior_ to _any_ use in templates).

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr.yml
```diff
@@ -0,0 +1,37 @@
...
+  cidr_addr_is_ipv4: true
+  cidr_addr_is_ipv6: false
+ipv6:
+  cidr: 2001:db8::1/96
+  cidr_net: 2001:db8::/96
+  cidr_addr: 2001:db8::1
+  cidr_is_ipv4: false
+  cidr_is_ipv6: true
+  cidr_addr_is_ipv4: false
+  cidr_addr_is_ipv6: true
```
pivotaljohn:
```suggestion
ipv4:
  addr: 192.0.2.1
  addr_is_ipv4: true
  addr_is_ipv6: false
  net: 192.0.2.0/24
  net_addr: 192.0.2.0
  net_is_ipv4: true
  net_is_ipv6: false
ipv6:
  addr: 2001:db8::1
  addr_is_ipv4: false
  addr_is_ipv6: true
  net: 2001:db8::/96
  net_addr: 2001:db8::
  net_is_ipv4: false
  net_is_ipv6: true
```

pkg/yttlibrary/ip.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// CIDRValue stores a parsed CIDR
+type CIDRValue struct {
+	addr                 net.IP
+	net                  *net.IPNet
+	*core.StarlarkStruct // TODO: keep authorship of the interface by delegating instead of embedding
+}
+
+func (cv *CIDRValue) Type() string { return "@ytt:ip.cidr" }
```
pivotaljohn:
```suggestion
```

Remove the data structure holding the result of a CIDR parse (see rational in `parse-cidr.yml`, above).

What's left is a type `IPNetValue` that represents an IP network.

pivotaljohn (changes requested):
_(This review was informed by a video call. I attempted to fold in context co-developed in that call.)_

Thematically (rationale, inline):
- add `string()` to types rather than presume string encoding.
- when parsing CIDR, return the tuple of IP address and network rather than a type that holds the two.
- minor namespacing tweaks designed to ensure additional modules in the `ytt` standard library can be included without collision or change.

@tylerschultz + @flawedmatrix — thank you for taking the time to work out the details, here. Please reach out for any clarifications or issues.

pkg/yamltemplate/filetests/ytt-library/ip/parse-addr.yml
tylerschultz: fixed

pkg/yamltemplate/filetests/ytt-library/ip/parse-addr.yml
tylerschultz: fixed

pkg/yttlibrary/ip.go
tylerschultz: fixed

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr.yml
tylerschultz: fixed

pkg/yttlibrary/ip.go
tylerschultz: fixed

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr.yml
tylerschultz: fixed

pkg/yttlibrary/ip.go
tylerschultz: removed

tylerschultz:
Thanks for the thorough and thoughtful review. The changes discussed/requested have been pushed.

pkg/yttlibrary/ip.go
```diff
@@ -0,0 +1,142 @@
...
+	m := orderedmap.NewMap()
+	m.Set("is_ipv4", starlark.NewBuiltin("ip.addr.is_ipv4", core.ErrWrapper(av.IsIPv4)))
+	m.Set("is_ipv6", starlark.NewBuiltin("ip.addr.is_ipv6", core.ErrWrapper(av.IsIPv6)))
+	m.Set("string", starlark.NewBuiltin("ip.addr.string", core.ErrWrapper(av.string)))
+	av.StarlarkStruct = core.NewStarlarkStruct(m)
+	return av
+}
+
+func (av *IPAddrValue) ConversionHint() string {
+	return "IPAddrValue does not automatically encode (hint: use .string())"
```
pivotaljohn:
```suggestion
	return av.Type() + " does not automatically encode (hint: use .string())"
```

I failed to catch this one in [`URLValue`](https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/yttlibrary/url.go#L230), but let's refer to the type using the same string that the Starlark `type()` would return.

pkg/yttlibrary/ip.go
```diff
@@ -0,0 +1,142 @@
...
+func (inv *IPNetValue) AsStarlarkValue() starlark.Value {
+	m := orderedmap.NewMap()
+	m.Set("addr", starlark.NewBuiltin("ip.net.addr", core.ErrWrapper(inv.Addr)))
+	m.Set("string", starlark.NewBuiltin("ip.net.string", core.ErrWrapper(inv.string)))
+	inv.StarlarkStruct = core.NewStarlarkStruct(m)
+	return inv
+}
+
+func (inv *IPNetValue) ConversionHint() string {
+	return "IPNetValue does not automatically encode (hint: use .string())"
```
pivotaljohn:
```suggestion
	return inv.Type() + " does not automatically encode (hint: use .string())"
```

Let's refer to the type using the same string that the Starlark `type()` would return.

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr-addr-wont-encode.yml
```diff
@@ -0,0 +1,11 @@
+#@ load("@ytt:ip", "ip")
+
+#@ addr, net = ip.parse_cidr("192.168.0.1/24")
+net: #@ net
```
pivotaljohn: This test appears to be flipped with `parse-cidr-net-wont-encode.yml`.

That said, with `parse-addr-wont-encode.yml` in place another test against the same type seems redundant.

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr-addr-wont-encode.yml
```diff
@@ -0,0 +1,11 @@
...
+
++++
+
+ERR: 
+- __ytt_tplXXX_set_node: Unable to convert value: IPNetValue does not automatically encode (hint: use .string())
```
pivotaljohn:
```suggestion
- __ytt_tplXXX_set_node: Unable to convert value: @ytt:ip.net does not automatically encode (hint: use .string())
```

(see rationale in implementation, below)

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr-net-wont-encode.yml
```diff
@@ -0,0 +1,11 @@
+#@ load("@ytt:ip", "ip")
+
+#@ addr, net = ip.parse_cidr("192.168.0.1/24")
+addr: #@ addr
+
++++
+
+ERR: 
+- __ytt_tplXXX_set_node: Unable to convert value: IPAddrValue does not automatically encode (hint: use .string())
```
pivotaljohn:
```suggestion
- __ytt_tplXXX_set_node: Unable to convert value: @ytt:ip.addr does not automatically encode (hint: use .string())
```

(see rationale in implementation, below)

pivotaljohn (changes requested):
Looks great, thank you, gents.

One minor tweak required: to use the Starlark type name when we report the error to the user.... details in the comments.

Also, there's a super minor swap in logic in the tests.

Otherwise, we are GTG!

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr-addr-wont-encode.yml
tylerschultz: fixed

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr-net-wont-encode.yml
tylerschultz: fixed

pkg/yamltemplate/filetests/ytt-library/ip/parse-cidr-addr-wont-encode.yml
tylerschultz: We've updated the test name to match what's being tested. I'll let you decide if this is redundant. Let me know if you'd rather I delete the test.

pkg/yttlibrary/ip.go
tylerschultz: fixed

pkg/yttlibrary/ip.go
tylerschultz: fixed

tylerschultz:
FYI - this PR is now ready for another review/merge?

pivotaljohn (approved):
Excellent.

Thank you for hanging in there as we learn our way to a better set of patterns for standard library modules. 🙏🏻

PR #435, opened 2021-06-16
Fix schema error reporting position of definition instead of the position of the offending data value

pivotaljohn:


pivotaljohn:
Reviewed this synchronously with @cppforlife and @DennisDenuto.

Issue #424, opened 2021-06-10
Nested array in a schema has a default value


cari-lynn:
**What steps did you take:**

Given a schemas file that has an array item that is a map of at least two map items. One of those map items is a map, with the value of an array. 

Combined with a data values file that only provides one map item (essentially allowing the nested array's defaults to be filled in).

Using this schema:
```yaml
#@schema/match data_values=True
---
resources:
  - name: ""
    source:
      id:
        - I shouldn't show up
```
and this data values:
```yaml
#@data/values
---
resources:
  - name: ""
```
Run `ytt -f values.yml -f schema.yml --enable-experiment-schema --data-values-inspect`

**What happened:**
The default value for the array item 'I shouldn't show up' is populated
```yaml
resources:
- name: ""
  source:
    id:
    - I shouldn't show up
```
**What did you expect:**
The default value for an [array](https://carvel.dev/ytt/docs/latest/lang-ref-ytt-schema/#inferring-defaults-for-arrays) is an empty array.
```yaml
resources:
- name: ""
  source:
    id: []
```

**Anything else you would like to add:**

**Environment:**

- ytt version (use `ytt --version`): v0.34.0
- OS (e.g. from `/etc/os-release`):

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

pivotaljohn:
**Note:** at this time, we are aiming for GA. The probability or impact of this bug is such that we _could_ accept it as a known issue. We'd (obviously) rather fix it.

If this defect can be addressed in a day or less, let's do it. Otherwise, let's chat about what to do as a team.

PR #436, opened 2021-06-21
Bug Fix: Provide correct default value for nested arrays in schema

gcheadle-vmware:
Schema: Compute default values for map items
- recursive to catch nested arrays
- bug fix for #424

### DIFF
--- a/pkg/cmd/template/schema_test.go
+++ b/pkg/cmd/template/schema_test.go
@@ -672,7 +672,7 @@ func TestSchema_Provides_default_values(t *testing.T) {
 	opts := cmdtpl.NewOptions()
 	opts.SchemaEnabled = true
 
-	t.Run("values specified in the schema are the default data values", func(t *testing.T) {
+	t.Run("initializes data values to the values set in the schema", func(t *testing.T) {
 		schemaYAML := `#@data/values-schema
 ---
 system_domain: "foo.domain"
@@ -690,7 +690,7 @@ system_domain: #@ data.values.system_domain
 		})
 		assertSucceeds(t, filesToProcess, expected, opts)
 	})
-	t.Run("array default to an empty list", func(t *testing.T) {
+	t.Run("defaults arrays to an empty list", func(t *testing.T) {
 		schemaYAML := `#@data/values-schema
 ---
 vpc:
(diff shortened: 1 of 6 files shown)


pivotaljohn:
There's a case we aren't catching here (I think I saw this in a Usability Test session):

```patch
diff --git a/pkg/cmd/template/schema_test.go b/pkg/cmd/template/schema_test.go
index 5205d76..9e3097b 100644
--- a/pkg/cmd/template/schema_test.go
+++ b/pkg/cmd/template/schema_test.go
@@ -727,6 +727,8 @@ db_conn:
   port:
     job:
     - I should not show
+    #@schema/nullable
+    meals: 42
 `
                dataValuesYAML := `#@data/values
 ---
@@ -743,6 +745,7 @@ rendered: #@ data.values
   - hostname: server.example.com
     port:
       job: []
+      meals: null
 `

                filesToProcess := files.NewSortedFiles([]*files.File{
```

... results in this:

```
=== RUN   TestSchema_Provides_default_values/array_default_to_an_empty_list_when_grandparent_map_is_omitted
    require.go:157: 
        	Error Trace:	schema_test.go:2224
        	            				schema_test.go:757
        	Error:      	Not equal: 
        	            	expected: "rendered:\n  db_conn:\n  - hostname: server.example.com\n    port:\n      job: []\n      meals: null\n"
        	            	actual  : "rendered:\n  db_conn:\n  - hostname: server.example.com\n    port:\n      job: []\n      meals: 42\n"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -5,3 +5,3 @@
        	            	       job: []
        	            	-      meals: null
        	            	+      meals: 42
        	            	 
        	Test:       	TestSchema_Provides_default_values/array_default_to_an_empty_list_when_grandparent_map_is_omitted
```

I haven't taken the time to look deeper... I wonder if defaults need to be calculated later?!? ugh! Will look deeper into this when I get a minute.

gcheadle-vmware:
`@schema/nullable` is a way to explicitly set the default of a map item to `null`, so maybe information about defaults should be surfaced during the processing of the annotation?

PR #438, opened 2021-06-22
Set line position of data value flags

cari-lynn:
Why: In order to report better schema errors the source line is now saved

For all flags not read from a file the error is [reported](https://github.com/vmware-tanzu/carvel-ytt/pull/438/files#diff-2ab902ad23ffe96207ff6466403834df5538e509639024bfc503aed26a377cb8R562-R573) in the format `key=value`. 

Two cases I want to call out: 
### 1st case:
 When reporting an error from a data value read from a file eg: `--data-value-file foo=/path/file.yml` the line will be reported with only the file contents, and they key name and file name are reported at the top:
```
key 'foo' (key=file arg) dvs1.yml:
    |
  1 | not an integer
    |
```
Whereas other flags errors are reported exactly as they were input on the command line (eg: key=value). Please let me know if you have any feedback on this format.

### 2nd case: 
When targeting a library via `--data-value @lib:foo=4` the `@lib:` prefix is included in the error:
```
key 'foo' (kv arg):
         |
       1 | @lib:foo=42
```

### DIFF
--- a/pkg/cmd/template/data_values_flags.go
+++ b/pkg/cmd/template/data_values_flags.go
@@ -202,7 +202,8 @@ func (s *DataValuesFlags) env(prefix string, valueFunc valueTransformFunc) ([]*w
 
 		// '__' gets translated into a '.' since periods may not be liked by shells
 		keyPieces := strings.Split(strings.TrimPrefix(pieces[0], keyPrefix+envKeyPrefix), envMapKeySep)
-		overlay := s.buildOverlay(keyPieces, val, "env var", envVar)
+		desc := fmt.Sprintf("key '%s' (env var)", strings.Join(keyPieces, dvsMapKeySep))
+		overlay := s.buildOverlay(keyPieces, val, desc, envVar)
 
 		dvs, err := workspace.NewDataValuesWithOptionalLib(overlay, libRef)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pivotaljohn:
Good work, here, @cari-lynn.

As I was reading through the format of the error messages, I realized that there's some distance between the words the user employs to provide a data value on the command line and the words we use to describe them in an error message.

Playing around, I wonder what you think of this:

```
One or more data values were invalid
====================================

(data-value-file arg) foo=./dvs1.yml:
    |
  1 | not an integer
    |

    = found: string
    = expected: integer (by schema.yml:3)


(data-values-env arg) DVS:
    |
  1 | DVS_foo=not an integer
    |

    = found: string
    = expected: integer (by schema.yml:3)


(data-values arg):
    |
  1 | foo=42
    |

    = found: string
    = expected: integer (by schema.yml:3)


(data-values-yaml arg):
    |
  1 | foo=not an integer
    |

    = found: string
    = expected: integer (by schema.yml:3)
```

The latest two would be a little tricky, but I wonder if it made sense that `dataValuesFlagsSource` could include a `Name` which is the name of the source.

Thoughts?

cari-lynn:
Thanks @pivotaljohn. I like keeping the error messages wording consistent with the flag names 👍   
I also added the suggested change to `dataValuesFlagsSource`. Let me know what you think.

pivotaljohn (approved):
LGTM!

PR #443, opened 2021-06-24
Use parent node's position for template.replace()

cari-lynn:
This adds positions for arrays and maps used in `template.replace` so that they can be shown in schema error messages. It supports:
* Maps: `_: #@ template.replace({'map': { 'nestedMap': frag_func(), 'otherMap': 'two', 'array': ['three']}})`
* Arrays: `- #@ template.replace([{'key': 'not an integer'}])`

Does **not** support function calls (eg: `library.get("lib").with_data_values({'foo':'4'}).eval())`) that will likely be a separate PR.

### DIFF
--- a/pkg/cmd/template/schema_test.go
+++ b/pkg/cmd/template/schema_test.go
@@ -473,16 +473,25 @@ data_values.yml:
 
 		schemaYAML := `#@data/values-schema
 ---
-foo: 0
+map:
+  nestedMap:
+    key: 1
+  otherMap: 2
+  array:
+  - 3
 `
 		dataValuesYAML := `#@ load("@ytt:template", "template")
 #@data/values
 ---
-_: #@ template.replace({'foo':'not a integer'})
+#@ def frag_func():
+key: one
+#@ end
+
+_: #@ template.replace({'map': { 'nestedMap': frag_func(), 'otherMap': 'two', 'array': ['three']}})
 `
 		templateYAML := `#@ load("@ytt:data", "data")
 ---
-rendered: #@ data.values.foo
+rendered: #@ data.values.map
 `
 		expectedErr := `
 One or more data values were invalid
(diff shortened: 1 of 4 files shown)


cppforlife:
hmm, my general thinking around positions have been is to use unknown position when position is not known. imho trying to assign "nearby" positions leads to confusion since that's not what the position actually is. if we really want to add "nearby" positions we should probably have a way to add that to the Position object.

pivotaljohn:
> hmm, my general thinking around positions have been is to use unknown position when position is not known. imho trying to assign "nearby" positions leads to confusion since that's not what the position actually is. if we really want to add "nearby" positions we should probably have a way to add that to the Position object.

I'd like to better understand the value of interpreting `Position` strictly, here. I've been thinking of it as "the source from which this value originated."

In cases where a `Node` originates from parsed YAML, we can finger the _exact_ position. The only category of source that we can't (right now/ever?) pinpoint an exact location is when the values are originating from Starlark values.

On its face, I don't see the confusion of indicating that the "source" of a value in that latter category is from the expression which injects it into the AST.

What do we gain/retain/avoid by introducing a separate notion for "nearby" over making `Position` be a "best effort"?

cppforlife:
> What do we gain/retain/avoid by introducing a separate notion for "nearby" over making Position be a "best effort"?

the main concern i have is us providing wrong info which may confuse user even more (imho its worse to point user at one place and code is actually in a different place vs not pointing them at all). example that _may_ be problematic (will it?):

```
#@ def foo():
blah: something
#@end

foo: #@ template.replace(foo())
```

cari-lynn:
Currently the positions of Starlark expressions are reported on the line that the _expression is evaluated on_ not where the variable or function is defined.

Given a data value file where the types are not correct:
```
#@data/values
---
#@ var = "comes from variable"

#@ def func():
#@   return "10m"
#@ end

db_conn:
  port: #@ var
  timeout: #@ func()
```
The error reports the line that the _expression is evaluated on_:
```
data_values.yml:
	 |
      11 |   port: #@ var
	 |

	 = found: string
	 = expected: integer (by schema.yml:4)

data_values.yml:
	 |
      12 |   timeout: #@ func()
	 |

	 = found: string
	 = expected: float (by schema.yml:7)
```
This PR follows the same convention.

As far as reporting an incorrect position, I believe that the error should always use the line where the `template.replace()` was called, since the data value will always be set based on the result of that line.

Perhaps a way to improve this in the future would be to have a stack trace like:
```
data_values.yml:
	 |
       6 |   return "10m"
         |
in <toplevel>
data_values.yml:
	 |
      12 |   timeout: #@ func()
         |

	 = found: string
	 = expected: float (by schema.yml:7)
```
Or to use a column position, so that we can report part of the line caused the error, for example:
```
data_values.yml:
	 |
       1 |   --- #@ template.replace(library.get("lib").with_data_values({'foo':'4'}).eval())`
         |                                                               ^

	 = found: string
	 = expected: float (by schema.yml:7)
```

cppforlife:
lets roll i guess with what's being proposed, but down the line i think we need to firm up our rules. i understand that currently we do not track where things are defined in starlark (we only do it for yaml), but i think that's ultimately what user should see (place of definition not evaluation because "fixing" needs to happen at definition place)

cari-lynn:
Yeah highlighting the line that will solve the error would be a better experience. This does improve the experience over no line numbers and no source line, but there room to improve here.

pivotaljohn (approved):
Looks great to me, @cari-lynn. Excellent work, here. The diffs in the outputs speak volumes of value delivered through this work. 👍🏻

PR #452, opened 2021-07-05
Enable schemas

pivotaljohn:
- removes the feature flag for Schema, enabling Schema by default
- ensures that when Schema is present, are used; otherwise defaults to previous behavior (i.e. Data Values declaring variables)
- updates most examples to prefer `@data/values-schema` over `@data/values`.


pivotaljohn:
Noting some successful exploratory testing done with this build:
- configured [CF for K8s](https://github.com/cloudfoundry/cf-for-k8s)
  - with no modification to the library.
  - converting _only_ the root library to use schema.
  - converting _only_ the `capi` library (and not the root library) to use schema.
  - overlaying a new Data Value (via `with_data_value_schema()`) in `capi` library and then overriding that value (via `with_data_values()`).

cari-lynn (approved):
Reviewed this 'in person'. All the necessary changes were made, and the examples look good.  👍

PR #453, opened 2021-07-08
Allow null array items

pivotaljohn:
After a number of design conversations, it became clear that not allowing `@schema/nullable` on array items was an unnecessary constraint.

This PR removes that constraint (which allowed for a subsequent refactor).


pivotaljohn:
Synchronous review with @cari-lynn. Fixed some typos and whitespace.

PR #472, opened 2021-08-11
Populate missing info during library schema error messages

gcheadle-vmware:


pivotaljohn (approved):
With some minor changes (as discussed), LGTM.

PR #474, opened 2021-08-12
Add assert.try_to

### LINKED ISSUE #475: Add new `try_to` builtin to the `assert` module to enable handling cases when errors may be expected
per @pivotaljohn in https://github.com/vmware-tanzu/carvel-ytt/pull/433:

1. add a new builtin to the `assert` module named `try_to` (see endnote 2)
2. that builtin takes one argument: a callable.
3. if the callable fails (i.e. returns an non-nil (Go) `error`), `try_to()` converts that `error` into its string equivalent and returns the tuple `(None, err.Error())
4. if the callable succeeds (i.e. returns a nil `error`), `try_to()` converts that into the tuple `(result, starlark.None)`

---

Summarizing the design work @cppforlife and I did on the 9th:

## Design Considerations
What the options, above, amount to is: the way that failures are provided as "errors" rather than "exceptions" (i.e. returning a value rather than ending the program).

On reflection, we realized there is a generic need for this in `ytt`. Use cases include:
- I want to test a function in Starlark that `fail()`s in certain conditions, how would I write a test that doesn't end the suite if such a condition would occur?
- I am writing a function for [`@schema/validate`](https://hackmd.io/pODV3wzbT56MbQTxbQOOKQ?view#schemavalidate); part of my validation logic wants to attempt to parse to determine if the value is well-formed... but if the parse function `fail()`s on ill-formed input, how do I capture that error? (note that we will want _all_ validations to execute so that clients can get a full report, rather than stop at the first one).
- aaaand, the use-case we've been entertaining in _this_ work: I want to check to see if a given input is of a certain kind (here, whether a hostname is a domain name or an IP address)... and it's _not_ a fatal condition if that parse is unsuccessful.

...

A two design principles lighting up strongly, here:
1. keep it simple
2. keep features orthogonal

Being guided by these helps reduce the risk of bugs and increases the expressiveness/power/usefulness of the tool.

## Suggested Approach
The simplest strategies we came up with is to provide the ability to "try" to invoke a function that might `fail()`s. If it does `fail()`, return an `err` value, instead of terminating the program.

That approach:
- _composes_ into all of the use cases enumerated, above.
- avoids requiring modules to provide both a version of a method that `fail()`s and another that returns an `err` (i.e. they need only provide one that `fail()`s, and the caller can decide if they want to catch it)
- employs a widely understood mechanism: that an "exception" (here, `fail()`) can be "caught", allowing the program to continue.

We note that it comes with risks:
- this is yet another language feature readers have to understand in order to grok template code;
- exception systems are commonly abused and creates accidental complexity (e.g. using exceptions for flow control);

We believe the upsides are worth those risks. We also believe we can mitigate some of it (through examples and documentation).

## Approach in Action

in terms of our shared example, we'd have:

```
...

#@ if data.values.IP_FAMILY == "ipv6":
#@   if data.values.HTTP_PROXY:
#@     hostname = url.parse_url(data.values.HTTP_PROXY).hostname() # could be foo.bar or 192.168.0.1 or [fe80::1]
#@     (ip, err) = assert.try_to(lambda: ip.parse_addr(hostname))
#@     if not err and not ip.is_ipv6():
#@       assert.fail('invalid HTTP_PROXY "' + data.values.HTTP_PROXY + '", expected to be an address of type "ipv6"')
#@     end
#@   end
#@ end
```
where:
- the invocation of `ip.parse_addr()` is deferred by wrapping it in a lambda (acting like a closure)
- a new builtin that takes a function, invokes it, catching any `fail()`, returning a tuple of the result and error (`assert.try_to(...)`).

---

## Next Steps

What we're proposing is to create a separate pull request in order to introduce this exception handling capability and make _this_ PR dependent on that work.

The new PR would:
1. add a new builtin to the `assert` module named `try_to` (see endnote 2)
2. that builtin takes one argument: a callable.
3. if the callable fails (i.e. returns an non-nil (Go) `error`), `try_to()` converts that `error` into its string equivalent and returns the tuple `(None, err.Error())
4. if the callable succeeds (i.e. returns a nil `error`), `try_to()` converts that into the tuple `(result, starlark.None)`

Thoughts?

---

_Endnote 1:_ There's prior art around custom handling of errors:
https://github.com/vmware-tanzu/carvel-ytt/blob/01e2421ff04a2fa22c634b7cb3eedf04e054d7aa/pkg/template/core/errs.go#L15-L35

_Endnote 2:_ ​`try` [is a reserved keyword in Starlark](https://github.com/google/starlark-go/blob/master/doc/spec.md#lexical-elements). We're also considered `catch`. `try_to` is semantically closer to Python's `try`: attempt this block of code.

_Originally posted by @pivotaljohn in https://github.com/vmware-tanzu/carvel-ytt/issues/433#issuecomment-897081618_

tylerschultz:
assert.try_to receives a callable that traps and returns failure errors.
This allows templates to handle error cases instead of fast failing and
exiting.

Co-authored-by: Christian Ang <angc@vmware.com>

This PR is a result of discussion in #433
Resolves: #475

### DIFF
--- a/pkg/yamltemplate/filetests/ytt-library/assert-try-to.yml
+++ b/pkg/yamltemplate/filetests/ytt-library/assert-try-to.yml
@@ -1,0 +1,61 @@
+#@ load("@ytt:assert", "assert")
+#@ load("@ytt:json", "json")
+
+#@ def try_to_fail_data_error():
+#@   _, err = assert.try_to(lambda: json.decode("}junk"))
+#@   return err
+#@ end
+
+#@ def bad_ref():
+#@   return (None).foo
+#@ end
+
+#@ def try_to_fail_bad_ref():
+#@   _, err = assert.try_to(bad_ref)
+#@   return err
+#@ end
+
+#@ def try_to_succeed_lambda():
+#@   decoded, err = assert.try_to(lambda: json.decode('{"ytt":"rules"}'))
+#@   if err:
+#@      return err
+#@   end
+#@   return decoded
+#@ end
+
+#@ def try_to_succeed_tuple():
+#@   tupleResult, err = assert.try_to(lambda: (1,2))
+#@   if err:
+#@      return err
+#@   end
+#@   return tupleResult
+#@ end
+
+#@ def foo():
+#@   return "foo"
+#@ end
+
+#@ def try_to_succeed_function():
+#@   functionResult, err = assert.try_to(foo)
+#@   if err:
+#@      return err
+#@   end
+#@   return functionResult
+#@ end
+
+test_try_to_fail_data_error: #@ try_to_fail_data_error()
+test_try_to_fail_bad_ref: #@ try_to_fail_bad_ref()
+test_try_to_succeed_lambda: #@ try_to_succeed_lambda()
+test_try_to_succeed_function: #@ try_to_succeed_function()
+test_try_to_succeed_tuple: #@ try_to_succeed_tuple()
+
++++
+
+test_try_to_fail_data_error: 'json.decode: invalid character ''}'' looking for beginning of value'
+test_try_to_fail_bad_ref: NoneType has no .foo field or method
+test_try_to_succeed_lambda:
+  ytt: rules
+test_try_to_succeed_function: foo
+test_try_to_succeed_tuple:
+- 1
+- 2
--- a/pkg/yttlibrary/assert.go
+++ b/pkg/yttlibrary/assert.go
@@ -16,7 +16,8 @@ var (
 		"assert": &starlarkstruct.Module{
 			Name: "assert",
 			Members: starlark.StringDict{
-				"fail": starlark.NewBuiltin("assert.fail", core.ErrWrapper(assertModule{}.Fail)),
+				"fail":   starlark.NewBuiltin("assert.fail", core.ErrWrapper(assertModule{}.Fail)),
+				"try_to": starlark.NewBuiltin("assert.try_to", core.ErrWrapper(assertModule{}.TryTo)),
 			},
 		},
 	}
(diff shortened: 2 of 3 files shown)


pkg/yttlibrary/assert.go
```diff
@@ -36,3 +38,21 @@ func (b assertModule) Fail(thread *starlark.Thread, f *starlark.Builtin, args st
...
 }
+
+func (b assertModule) TryTo(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() != 1 {
+		return starlark.None, fmt.Errorf("expected exactly one argument")
+	}
+
+	lambda := args.Index(0)
+	if _, ok := lambda.(starlark.Callable); !ok {
+		return starlark.None, errors.New("expected argument to be a callable function or lambda")
```
cppforlife: lets change this a bit to be more consistent with [0]:

```
Expected argument to be a function, but was %T
```

https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/yttlibrary/overlay/assert_annotation.go#L113-L114

pkg/yttlibrary/assert.go
```diff
@@ -36,3 +38,21 @@ func (b assertModule) Fail(thread *starlark.Thread, f *starlark.Builtin, args st
...
+	}
+
+	var errVal starlark.Value = starlark.None
+	retVal, err := starlark.Call(thread, lambda, nil, nil)
+	if err != nil {
+		errVal = starlark.String(err.Error())
+	}
+	return starlark.Tuple{retVal, errVal}, nil
```
cppforlife, line 56: might be worth throwing a test for a function that returns a tuple. (i guess we will have a tuple within a tuple.)

pkg/yamltemplate/filetests/ytt-library/assert-try-to.yml
```diff
@@ -0,0 +1,39 @@
...
+
+#@ def try_to_succeed_function():
+#@   functionResult, err = assert.try_to(foo)
+#@   if err:
+#@      return err
+#@   end
+#@   return functionResult
+#@ end
+
+test_try_to_fail: #@ try_to_fail()
```
cppforlife: lets also add another test that accesses a unknown field like `(None).foo` (results in `NoneType has no .foo field or method` error)

pkg/yttlibrary/assert.go
cppforlife: nit: lets also just use fmt.Errorf. no need to import errors after that.

PR #476, opened 2021-08-18
golangci-linter: enable godoc checks for all new commits

pivotaljohn:
- bump golanci-lint version in the github action
- switch to revive as golint is deprecated
- change lingering nolint: golint to revive

@joe-kimmel-vmw explained:
- `fetch-depth: '0'` as otherwise the tooling will only grab a “shallow” clone from the most recent commit, and be unable to distinguish new code from old code
- `exclude-use-default: false`  iirc the godoc linter is an exclude-list by default, and this enables that linter by not using the default exclude list
- `new-from-rev: <sha>` lets you “grandfather” in all the old changes without godoc comments and only apply linter suite to new diffs (which it can distinguish bc we gave it infinite fetch-depth above).

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -13,8 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: '0'
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
-          version: v1.34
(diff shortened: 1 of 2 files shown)


joe-kimmel-vmw:
LGTM and probably easiest to merge relatively soon,  while `e0cbac...` is still the most recent commit.   or alternatively, may want to consult with the team and make sure everyone with work-in-progress is OK to rebase and/or start adding godocs to their open branches.

cari-lynn (approved):
LGTM

PR #483, opened 2021-09-04
Parse floats correctly and fix mistake from go-yaml/yaml#171

### LINKED ISSUE #290: Float parsing is unnecessarily restrictive
PR #171 restricted the possible float syntax, but it's not clear that it was the right thing to do.
The YAML spec itself provides a regexp for floats (`-?[1-9](\.[0-9]*[1-9])?(e[-+][1-9][0-9]*)?`) that is cast into doubt by the example immediately following (`2.3e4`) which is not matched by the regexp.

Also, JSON allows floats of the form `2e5` and YAML is supposed to be able to read JSON.

As it seems other implementations treat `[0-9]+e[0-9]+` as a string, perhaps we should do that too, despite JSON-incompatibility, but treat it as a float if there's a decimal point in the mantissa or a + or - sign before the exponent.

pivotaljohn:
The regular expression is copy & pasted form the one in the spec.
The change suggested in go-yaml/yaml#171 and integrated was improper.

Closes go-yaml/yaml#290

(cherry-pick of go-yaml/yaml@7b8349ac747c6a24702b762d2c4fd9266cf4f1d6)

Signed-off-by: John Ryan <jtigger@infosysengr.com>

### DIFF
--- a/pkg/yamlmeta/parser_test.go
+++ b/pkg/yamlmeta/parser_test.go
@@ -5,11 +5,11 @@ package yamlmeta_test
 
 import (
 	"fmt"
-	"github.com/k14s/difflib"
 	"os"
 	"strings"
 	"testing"
 
+	"github.com/k14s/difflib"
 	"github.com/k14s/ytt/pkg/filepos"
 	"github.com/k14s/ytt/pkg/yamlmeta"
 )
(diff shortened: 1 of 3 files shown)


cppforlife:
@jtigger lets add a test in yamlmeta/parser_test.go to avoid accidentally regressing on this bug.

pkg/yamlmeta/parser_test.go
```diff
@@ -5,11 +5,12 @@ package yamlmeta_test
 
 import (
 	"fmt"
-	"github.com/k14s/difflib"
 	"os"
 	"strings"
 	"testing"
 
+	"github.com/k14s/difflib"
+
```
cppforlife: nit: remove newline

pivotaljohn: done

PR #490, opened 2021-09-20
@schema/default annotation

gcheadle-vmware:


pkg/schema/annotations.go
```diff
@@ -87,6 +93,43 @@ func NewNullableAnnotation(ann template.NodeAnnotation, valueType yamlmeta.Type,
...
+			position:    pos,
+			description: fmt.Sprintf("expected @%v annotation to contain default value", AnnotationDefault),
+			expected:    fmt.Sprintf("%s (by %s)", inferredType.String(), inferredType.GetDefinitionPosition().AsCompactString()),
+			found:       "missing value",
+			hints: []string{
+				"value must be the same type as the annotated node",
+				"value must be in Starlark format, e.g.: {'key': 'value'}, True"},
+		}
+	}
+	val, err := core.NewStarlarkValue(ann.Args[0]).AsGoValue()
```
pivotaljohn: To call this feature "done", we need to support _any_ Starlark expression (e.g. a function call)
Do that in _this_ story? or write another story?

pkg/schema/annotations.go
```diff
@@ -87,6 +93,43 @@ func NewNullableAnnotation(ann template.NodeAnnotation, valueType yamlmeta.Type,
...
+	if err != nil {
+		return nil, schemaAssertionError{
+			error:       err,
+			position:    pos,
+			description: fmt.Sprintf("invalid @schema/default argument"),
+			expected:    "A valid starlark value",
```
pivotaljohn:
```suggestion
			expected:    "A valid Starlark value",
```

pkg/schema/schema.go
```diff
@@ -152,6 +158,25 @@ func getType(node yamlmeta.ValueHoldingNode) (yamlmeta.Type, error) {
 	return typeOfValue, nil
 }
 
+func getExplicitDefaultValue(node yamlmeta.ValueHoldingNode) (interface{}, error) {
```
pivotaljohn: ⭐  for the naming (extending the "implicit" concept)

pkg/schema/annotations.go
pivotaljohn: talking about this synchronously...

In fact, by the time this annotation is processed, all of the Starlark has been evaluated. Therefore the only other type are function types (i.e. lambdas). Those are already not supported in Data Values.

pivotaljohn (changes requested):
also (as discussed):

- disallow `@schema/default` on array item, erroring out explicitly.
- when we do the attendant documentation on this annotation, let's include an example of default an array within an array.
- support `@schema/default` on documents

pkg/schema/annotations.go
```diff
@@ -87,6 +93,43 @@ func NewNullableAnnotation(ann template.NodeAnnotation, valueType yamlmeta.Type,
...
+			position:    pos,
+			description: fmt.Sprintf("expected @%v annotation to contain value with type of annotated node", AnnotationDefault),
+			expected:    fmt.Sprintf("%s (by %s)", inferredType.String(), inferredType.GetDefinitionPosition().AsCompactString()),
+			found:       fmt.Sprintf("%T", ann.Kwargs[0]),
+			hints: []string{
+				"value must be the same type as the annotated node",
+				"value must be in Starlark format, e.g.: {'key': 'value'}, True"},
+		}
+	}
+	if len(ann.Args) == 0 {
```
cppforlife: what if there are 2 args

pkg/schema/schema.go
```diff
@@ -152,6 +158,25 @@ func getType(node yamlmeta.ValueHoldingNode) (yamlmeta.Type, error) {
...
+	}
+	annsCopy := append([]Annotation{}, anns...)
+	if len(annsCopy) == 0 {
+		return nil, nil
+	}
+
+	var defaultVal interface{}
+	for _, ann := range annsCopy {
+		if defaultAnn, ok := ann.(*DefaultAnnotation); ok {
+			defaultVal = defaultAnn.val
```
cppforlife: - shouldnt be accessing private fields on objects outside of objects
- shouldnt we be deep copying this value somehow?

pkg/schema/schema.go
```diff
@@ -97,8 +97,14 @@ func NewMapItemType(item *yamlmeta.MapItem) (*MapItemType, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	return &MapItemType{Key: item.Key, ValueType: typeOfValue, defaultValue: typeOfValue.GetDefaultValue(), Position: item.Position}, nil
+	defaultValue, err := getExplicitDefaultValue(item)
+	if err != nil {
+		return nil, err
+	}
+	if defaultValue == nil {
```
cppforlife: what if the explicit default is actually specified to be null?

pkg/schema/schema.go
```diff
@@ -152,6 +158,25 @@ func getType(node yamlmeta.ValueHoldingNode) (yamlmeta.Type, error) {
...
 
+func getExplicitDefaultValue(node yamlmeta.ValueHoldingNode) (interface{}, error) {
+	anns, err := collectAnnotations(node)
+	if err != nil {
+		return nil, NewSchemaError("Invalid schema", err)
+	}
+	annsCopy := append([]Annotation{}, anns...)
+	if len(annsCopy) == 0 {
+		return nil, nil
+	}
```
cppforlife: this copies a slice into a new slice, but that doesnt actually copy deep -- dont need this code.

pkg/yamlmeta/internal/yaml.v2/decode_test.go
```diff
@@ -700,7 +700,7 @@ var unmarshalTests = []struct {
 		M{"a": 123456e1},
 	}, {
 		"a: 123456E1\n",
-		M{"a": 123456E1},
+		M{"a": 123456e1},
```
cppforlife: is this go version difference, or something else?

pkg/cmd/template/schema_test.go
```diff
@@ -783,6 +783,60 @@ dvs1.yml:
 		assertFails(t, filesToProcess, expectedErr, cmdOpts)
 	})
 
+	t.Run("when a data value of the wrong type used in @schema/default annotation", func(t *testing.T) {
+
```
cppforlife, line 796: nit: remove newline

pkg/schema/annotations.go
gcheadle-vmware: Instead of leaving this untested error in the code, we want to panic to signal that something has unexpectedly gone wrong.

pkg/yamlmeta/internal/yaml.v2/decode_test.go
gcheadle-vmware: We think this might be from a dependency bump that did not get fully updated. You can see this happening on develop after running `./hack/build.sh` with `go version go1.16.8 darwin/amd64`.

pkg/schema/schema.go
gcheadle-vmware: Can you explain why we should consider deep copying this value?

pkg/schema/schema.go
gcheadle-vmware: Updated to allow null.

gcheadle-vmware:
@pivotaljohn , @cppforlife; This PR is ready to be looked at again, thank you for the first round of comments.

pkg/schema/schema.go
cppforlife: lets say if default is an map and that map gets modified by something else down the line (who knows if that happens in current/future code), then default value would be affected, which means next time default value is used, it will return wrong copy (aka modified value).

pkg/cmd/template/schema_test.go
```diff
@@ -2533,6 +2743,141 @@ foo: 0
...
+schema.yml:
+    |
+  5 | - bar
+    |
+
+    = found: @schema/default annotation on array item
+    = expected: @schema/default annotation to be on map item
+    = hint: place annotation on the map item containing this array item
+    = hint: default value should set the value for the array
```
pivotaljohn: Observing:
- with each hint that directs action, we want to avoid misleading the user
- I wonder if we can convey the message in fewer words (reducing the amount of reading the user needs to do)
- technically, an array can be held by _either_ a map item _or_ a document.

Something like...

```suggestion
Invalid schema - @schema/default not supported on array item
============================================================

schema.yml:
    |
  5 | - thing
    |

    = hint: do you meant to set a default value for the array?
    = hint: set an array's default by annotating its parent
```

pkg/schema/schema.go
```diff
@@ -97,8 +102,11 @@ func NewMapItemType(item *yamlmeta.MapItem) (*MapItemType, error) {
 	if err != nil {
 		return nil, err
 	}
-
-	return &MapItemType{Key: item.Key, ValueType: typeOfValue, defaultValue: typeOfValue.GetDefaultValue(), Position: item.Position}, nil
+	defaultValue, err := getValue(item, typeOfValue)
+	if err != nil {
+		return nil, err
+	}
+	return &MapItemType{Key: item.Key, ValueType: typeOfValue, defaultValue: defaultValue, Position: item.Position}, nil
```
pivotaljohn: nit: let's format this function just like the other `New<NodeType>Type`:

```suggestion
	
	defaultValue, err := getValue(item, typeOfValue)
	if err != nil {
		return nil, err
	}
	
	return &MapItemType{Key: item.Key, ValueType: typeOfValue, defaultValue: defaultValue, Position: item.Position}, nil
```

And the whitespace format seems easiest to read for me.

pkg/cmd/template/schema_test.go
```diff
@@ -783,6 +783,60 @@ dvs1.yml:
...
+====================================
+
+schema.yml:
+    |
+  4 | int: 1
+    |
+
+    = found: boolean
+    = expected: integer (by schema.yml:4)
+
```
pivotaljohn: This could be confusing to the uninitiated that the effective value of the node could be overridden by a `@schema/default`.

The deeper we go into implementing Schema, the more and more valuable the ability to reference a _range_ of source lines that we could display here. I suspect that's a mighty undertaking.

Until then, we at least should mention the potentially missing factor:

```suggestion
schema.yml:
    |
  4 | int: 1
    |

    = found: boolean
    = expected: integer (by schema.yml:4)
    = hint: is the default value set using @schema/default?

```

or similar.

pkg/cmd/template/schema_test.go
```diff
@@ -2533,6 +2743,141 @@ foo: 0
...
+schema.yml:
+    |
+  4 | foo: 0
+    |
+
+    = found: missing value
+    = expected: integer (by schema.yml:4)
+    = hint: value must be the same type as the annotated node.
+    = hint: value must be in Starlark format, e.g.: {'key': 'value'}, True.
```
pivotaljohn: These hints _are_ true, but I wonder if they are helpful prompts for the person who (apparently?) forgot to include a value in the annotation.

 ```suggestion
syntax error in @schema/default annotation

schema.yml:
    |
  4 | foo: 0
    |

    = found: missing value (in @schema/default above this item)
    = expected: integer (by schema.yml:4)
```

(also, seeing that we have multiple kinds of errors like this, trying out what it might look like to have a more generic header for errors like this)

File: pkg/cmd/template/schema_test.go [hunk 14]
```diff
@@ -2533,6 +2743,141 @@ foo: 0
...
+
+schema.yml:
+    |
+  4 | foo: 0
+    |
+
+    = found: 2 values
+    = expected: integer (by schema.yml:4)
+    = hint: value must be the same type as the annotated node.
+    = hint: value must be in Starlark format, e.g.: {'key': 'value'}, True.
```
pivotaljohn: Similar "critique" as above. When writing hints, I'm asking myself, "what is the most likely miscommunication with the user, here? what can we succinctly add to repair that ?"

```suggestion
Invalid schema
==============
syntax error in @schema/default annotation

schema.yml:
    |
  4 | foo: 0
    |

    = found: 2 values (in @schema/default above this item)
    = expected: integer (by schema.yml:4)
```

pkg/cmd/template/schema_test.go
```diff
@@ -2533,6 +2743,141 @@ foo: 0
...
+
+schema.yml:
+    |
+  4 | foo: 0
+    |
+
+    = found: starlark.Tuple
+    = expected: integer (by schema.yml:4)
+    = hint: value must be the same type as the annotated node.
+    = hint: value must be in Starlark format, e.g.: {'key': 'value'}, True.
```
pivotaljohn: - `starlark.Tuple` will likely be confusing to most users. Likely more broadly approachable if we refer to "keyword argument".
- suggestion includes attempt to refer to this collection of errors with the same title.

```suggestion
syntax error in @schema/default annotation

schema.yml:
    |
  4 | foo: 0
    |

    = found: (keyword argument in @schema/default above this item)
    = expected: integer (by schema.yml:4)
    = hint: this annotation only accepts one argument: the default value
    = hint: value must be in Starlark format, e.g.: {'key': 'value'}, True.
```

pivotaljohn (changes requested):
I remember skipping over the tests in our first round of review, but had forgotten to circle back.

I did some exploratory testing and came back with some suggestions that I hope will improve the messaging with our users. Take a look!

pkg/yamlmeta/internal/yaml.v2/decode_test.go
pivotaljohn: I suspect this slipped in on accident. It's related to the floating-point fix from a prior commit.

In Go, either case is permitted (i.e. `e` vs. `E`).

Also, we're explicitly excluding these tests in `ytt`'s suite.

I recommend that until we bring `k14s/yaml` fully up to the current state of `ytt`'s copy of that library, let's make no further changes to these files (i.e. remove this particular change).

pkg/schema/schema.go
gcheadle-vmware: Changed so that we deep copy values that satisfy the node interface.

pkg/cmd/template/schema_test.go
gcheadle-vmware: Tried out this change, and it would apply to all mismatched type errors, not just ones that had the `@schema/default` annotation on them. I had to change a fair amount of tests and left in a TODO to remove this hint eventually. lmk if that's what you were expecting or not.

pivotaljohn (approved):
LGTM.

pivotaljohn:
Read this over with @cppforlife. We're both GTG on this PR. 👍🏻 

...

Things to keep in mind, going forward:
- @gcheadle-vmware and I were wrestling with how to message to **Configuration Authors** when their `@schema/default` values are mistyped (especially given that we don't show the annotation in the error) without muddying the waters for **Configuration Consumers**. At this time, we settled for a hint ([example](https://github.com/vmware-tanzu/carvel-ytt/pull/490/files#diff-2ab902ad23ffe96207ff6466403834df5538e509639024bfc503aed26a377cb8R317))
  - but.... if we validate the type of a default value either _while_ constructing the schema, itself, or _en masse_ immediately after the schema is built (but before Data Value Preprocessing occurs), then we could provide an audience-specific error message.
- today, we generally panic when someone includes a function value (i.e. lambda) in the YAML structure. Ideally, we would _error_ when it's anticipated bad user input and _panic_ when it is the result of `ytt` coding error. ([example](https://github.com/vmware-tanzu/carvel-ytt/pull/490/files#diff-3cd3d8e2686249d524141b98b510ce20456b45eb99a8a1b082ebbd77602d862dR129))
- I _think_ we're deep copying when we need to, but I'm not 100% certain. _If others have similar uncertainty_, it would be worth taking a deliberate run at the flow of values through to ensure that when we share data from one structure to another we _are_ defensively copying. ... it _might_ be possible to come up with a specific pattern/technique that can make this easier to _know_ we have it correct in the future.

These could be incorporated in follow-up stories or be independent Issues/PRs and prioritized accordingly.

...

That said, this one has landed. 🎉

PR #501, opened 2021-10-07
Ensure assert.try_to returns `None` on failure

pivotaljohn:
... otherwise, if the wrapped function returns `nil` for the result
the variable assigned that rvalue becomes _undefined_.

### DIFF
--- a/pkg/yttlibrary/assert.go
+++ b/pkg/yttlibrary/assert.go
@@ -48,10 +48,9 @@ func (b assertModule) TryTo(thread *starlark.Thread, f *starlark.Builtin, args s
 		return starlark.None, fmt.Errorf("expected argument to be a function, but was %T", lambda)
 	}
 
-	var errVal starlark.Value = starlark.None
 	retVal, err := starlark.Call(thread, lambda, nil, nil)
 	if err != nil {
-		errVal = starlark.String(err.Error())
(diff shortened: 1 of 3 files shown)


pkg/yttlibrary/assert.go
```diff
@@ -51,6 +51,7 @@ func (b assertModule) TryTo(thread *starlark.Thread, f *starlark.Builtin, args s
 	var errVal starlark.Value = starlark.None
 	retVal, err := starlark.Call(thread, lambda, nil, nil)
 	if err != nil {
+		retVal = starlark.None
 		errVal = starlark.String(err.Error())
```
cppforlife: what do you think of doing explicit return here

return starlark.Tuple{starlark.None, starlark.String(err.Error()}, nil

and below

return starlark.Tuple{retVal, starlark.None}, nil

pivotaljohn: sure... will incorporate. More durably satisfies the "BuiltIn contract" (i.e. returning `starlark.Value`s) 👍🏻

cppforlife: typically a good thing anyway, otherwise it's similar to thinking about fallthrough cases in switch statements (and those things are ripe for introducing bugs)

pivotaljohn: done.

PR #517, opened 2021-10-21
Disallow both 'schema/any' and 'schema/nullable' on a node

gcheadle-vmware:


pkg/cmd/template/schema_test.go
```diff
@@ -2753,6 +2726,39 @@ foo: 0
...
+==============
+@schema/type, and @schema/nullable are mutually exclusive
+
+schema.yml:
+    |
+  5 | foo: 0
+    |
+
+    = found: both @schema/type, and @schema/nullable annotations
+    = expected: one of schema/type, or schema/nullable
```
pivotaljohn: This PR is highlighting a weakness in our overall error messaging around schema:
1. we might have baked in the assumption that only lines that have nodes can be reported on
2. there's currently no plumbing to present multiple lines in such error messages.

While this story isn't about these issues, it's being tainted by that weakness.

Let's chat about whether this is something to tackle right here or defer it.

pivotaljohn (commented):
Note comment about error message format.

Wondering what reasonably-scoped options we have here that would allow us to point not just _near_ where the issue is, but _on the line(s)_ where the issue is.

pkg/cmd/template/schema_test.go
pivotaljohn: We decided to defer the large error message improvements.

Thank you, @gcheadle-vmware for capturing that in #521.

Issue #325, opened 2021-03-17
Better error message when @data/values is on non-documents


dprotaso:
**Describe the problem/challenge you have**
ytt is spitting out the error:
```
ytt: Error: Expected YAML document to be annotated with data/values but was *yamlmeta.MapItem
```

But I have no idea what what file/line it is complaining about

gcheadle-vmware:
Hey @dprotaso,

Thanks for reaching out! That error message could definitely use more information regarding the context of the error. 

We are curious what command and files you used, that way we could get a better understanding of how this error happened. We are guessing that you wanted to provide a data values file but did not annotate the top of the file with `#@data/values`.

cc @DennisDenuto

dprotaso:
```
#@data/values
test: hello
```

`ytt -f` that file

gcheadle-vmware:
Thank you for the additional context. We can certainly run a story to improve this error message. 

A Data Values file is a YAML document annotated with `#@data/values`. So you'll need to add a document start "`---`" under the `#@data/values` annotation.

Proposed Story:
### Problem
When files are annotated using `@data/values`, but no document start is provided, `ytt` errors with an unspecific error message:  `Expected YAML document to be annotated with data/values but was *yamlmeta.MapItem`. This error message can be confusing because it does not refer to a line number nor a file name. 

### Acceptance Criteria
given `values.yml`:
```
#@data/values 

some-val: val
```
when I run `ytt -f values.yml`, I should see a clear error message telling me:
1. What went wrong
2. Where it went wrong
3. hint on how to fix the problem (nice to have)

Feel free to add a ":+1:" reaction to this comment if this story looks good to you.

pivotaljohn:
I agree with this diagnosis...

... and I'd even go further:
- direct the user's attention to _where_, first — mention the spot where error occurred at the beginning
- we're leaking details that are irrelevant to the user (they don't care and might be confused by mention of the Go package in which the struct lives); let's coalesce all our communication around a ubiquitous language
- omitting the `@` prefix misses an opportunity to improve visual parsing.

Something _like_,

```console
Found @data/values on Map Item (values.yml:3); only Documents can be annotated with @data/values
```

aaronshurley:
This improvement makes sense. We're going to move this to the Unprioritized Backlog for right now so that we can group it with similar work. Contributions are welcome :)

pivotaljohn:
Getting to actionable acceptance criteria:
[...]

PR #518, opened 2021-10-22
Better err msg when @ann is on wrong type of node 

vmunishwar:
#325

### DIFF
--- a/pkg/cmd/template/cmd_data_values_test.go
+++ b/pkg/cmd/template/cmd_data_values_test.go
@@ -442,7 +442,8 @@ str: str`)
 
 	out := opts.RunWithFiles(cmdtpl.Input{Files: filesToProcess}, ui)
 
-	expectedErrorMessage := "Expected YAML document to be annotated with data/values but was *yamlmeta.MapItem"
+	expectedErrorMessage := "Found @data/values on Map Item (data.yml:4); only Documents (---) can be annotated with @data/values"
+
 	require.EqualError(t, out.Err, expectedErrorMessage)
 }
 
--- a/pkg/cmd/template/cmd_data_values_test.go
+++ b/pkg/cmd/template/cmd_data_values_test.go
@@ -442,7 +442,7 @@ str: str`)
 
 	out := opts.RunWithFiles(cmdtpl.Input{Files: filesToProcess}, ui)
 
-	expectedErrorMessage := "Found @data/values on Map Item (data.yml:4); only Documents (---) can be annotated with @data/values"
+	expectedErrorMessage := "Found @data/values on map item (data.yml:4); only documents (---) can be annotated with @data/values"
 
 	require.EqualError(t, out.Err, expectedErrorMessage)
 }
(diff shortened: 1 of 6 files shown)


pkg/workspace/doc_extractor.go
```diff
@@ -89,8 +89,8 @@ func (v DocExtractor) checkNonDocs(val interface{}, annName template.AnnotationN
 				// TODO check for annotation emptiness
 				_, isDoc := node.(*yamlmeta.Document)
 				if !isDoc {
-					errMsg := "Expected YAML document to be annotated with %s but was %T"
-					return fmt.Errorf(errMsg, annName, node)
+					errMsg := "Found @%s on %s (%s:%d); only Documents (---) can be annotated with @%s"
+					return fmt.Errorf(errMsg, annName, node.UserFriendlyName(), node.GetPosition().GetFile(), node.GetPosition().LineNum(), annName)
```
cppforlife: take a look at https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/yttlibrary/overlay/op.go#L86 for how we typically stringify positions.

vmunishwar: Thank you. Found a method AsCompactString to use. I have updated the code.

pkg/cmd/template/cmd_data_values_test.go
```diff
@@ -442,7 +442,8 @@ str: str`)
 
 	out := opts.RunWithFiles(cmdtpl.Input{Files: filesToProcess}, ui)
 
-	expectedErrorMessage := "Expected YAML document to be annotated with data/values but was *yamlmeta.MapItem"
+	expectedErrorMessage := "Found @data/values on Map Item (data.yml:4); only Documents (---) can be annotated with @data/values"
+
```
pivotaljohn:
```suggestion
	expectedErrorMessage := "Found @data/values on map item (data.yml:4); only documents (---) can be annotated with @data/values"

```

Let's make node types lowercase as they are ordinary (not proper) nouns.

pkg/yamlmeta/ast.go
```diff
@@ -27,7 +27,10 @@ type Node interface {
 
 	Check() TypeCheck
 
+	UserFriendlyName() string
```
pivotaljohn:
```suggestion
	DisplayName() string
```

Played with this together: is `DisplayName` a better name for this method?

pivotaljohn (changes requested):
Just a few renames and this will be good to go!

pkg/cmd/template/cmd_data_values_test.go
vmunishwar: Thank you. I have made the changes.

pkg/yamlmeta/ast.go
vmunishwar: Done. Thank you.

pivotaljohn (approved):
Congratulations on your first merged contribution to `ytt`, @vmunishwar!

This looks great. Every internal detail we can convert into something more intuitive (like referring to a thing as a "map item" rather than a "yamlmeta.MapItem") makes `ytt` _that_ much friendlier and easier to use. 👍🏻

PR #519, opened 2021-10-25
Move schema logic from yamlmeta package into schema package

pivotaljohn:
Overall Goals:
1. restore `yamlmeta` responsibilities to be no more than construction and printing of ASTs to/from YAML.
2. set the stage to be able to type check values being set on AST _while_ those values are being set (e.g. via an overlay or programmatically).

To do:

- [x] replace "type" field from Node with more generic "meta"
- [x] move all schema logic from `yamlmeta` package to `schema` package
- [x] move all data values-specific schema logic from `schema` to `datavalues` package


pivotaljohn:
At this point, converted one test to explore readability. If this approach is agreeable, intend to absorb all Node construction through these builders.

pkg/yamlmeta/builder.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// DocumentBuilder is builder used to configure and produce a Document.
+// Use NewDocumentBuilder() to create instances.
+type DocumentBuilder struct {
+    newDoc *Document
+}
+
+// NewDocumentBuilder initializes a builder that produces a Document
+func NewDocumentBuilder() *DocumentBuilder {
```
cppforlife: i would be curious to see how builder+struct-init approach feels like e.g. 

```go
NewDocument(&yamlmeta.Document{
  Comments ...
  ...
})
```

pivotaljohn: Interesting. The overall plan is to ultimately unexport those fields. The idea is to channel access through methods so that additional behaviors can be attached to such actions (e.g. when an overlay writes to an AST, triggering a type check of the resulting subtree).

Do you know of a way to do what you're suggestion without exporting those fields? Or have another idea on how we can make clear that modifications ought not be done through field access?

cppforlife: you dont have to use same object for "input" eg.

```go
NewDocument(yamlmeta.DocumentOpts{
  Comments ...
  ...
})
```

effectively you end up having shadow structs purely for instantiation.

pivotaljohn:
Evaluating this other approach of creating `Node`s...

The "overhead" of having a parallel [Composite](https://en.wikipedia.org/wiki/Composite_pattern) of [Prototypes](https://en.wikipedia.org/wiki/Prototype_pattern) seems nominal to me. 👍🏻 

Thinking through maintenance of this approach: 
- once all of the fields of the `Node` composite are unexported, any needs for additional data will necessarily plumb _through_ the `...Proto` composite: so the two families of `struct`s will naturally stay in sync.
- that said, the abstraction that is `Node` seems particularly stable: changes appear to be unlikely/rare.

I'm poised to roll out this change across the codebase. How's this looking to you, @cppforlife?

pkg/yamlmeta/construct.go
```diff
@@ -0,0 +1,150 @@
...
+	Value    interface{}
+	Comments []*CommentProto
+	Position *filepos.Position
+}
+
+// CommentProto is a prototype for a Comment
+type CommentProto struct {
+	Data     string
+	Position *filepos.Position
+}
```
cppforlife, line 150: would be good to apply forbidding marshaling/unmarshaling to protos: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/yamlmeta/node.go#L341-L359

cppforlife:
> I'm poised to roll out this change across the codebase. How's this looking to you, @cppforlife?

one more question... should there be any deep copying involved anywhere here?

pivotaljohn:
> one more question... should there be any deep copying involved anywhere here?

I'm pretty sure we are. I'll double-check before I proceed with that front.

In the meantime, I've worked through an approach for extracting schema information/logic from `yamlmeta`. Note that there's a number of shims in place, at this time; there's a lot of code clean-up I intend. I'm just currently narrowly focused on getting logic into the appropriate package without breaking anything.

gcheadle-vmware (commented):
Overall I think this refactor captures the schema creation and check logic into it's own package, separate from `yamlmeta`, and creates a stronger connection between dataValues and Schema.
Additional notes I took while review changes:
- `data values.Envelope` -> connects schema and datavalues as they are both kept in envelopes. 
- The responsibility for checking a Node is now in schema package, and changing the old 'type' to 'meta' simplifies the `yamlmeta` package while making the purpose of schema more clear.
- `collectDataValuesDocs` and `collectSchemaDocs` are in the same package and named similarly, but do different things, maybe a rename for one of these?

Besides my last bullet point, I did not see anything that set off alarms, and I think that refactor should make working in schema easier for people who have not seen it. Lastly, I had trouble understanding the purpose of the `Proto` type structs (DocumentProto, MapProto, ...). When searching for usages of these structs, I did not find anything outside of the `construct.go` file. Not sure if this is something I am missing something here, but I would love to heard the motivation behind that implementation.

pivotaljohn:
> Lastly, I had trouble understanding the purpose of the `Proto` type structs (DocumentProto, MapProto, ...). When searching for usages of these structs, I did not find anything outside of the `construct.go` file. Not sure if this is something I am missing something here, but I would love to heard the motivation behind that implementation.

Yeah... that was a sketch for what will come in a follow-up PR: instead of creating AST by directly instantiating types of `Node` (which requires write access to fields in those structs), create AST by building the exact same structure and then call a constructor with that input. The constructor builds the _actual_ AST from the prototype.

_This_ PR got to be so big, I decided to split the rest of the work into a subsequent PR. In the follow PR, this technique will be pushed out to all places where AST is being built.

pivotaljohn:
@gcheadle-vmware pointed out:
> * `collectDataValuesDocs` and `collectSchemaDocs` are in the same package and named similarly, but do different things, maybe a rename for one of these?

Nice catch!

- `collectDataValuesDocs` _both_ collects from the library all the Documents annotated `@data/values` and returns the pile of envelopes for them.
- `collectSchemaDocs` _given_ all the files that contain Documents annotated `@data/values-schema` produces the corresponding envelopes for them.

In general, there's an opportunity for the handling/treatment of Data Values and Data Values Schema to be shaped as similarly as possible/reasonable. I suspect the differences that exist are not due to an intrinsic difference in how they should be handled.

For example, I can imagine something like `DataValuesSchemaPreProcessing` which basically mirrors the behavior of `DataValuesPreProcessing` as appropriate. _Ideally_ the only differences between the two ought to be what's intrinsically different in the handling of Data Values versus their Schema.

I propose we tackle this in a subsequent PR that has this intention.

Thoughts?

gcheadle-vmware:
> For example, I can imagine something like `DataValuesSchemaPreProcessing` which basically mirrors the behavior of `DataValuesPreProcessing` as appropriate. _Ideally_ the only differences between the two ought to be what's intrinsically different in the handling of Data Values versus their Schema.
> 
> I propose we tackle this in a subsequent PR that has this intention.

This feels like a solid direction to go in the future, and I think it fits the theme of making a stronger connection between schemas and datavalues. 👍

pivotaljohn:
> > For example, I can imagine something like `DataValuesSchemaPreProcessing` which basically mirrors the behavior of `DataValuesPreProcessing` as appropriate. _Ideally_ the only differences between the two ought to be what's intrinsically different in the handling of Data Values versus their Schema.
> > I propose we tackle this in a subsequent PR that has this intention.
> 
> This feels like a solid direction to go in the future, and I think it fits the theme of making a stronger connection between schemas and datavalues. 👍

Awesome. Thank you, sir.

PR #525, opened 2021-10-27
Generate OpenAPI v3.0 schema for Arrays

cari-lynn:
Important takeaways: 
1. maps and arrays supported 
2. `@schema/type any=True` and `@schema/nullable` not yet supported
3. known issue where functions must be defined prior to `@data/values-schema` definition (not introduced in this PR) (covered in #526)


pkg/schema/openapi.go
```diff
@@ -29,15 +29,13 @@ func (o *OpenAPIDocument) AsDocument() *yamlmeta.Document {
 		{Key: "info", Value: &yamlmeta.Map{Items: []*yamlmeta.MapItem{
 			{Key: "version", Value: "0.1.0"},
 			{Key: "title", Value: "Schema for data values, generated by ytt"},
-		}},
-		},
+		}}},
```
cppforlife, line 32: 🎊

pkg/schema/annotations.go
```diff
@@ -194,6 +194,12 @@ func processOptionalAnnotation(node yamlmeta.Node, optionalAnnotation template.A
 			}
 			return typeAnn, nil
 		case AnnotationDefault:
+			if _, ok := node.(*yamlmeta.ArrayItem); ok {
```
cppforlife: should this check for yamlmeta.Map?

cppforlife: any reason why this error should not be inside NewDefaultAnnotation?

pivotaljohn: > should this check for yamlmeta.Map?

`ytt`'s parser doesn't place annotations on Nodes with an unknown position (this effectively means Maps and Arrays).
https://github.com/vmware-tanzu/carvel-ytt/blob/030dd4730c6639eb99e890cf042b3dfb3e6f53d5/pkg/yamlmeta/parser.go#L238-L240

Given that post-condition from the parser, why would we check for such nodes?

...

> any reason why this error should not be inside NewDefaultAnnotation?

Yeah, this was a choice @cari-lynn and I thought through and made consciously:
- it felt inappropriate to require a `Node` to construct the annotation
- the error checking _insider_ the constructor is about the contents of the annotation itself; seems fitting that `processOptionalAnnotation()` (or whatever method this ultimately turns out to be) holds the responsibility to determine which kinds of `Node` can accept such an annotation.

cppforlife: > ytt's parser doesn't place annotations on Nodes with an unknown position (this effectively means Maps and Arrays).

_current internal_ limitations of one pieces of code should not influence decisions typically in another. Maps go through this code path and they can hold annotations.

cppforlife: (accidently deleted my prev comment)

> ytt's parser doesn't place annotations on Nodes with an unknown position (this effectively means Maps and Arrays).

_current internal_ limitations of one piece of code should not influence decisions of another. Map objects do go through this code path, and they can hold annotations based on their API.

cppforlife: > Yeah, this was a choice @cari-lynn and I thought through and made consciously:

ok, sounds good.

pivotaljohn: > _current internal_ limitations of one piece of code should not influence decisions of another. Map objects do go through this code path, and they can hold annotations based on their API.

Yes; suspected that was the motivation, but wanted to ask and not assume.

Same would be true for yamlmeta.Array and yamlmeta.DocumentSet as well. We'll include these checks.

FYI: there's likely other places where we've assumed that only `ValueHoldingNode`s are annotated; we can capture the intention to be repair that in existing code in a separate story.

pivotaljohn: updated.

pivotaljohn:
https://kubernetes.slack.com/archives/CH8KCCKA5/p1635866136257400?thread_ts=1635865969.257000&cid=CH8KCCKA5

PR #531, opened 2021-11-01
Include "nullable" in schema inspect to OpenAPI v3

pivotaljohn:


pivotaljohn:
In "draft" because there's some work to be done on Schema itself to properly support this use case.

cppforlife:
lgtm

Issue #365, opened 2021-04-20
Generate OpenAPI Schema for `any` type


joaopapereira:
**As a** Configuration Author
**I want** to generate OpenAPI from my configuration
**so that** Configuration Consumers can use it.

---

### :green_circle: Scenario 1: Convert data values schema with any annotation on Document to OpenAPI schema

**Given** the following schema exists
```yaml
#! schema.yml
#@data/values-schema
#@schema/type any=True
---
foo:
  int_key: 10
```

**When** I execute `ytt -f . --data-values-schema-inspect -o openapi-v3`
**Then** I should see
```yaml
---
openapi: 3.0.0
info:
  version: 0.1.0
  title: Openapi schema generated from ytt Data Values Schema
paths: {}
components: 
  schemas:
    dataValues:
      default: 
        foo:
          int_key: 10
```

### :green_circle: Scenario 2: Convert data values schema with any annotation on Map to OpenAPI schema

**Given** the following schema exists
```yaml
#! schema.yml
#@data/values-schema
---
#@schema/type any=True
foo:
  int_key: 10
```

**When** I execute `ytt -f . --data-values-schema-inspect -o openapi-v3`
**Then** I should see
```yaml
---
openapi: 3.0.0
info:
  version: 0.1.0
  title: Openapi schema generated from ytt Data Values Schema
paths: {}
components: 
  schemas:
    dataValues:
      type: object
      properties:
        foo:
          default: 
            int_key: 10
```

### :green_circle: Scenario 3: Convert data values schema with any annotation on Array to OpenAPI schema

**Given** the following schema exists
```yaml
#! schema.yml
#@data/values-schema
---
#@schema/type any=True
foo:
- int_key: 10
```

**When** I execute `ytt -f . --data-values-schema-inspect -o openapi-v3`
**Then** I should see
```yaml
---
openapi: 3.0.0
info:
  version: 0.1.0
  title: Openapi schema generated from ytt Data Values Schema
paths: {}
components: 
  schemas:
    dataValues:
      type: object
      properties:
        foo:
          default: 
          - int_key: 10
```

### Key Notes from Resources

- [JSON Schema Validation §4.2: "Missing Keyword"](https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00#section-4.2)
  > Validation keywords that are missing never restrict validation.  In some cases, this no-op behavior is identical to a keyword that exists with certain values, and these values are noted where known.
[...]

PR #535, opened 2021-11-03
"Any" explicitly includes `nullable`

pivotaljohn:
- by specification, `nullable` can be interpreted as implied.
- however, there is at least one implementation in the wild that would
  mark a `null` input as invalid.
- for other implementations, this property/keyword would be ignored.

(in response to analysis: https://github.com/vmware-tanzu/carvel-ytt/issues/365#issuecomment-959660526)


cppforlife:
lgtm

PR #540, opened 2021-11-09
Refactor `schema_test` into consumer and author tests

cari-lynn:
To make this test file smaller I split it into to themes. 
- One for for configuration authors and one for configuration consumers. The distinction shows up because configuration consumers will include a data value file, and the results of that test depend on the contents of the data values. 
- Configuration authors do not need a data values file, and deals with simply providing values as defaults, and the error messages from syntax of the schema file itself.

I am not attached to this honestly, so let me know your thoughts or other ways to split the files.

impl note: I deleted one test that was redundant. See commits for details


pkg/cmd/template/schema_author_test.go
```diff
@@ -0,0 +1,671 @@
...
+func TestSchema_When_invalid_reports_error(t *testing.T) {
+	opts := cmdtpl.NewOptions()
+
+	t.Run("array with fewer than one element", func(t *testing.T) {
+		schemaYAML := `#@data/values-schema
+---
+vpc:
+  subnet_ids: []
+`
+		dataValuesYAML := `#@data/values
```
gcheadle-vmware: Should we have a data/values file when we are testing behavior around authoring schema?

pkg/cmd/template/schema_author_test.go
```diff
@@ -0,0 +1,671 @@
...
+		assertFails(t, filesToProcess, expectedErr, opts)
+	})
+	t.Run("when schema/type and schema/nullable annotate a map", func(t *testing.T) {
+		schemaYAML := `#@data/values-schema
+---
+#@schema/type any=True
+#@schema/nullable
+foo: 0
+`
+		templateYAML := `#@ load("@ytt:data", "data")
```
gcheadle-vmware: Maybe `templateYAML` is unnecessary in this test?

gcheadle-vmware (approved):
I really like this change since it simplifies the schema test files and reduces total lines of code.

pkg/cmd/template/schema_author_test.go
cari-lynn: good catch. This test and the one following it have data values files that are not being used in the test. I'll update to remove the data values files.

pkg/cmd/template/schema_author_test.go
cari-lynn: good catch, removed

cppforlife:
nice!

pivotaljohn (approved):
Where a given feature is _clearly_ targeting a specific proto-persona, writing tests in those terms is a slick way of folding in more context into the codebase. ❤️

PR #543, opened 2021-11-11
Use node display name in messages 

vmunishwar:


pkg/yamltemplate/filetests/if-err-too-many-end.yml
```diff
@@ -12,4 +12,4 @@ metadata:
 ERR: 
 - unexpected end
     stdin:8 |     #@ end
-    stdin:? | __ytt_tplXXX_start_node(11) # *yamlmeta.Document (generated)
\ No newline at end of file
+    stdin:? | __ytt_tplXXX_start_node(11) # document (generated)
```
pivotaljohn: In this case, the reader is _at least_ an advanced ytt user, if not a maintainer. Also, internal details abound in this output. By my eye, it's appropriate to present the Go type, here.

vmunishwar: Thanks for the review. I have made the required changes.

pkg/schema/schema.go
```diff
@@ -252,9 +252,12 @@ func inferTypeFromValue(value interface{}, position *filepos.Position) (yamlmeta
 		return &ScalarType{ValueType: *new(bool), defaultValue: typedContent, Position: position}, nil
 	case nil:
 		return nil, nil
+	case yamlmeta.Node:
+		return nil, fmt.Errorf("Expected value '%s' to be a map, array, or scalar, but was %s", value, typedContent.DisplayName())
+	default:
+		return nil, fmt.Errorf("Expected value '%s' to be a map, array, or scalar, but was %T", value, value)
 	}
```
cari-lynn:
```suggestion
```
Style comment: newline can be removed

pkg/yamltemplate/starlark_fragment.go
```diff
@@ -17,6 +17,7 @@ const (
 	starlarkFragmentType = "yamlfragment"
 )
 
+// StarlarkFragment is an adapter for yamlmeta.Node values in Starlark code
```
cari-lynn, line 20: Great to see the godocs comment added here! It will definitely help with understanding this code.

cari-lynn (approved):
This is definitely an improvement to error messages. And it is nice to see that there are many places we will be able to use the display name.

pkg/schema/schema.go
vmunishwar: Thanks for the review. Committed the changes.

gcheadle-vmware (approved):
LGTM! Great work

PR #545, opened 2021-11-16
`@schema/desc` adds `description` key to OpenAPI Doc

cari-lynn:
When a schema node is annotated with `@schema/desc` and exported as an openapi document, the `description` key is added.


pkg/schema/annotations.go
```diff
@@ -146,6 +192,11 @@ func (n *DefaultAnnotation) NewTypeFromAnn() yamlmeta.Type {
 	return nil
 }
 
+// NewTypeFromAnn returns type information given by annotation. DescriptionAnnotation has no type information.
+func (n *DescriptionAnnotation) NewTypeFromAnn() yamlmeta.Type {
+	return nil
+}
+
```
pivotaljohn: Can we massage `Annotation` into something more broadly suitable?... 🤔 

Here's an attempt:

I think we're "discovering" three (3) types of annotations:
- `TypingAnnotation` — contributes to the schema-type of the node
- `ValueHavingAnnotation` — contributes to the value of the node
- `DocumentingAnnotation` — contributes to the description of the node/type

And we'd break up `collectionAnnotation()` into three separate (very similar) functions that gather each collection.

pkg/schema/openapi.go
```diff
@@ -74,16 +86,26 @@ func (o *OpenAPIDocument) calculateProperties(schemaVal interface{}) *yamlmeta.M
...
 	case *AnyType:
-		return &yamlmeta.Map{Items: []*yamlmeta.MapItem{
+		m := &yamlmeta.Map{Items: []*yamlmeta.MapItem{
 			{Key: "nullable", Value: true},
 			{Key: "default", Value: typedValue.GetDefaultValue()},
 		}}
+		if typedValue.GetDescription() != "" {
+			m.Items = append(m.Items, &yamlmeta.MapItem{Key: "description", Value: typedValue.GetDescription()})
+		}
+		return m
```
pivotaljohn: We're using two different names for the same variable: `properties` vs. `m`. At this point, it's a difference without a reason (at least that I can see, right now).

Let's make them all consistent.

pkg/schema/type.go
```diff
@@ -17,6 +17,7 @@ var _ yamlmeta.Type = (*ArrayType)(nil)
 var _ yamlmeta.Type = (*ArrayItemType)(nil)
 var _ yamlmeta.Type = (*AnyType)(nil)
 var _ yamlmeta.Type = (*NullType)(nil)
+var _ yamlmeta.Type = (*ScalarType)(nil)
```
pivotaljohn, line 20: Nice catch; thank you.

pkg/schema/type.go
```diff
@@ -96,8 +102,8 @@ func (a *ArrayItemType) SetDefaultValue(val interface{}) {
 }
 
 // SetDefaultValue sets the default value to `val`
-func (a *ScalarType) SetDefaultValue(val interface{}) {
-	a.defaultValue = val
+func (m *ScalarType) SetDefaultValue(val interface{}) {
+	m.defaultValue = val
```
pivotaljohn: I can see `s` or `t` ... but `m`? 😁

pkg/schema/openapi.go
```diff
@@ -52,17 +52,29 @@ func (o *OpenAPIDocument) calculateProperties(schemaVal interface{}) *yamlmeta.M
 		property := yamlmeta.Map{Items: []*yamlmeta.MapItem{
 			{Key: "type", Value: "object"},
 			{Key: "additionalProperties", Value: false},
-			{Key: "properties", Value: &yamlmeta.Map{Items: properties}},
 		}}
+
+		if typedValue.GetDescription() != "" {
+			property.Items = append(property.Items, &yamlmeta.MapItem{Key: "description", Value: typedValue.GetDescription()})
```
pivotaljohn: Given that we're using the same string in multiple places, should we extract the OpenAPI property names into constants?

pivotaljohn (approved):
Overall, LGTM. 👍🏻 

Comments are reacting to some internal structure thoughts... these can easily be addressed in the upcoming stories too.

pkg/schema/annotations.go
cari-lynn: Great suggestion, added a function for this

pkg/schema/openapi.go
```diff
@@ -9,6 +9,14 @@ import (
 	"github.com/k14s/ytt/pkg/yamlmeta"
 )
 
+// keys used when generating an OpenAPI Document
+const (
+	Type        = "type"
+	Default     = "default"
+	Nullable    = "nullable"
+	Description = "description"
```
pivotaljohn:
```suggestion
	typeProp        = "type"
	defaultProp     = "default"
	nullableProp    = "nullable"
	descriptionProp = "description"
```

Suggestions:
- un-export these constants as they are not intended to be used outside the package.
- need to disambiguate `type` from the Go keyword... for example add the `Prop` suffix. 🤷🏻

pivotaljohn (approved):
One request to un-export the constants for the OpenAPI property names, but otherwise GTG.

(feel free to merge after addressing that suggestion)

PR #546, opened 2021-11-17
bug fix for fmt percentage sign handling

vmunishwar:


pivotaljohn (approved):
This is the fix!

For another day: I suspect that we mean to write out _uninterpreted strings_ throughout this function. This bug originated because we're using the `fmt` package rather than some way to output simple strings.

Issue #310, opened 2021-02-25
duplicate key check in plain yamls


mhubig:
ytt crashes when running  against the current tekton release:

```shell
ytt --file-mark '*.yaml:type=yaml-plain' -f https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.21.0/release.yaml

panic: Unexpected duplicate key: name

goroutine 1 [running]:
github.com/k14s/ytt/pkg/yamlmeta.convertToGo(0x158b5c0, 0xc0002d5a40, 0x0, 0x0)
        github.com/k14s/ytt/pkg/yamlmeta/convert.go:65 +0x7ca
github.com/k14s/ytt/pkg/yamlmeta.convertToGo(0x158b300, 0xc0002d54a0, 0xc000197e50, 0x0)
        github.com/k14s/ytt/pkg/yamlmeta/convert.go:74 +0x454
github.com/k14s/ytt/pkg/yamlmeta.convertToGo(0x158b5c0, 0xc0002d5380, 0xc000197df0, 0x0)
        github.com/k14s/ytt/pkg/yamlmeta/convert.go:67 +0x5ed
github.com/k14s/ytt/pkg/yamlmeta.convertToGo(0x158b5c0, 0xc0002d4f00, 0x10ff7a7, 0xc00046853a)
        github.com/k14s/ytt/pkg/yamlmeta/convert.go:67 +0x5ed
github.com/k14s/ytt/pkg/yamlmeta.(*Document).AsYAMLBytes(0xc00032e8a0, 0xc00014c940, 0x4, 0x4, 0x4, 0x0)
        github.com/k14s/ytt/pkg/yamlmeta/document.go:25 +0x3c
github.com/k14s/ytt/pkg/yamlmeta.(*YAMLPrinter).Print(0xc0003ce5e0, 0xc00032e8a0, 0x0, 0x0)
        github.com/k14s/ytt/pkg/yamlmeta/printers.go:36 +0x42
github.com/k14s/ytt/pkg/yamlmeta.(*DocumentSet).AsBytesWithPrinter(0xc000116000, 0x1674e98, 0x2e, 0x30, 0xc000220800, 0x2e, 0x40)
        github.com/k14s/ytt/pkg/yamlmeta/document_set.go:58 +0xe8
github.com/k14s/ytt/pkg/yamlmeta.(*DocumentSet).AsBytes(...)
        github.com/k14s/ytt/pkg/yamlmeta/document_set.go:43
[...]

PR #550, opened 2021-11-22
(re)enable map key override for plain YAML

pivotaljohn:
Addresses #310

### DIFF
--- a/pkg/workspace/template_loader.go
+++ b/pkg/workspace/template_loader.go
@@ -159,12 +159,22 @@ func (l *TemplateLoader) EvalYAML(libraryCtx LibraryExecutionContext, file *file
 		return nil, nil, err
 	}
 
-	l.ui.Debugf("### ast\n")
-	docSet.Print(l.ui.DebugWriter())
+	l.ui.Debugf("### ast ")
 
+	// is this plain YAML?
 	if !file.IsTemplate() && !file.IsLibrary() || !yamltemplate.HasTemplating(docSet) {
+		// YAML spec requires map keys to be unique.
+		// Tools retain just the last instance: each subsequent map item overrides the value of any previous.
+		docSet.OverrideMapKeys()
+
+		l.ui.Debugf("(plain)\n")
+		docSet.Print(l.ui.DebugWriter())
+
 		return nil, docSet, nil
 	}
+
+	l.ui.Debugf("(templated)\n")
+	docSet.Print(l.ui.DebugWriter())
 
 	tplOpts := yamltemplate.TemplateOpts{
 		IgnoreUnknownComments:   l.opts.IgnoreUnknownComments,
--- a/pkg/yamlmeta/map_key_overrides.go
+++ b/pkg/yamlmeta/map_key_overrides.go
@@ -1,0 +1,32 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package yamlmeta
+
+import (
+	"github.com/k14s/ytt/pkg/orderedmap"
+)
+
+type overrideMapKeys struct{}
+
+// Visit if `node` is a Map, among its MapItem's that have duplicate keys, removes all but the last.
+// This visitor always returns `nil`
+func (r *overrideMapKeys) Visit(node Node) error {
+	mapNode, isMap := node.(*Map)
+	if !isMap {
+		return nil
+	}
+
+	lastItems := orderedmap.NewMap()
+	for _, item := range mapNode.Items {
+		lastItems.Set(item.Key, item)
+	}
+
+	var newItems []*MapItem
+	lastItems.Iterate(func(_, value interface{}) {
+		newItems = append(newItems, value.(*MapItem))
+	})
+	mapNode.Items = newItems
+
+	return nil
+}
--- a/pkg/yamlmeta/map_key_overrides_test.go
+++ b/pkg/yamlmeta/map_key_overrides_test.go
@@ -4,7 +4,6 @@
 package yamlmeta_test
 
 import (
-	"os"
 	"testing"
 
 	"github.com/k14s/ytt/pkg/yamlmeta"
@@ -25,7 +24,7 @@ func TestMapKeyOverridePlainYAML(t *testing.T) {
 		expectedDocSet := docSet.DeepCopyAsNode()
 		docSet.OverrideMapKeys()
 
-		printer := yamlmeta.NewPrinterWithOpts(os.Stdout, yamlmeta.PrinterOpts{ExcludeRefs: true})
+		printer := yamlmeta.NewPrinterWithOpts(nil, yamlmeta.PrinterOpts{ExcludeRefs: true})
 
 		result := printer.PrintStr(docSet)
 		expected := printer.PrintStr(expectedDocSet)
@@ -55,7 +54,7 @@ func TestMapKeyOverridePlainYAML(t *testing.T) {
 		}
 		docSet.OverrideMapKeys()
 
-		printer := yamlmeta.NewPrinterWithOpts(os.Stdout, yamlmeta.PrinterOpts{ExcludeRefs: true})
+		printer := yamlmeta.NewPrinterWithOpts(nil, yamlmeta.PrinterOpts{ExcludeRefs: true})
 
 		result := printer.PrintStr(docSet)
 		expected := printer.PrintStr(expectedDocSet)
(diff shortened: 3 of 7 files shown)


pkg/workspace/template_loader.go
```diff
@@ -159,13 +159,20 @@ func (l *TemplateLoader) EvalYAML(libraryCtx LibraryExecutionContext, file *file
 		return nil, nil, err
 	}
 
-	l.ui.Debugf("### ast\n")
-	docSet.Print(l.ui.DebugWriter())
+	l.ui.Debugf("### ast ")
 
 	if !file.IsTemplate() && !file.IsLibrary() || !yamltemplate.HasTemplating(docSet) {
+		docSet.OverrideMapKeys()
```
cppforlife, line 168: feels a little weird doing it here... but i guess that it makes sense since we are "evaling" yaml. should we leave a comment here?

pkg/yamlmeta/map_key_overrides.go
```diff
@@ -0,0 +1,51 @@
...
+	mapNode, isMap := node.(*Map)
+	if !isMap {
+		return nil
+	}
+
+	keyToIdxs := r.collectIndexesByKey(mapNode)
+	idxsToRemove := r.selectRedundantIdxs(keyToIdxs)
+	for _, idx := range idxsToRemove {
+		mapNode.Items = append(mapNode.Items[:idx], mapNode.Items[idx+1:]...)
+	}
```
cppforlife: how about something like this (havent actually ran):

```
lastItemIndex := map[interface{}]int{}
for idx, item := range mapNode.Items {
	lastItemIndex[item.Key] = idx
}

newItems := []*MapItem{}
seenIndex := map[interface{}]struct{}{}
for _, item := range mapNode.Items {
	idx, found := lastItemIndex[item.Key]
	if !found {
		panic("Internal inconsistency: ...")
	}
	if _, found := seenIndex[item.Key]; !found {
		seenIndex[item.Key] = struct{}{}
		newItems = append(newItems, mapNode.Items[idx])
	}
}

mapNode.Items = newItems
```

pkg/yamlmeta/map_key_overrides_test.go
```diff
@@ -0,0 +1,98 @@
...
+				},
+			}},
+		}
+		expectedDocSet := &yamlmeta.DocumentSet{
+			Items: []*yamlmeta.Document{{
+				Value: &yamlmeta.Map{
+					Items: []*yamlmeta.MapItem{
+						{Key: "ree", Value: 4},
+						{Key: "foo", Value: 5},
+						{Key: "bar", Value: 6},
```
cppforlife: ordering of keys should probably remain the same... foo, bar, ree.

pkg/yamlmeta/map_key_overrides_test.go
```diff
@@ -0,0 +1,98 @@
...
+					},
+				},
+			}},
+		}
+		docSet.OverrideMapKeys()
+
+		printer := yamlmeta.NewPrinterWithOpts(os.Stdout, yamlmeta.PrinterOpts{ExcludeRefs: true})
+
+		result := printer.PrintStr(docSet)
+		expected := printer.PrintStr(expectedDocSet)
```
cppforlife: not sure if we should use a printer for this. cant we just do direct equal?

pkg/yamlmeta/walk.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2020 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package yamlmeta
+
+// Visitor performs an operation on the given Node while traversing the AST.
+// Typically defines the action taken during a Walk().
+type Visitor interface {
+	visit(Node) error
```
cppforlife: is this private intentionally? if so, any reason why not make walk private as well?

pkg/workspace/template_loader.go
pivotaljohn: Yeah... I think part of what forces this awkwardness of the location of this logic is that we can't perform this check _until_ the YAML has been parsed. And we don't want to de-dup map keys within a template right here... 🤷🏻 

Yeah, I can contextualize this a bit... good point.

pkg/yamlmeta/map_key_overrides.go
pivotaljohn: nice. 👍🏻   Far far simpler, no sorting, etc. And summarizing it sounds a lot like what the desired behavior is: keep only the last.

pkg/yamlmeta/map_key_overrides_test.go
pivotaljohn: I went this way in part because that's how we've compared DocSets in other places... so, some consistency.  Perhaps there's something to be gained in doing the direct equal... I can take a look.

pkg/yamlmeta/walk.go
pivotaljohn: Yes, I did this intentionally. I didn't necessarily want to force the consuming programmer to export this function.

But one nice feature of an exported function name is that it's obviously the starting point in a set of functions attached to `struct`. So, I might just reverse this.

I see no causal connection between whether yamlmeta.Walk() should be exported and Visitor.visit be...  In fact, in order for this walker to be useful in other packages, it would _need_ to be exported.

cppforlife: >  In fact, in order for this walker to be useful in other packages, it would need to be exported.

yeah, that's what i was nudging towards -- nobody outside of this pkg would be able to impl Visitor if visit stays private.

pkg/yamlmeta/map_key_overrides_test.go
pivotaljohn: 👍🏻  Override is replace, not append.

pkg/yamlmeta/map_key_overrides.go
pivotaljohn: You'll notice I attempted something _even more_ conceptually simpler.

cppforlife: > And summarizing it sounds a lot like what the desired behavior is: keep only the last.

well keep only the last in the "space" of the first one -- because its effectively replacing it (vs delete+insert)

pkg/yamlmeta/map_key_overrides.go
```diff
@@ -0,0 +1,32 @@
...
+
+// Visit if `node` is a Map, among its MapItem's that have duplicate keys, removes all but the last.
+// This visitor always returns `nil`
+func (r *overrideMapKeys) Visit(node Node) error {
+	mapNode, isMap := node.(*Map)
+	if !isMap {
+		return nil
+	}
+
+	lastItems := orderedmap.NewMap()
```
cppforlife, line 20: the only concern i have is that orderedmap throws away MapItem metadata. not sure how important this is here -- it's a little weird because you might be working with DocumentSet and suddently your MapItems get rewritten in a lossy way when calling OverrideMapKeys.

pivotaljohn: But, in that case, the `value` in the `orderedmap.Map` _is_ a `yamlmeta.MapItem`... so, meta is preserved. Or am I missing something?

pkg/yamlmeta/walk.go
gcheadle-vmware: +1 this, I stole the walk.go file for the story I'm working on, and struggled to use it without making it Public.

pkg/yamlmeta/map_key_overrides.go
cppforlife: oh jeez. thats very sneaky.

pkg/yamlmeta/map_key_overrides_test.go
```diff
@@ -0,0 +1,98 @@
...
+						{Key: "foo", Value: 1},
+						{Key: "bar", Value: 2},
+					},
+				},
+			}},
+		}
+		expectedDocSet := docSet.DeepCopyAsNode()
+		docSet.OverrideMapKeys()
+
+		printer := yamlmeta.NewPrinterWithOpts(os.Stdout, yamlmeta.PrinterOpts{ExcludeRefs: true})
```
cppforlife: nit: it's a little weird to have os.Stdout. lets throw in `nil` instead.

pivotaljohn: Copy-pasta from `parser_test.go`. I'll fix more broadly in a separate commit. 👍🏻

pivotaljohn:
Approval: https://kubernetes.slack.com/archives/CH8KCCKA5/p1638898347141400?thread_ts=1637592191.462200&cid=CH8KCCKA5

PR #552, opened 2021-11-24
Add annotation positions to schema error messages

gcheadle-vmware:
- only implemented on syntax errors when creating schema

### DIFF
--- a/pkg/cmd/template/schema_author_test.go
+++ b/pkg/cmd/template/schema_author_test.go
@@ -275,8 +275,8 @@ foo: 0
 		expectedErr2 := `
 Invalid schema
 ==============
-expected @schema/type annotation to have keyword argument and value
 
+expected @schema/type annotation to have keyword argument and value
 schema.yml:
     |
   3 | #@schema/type any
@@ -309,6 +309,8 @@ Invalid schema
 @schema/nullable, and @schema/type any=True are mutually exclusive
 schema.yml:
     |
+  3 | #@schema/type any=True
+  4 | #@schema/nullable
   5 | foo: 0
     |
 
@@ -423,6 +425,7 @@ Invalid schema - @schema/default not supported on array item
 
 schema.yml:
     |
+  4 | #@schema/default "baz"
   5 | - bar
     |
 
@@ -511,10 +514,11 @@ Invalid schema
 syntax error in @schema/desc annotation
 schema.yml:
     |
+  3 | #@schema/desc
   4 | key: val
     |
 
-    = found: missing value (in @schema/desc above this item)
+    = found: missing value in @schema/desc (by schema.yml:3)
     = expected: string
 `
 
@@ -537,10 +541,11 @@ Invalid schema
 syntax error in @schema/desc annotation
 schema.yml:
     |
+  3 | #@schema/desc "two", "strings"
   4 | key: val
     |
 
-    = found: 2 values (in @schema/desc above this item)
+    = found: 2 values in @schema/desc (by schema.yml:3)
     = expected: string
 `
 
@@ -563,10 +568,11 @@ Invalid schema
 syntax error in @schema/desc annotation
 schema.yml:
     |
+  3 | #@schema/desc 1
   4 | key: val
     |
 
-    = found: Non-string value (in @schema/desc above this item)
+    = found: Non-string value in @schema/desc (by schema.yml:3)
     = expected: string
 `
 
@@ -589,10 +595,11 @@ Invalid schema
 syntax error in @schema/desc annotation
 schema.yml:
     |
+  3 | #@schema/desc key=True
   4 | key: val
     |
 
-    = found: keyword argument (in @schema/desc above this item)
+    = found: keyword argument in @schema/desc (by schema.yml:3)
     = expected: string
     = hint: this annotation only accepts one argument: a string.
 `
--- a/pkg/cmd/template/schema_author_test.go
+++ b/pkg/cmd/template/schema_author_test.go
@@ -194,10 +194,11 @@ Invalid schema
 unknown @schema/type annotation keyword argument
 schema.yml:
     |
+  3 | #@schema/type unknown_kwarg=False
   4 | foo: 0
     |
 
-    = found: unknown_kwarg
+    = found: unknown_kwarg (by schema.yml:3)
     = expected: A valid kwarg
     = hint: Supported kwargs are 'any'
 `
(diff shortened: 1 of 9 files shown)


pkg/filepos/position.go
```diff
@@ -109,3 +109,16 @@ func (p *Position) DeepCopyWithLineOffset(offset int) *Position {
 	*newPos.lineNum += offset
 	return newPos
 }
+
+// IsNextTo compares the location of one position with another.
+func (p *Position) IsNextTo(otherPostion *Position) bool {
```
pivotaljohn, line 114: 😎

pkg/schema/error.go
```diff
@@ -125,24 +133,51 @@ type schemaError struct {
...
 
 type schemaAssertionError struct {
 	error
-	position    *filepos.Position
-	description string
-	expected    string
-	found       string
-	hints       []string
+	annPositions []*filepos.Position
+	position     *filepos.Position
```
pivotaljohn, line 150: Is it possible to merge these into a single slice of `Position`s?  I dug in a bit and it looked feasible.

If we could, it would simplify a bunch of code.

pkg/template/evaluation_ctx.go
```diff
@@ -156,9 +157,15 @@ func (e *EvaluationCtx) TplStartNodeAnnotation(
...
 		return starlark.None, err
 	}
-
 	annName := AnnotationName(annNameStr)
-	annVals := args.Index(2).(starlark.Tuple)
+
+	lineNum, err := strconv.Atoi(args.Index(2).String())
+	if err != nil {
+		//what error would this be? string should be line number... when would it not
+		return starlark.None, err
```
pivotaljohn: Looking at this error case got me thinking...

It's technically possible that the line number of the annotation is fully unknown. This can happen, for example, when an AST is being built in memory. In that case, there _is no_ comment to refer to.

Perhaps an appropriate thing to do here is...

- on the calling side:
    - if the annotation's Position is known, pass it along as you are.
    - if not, pass in `None`.
- on the receiving end...
   - if the annotation's Position is known, then Bob's yer uncle (and if for some crazy reason that's not an integer... seems like panic'ing is in order).
   - if the value is `None`, then build the "source" from the annotation, itself (e.g. `NodeAnnotation.AsSourceString()`)

Happy to chat more about this... feels like something that _could_ be deferred so long as we have a simple fallback strategy for when positions are unknown.

pkg/cmd/template/schema_author_test.go
```diff
@@ -290,6 +309,8 @@ Invalid schema
 @schema/nullable, and @schema/type any=True are mutually exclusive
 schema.yml:
     |
+  3 | #@schema/type any=True
+  4 | #@schema/nullable
   5 | foo: 0
     |
```
pivotaljohn, line 315: A sweet example of when multiple annotations are part of the "problem." 👍🏻

gcheadle-vmware:
Still wanting to follow up on the error output case for multiple file names within a slice of positions. 
Also need to clean up import sorting.

pkg/schema/error.go
```diff
@@ -155,29 +154,34 @@ type schemaAssertionError struct {
 	hints        []string
 }
 
-type annErr struct {
+type posInfo struct {
 	Pos       string
 	Source    string
 	SkipLines bool
 }
```
pivotaljohn: Nice separating the calculation of the "view" from the presentation of it. 👍🏻

pkg/schema/error.go
```diff
@@ -155,29 +154,34 @@ type schemaAssertionError struct {
...
-		} else {
-			skipLines := !p.IsNextTo(annPosList[i+1])
-			AnnPositions = append(AnnPositions, annErr{Pos: p.AsIntString(), Source: p.GetLine(), SkipLines: skipLines})
+	allPositions := append(annPosList, nodePos)
+	var positionsInfo []posInfo
+	for i, p := range allPositions {
+		//last position in list
+		skipLines := false
+		if i < len(allPositions)-1 {
+			skipLines = !p.IsNextTo(allPositions[i+1])
```
pivotaljohn: nit: improvement in readability?  

```suggestion
		if i > 0 {
			skipLines = !p.IsNextTo(allPositions[i-1])
		}
```

... and swap the order of `...` and source in the template:

```diff
  {{- range .Positions}}
- {{pad "|" .Pos}} {{.Source}}
- {{- if .SkipLines}}
- {{pad "|" ""}} {{"..."}}
- {{- end}}
+ {{- if .SkipLines}}
+ {{pad "|" ""}} {{"..."}}
+ {{- end}}
+ {{pad "|" .Pos}} {{.Source}}
  {{- end}}
```

pkg/template/evaluation_ctx.go
```diff
@@ -159,10 +159,15 @@ func (e *EvaluationCtx) TplStartNodeAnnotation(
...
-	if err != nil {
-		//what error would this be? string should be line number... when would it not
-		return starlark.None, err
+	var position *filepos.Position
+	if _, ok := args.Index(2).(starlark.NoneType); ok {
+		position = filepos.NewUnknownPosition()
+	} else {
+		lineNum, err := strconv.Atoi(args.Index(2).String())
+		if err != nil {
+			panic(fmt.Sprintf("expected line num to be int or None, but was %v", args.Index(2).String()))
```
pivotaljohn: What's the `err`, here? would it be useful context to include in the panic?

pkg/template/instructions.go
```diff
@@ -58,7 +58,11 @@ func (is *InstructionSet) NewEndCtxNone() Instruction {
 
 func (is *InstructionSet) NewStartNodeAnnotation(nodeTag NodeTag, ann Annotation) Instruction {
 	collectedArgs := is.CollectNodeAnnotation.WithArgs(ann.Content).AsString()
-	return is.StartNodeAnnotation.WithArgs(nodeTag.AsString(), `"`+string(ann.Name)+`"`, ann.Position.AsIntString(), collectedArgs)
+	annLineNum := "None"
+	if ann.Position.IsKnown() {
+		annLineNum = ann.Position.AsIntString()
+	}
```
pivotaljohn: Two suggestions in one:
- defer to Starlark itself for the name of `None`: it's _that_ package's responsibility. Also "documents" this dependency on a Starlark thing via an `import`
- is the "either-or" aspect (slightly) more readily clear with the `else`?

```suggestion
	var annLineNum string
	if ann.Position.IsKnown() {
		annLineNum = ann.Position.AsIntString()
	} else {
		annLineNum = starlark.None.String()
	}
```

pivotaljohn (approved):
With these suggestions, take on what you agree with...

otherwise LGTM. 👍🏻

pkg/template/evaluation_ctx.go
gcheadle-vmware: Yeah, after looking at what the error might be here, I think it is smart to include it.

Issue #218, opened 2020-10-15
add assert.equals to @ytt:assert


cppforlife:
this would help folks write tests within ytt and not have implement equals themselves (like in cf-for-k8s: https://github.com/cloudfoundry/cf-for-k8s/blob/18b4f3f676a85e49e675120f5b720aab91ac3892/config/istio/gateway_test.star#L77-L81)

- [x] #555
- [x] document the `assert.equals()` function

gcheadle-vmware:
Released in ytt v0.39.0

PR #555, opened 2021-11-29
Add assert-equals #218

seemiller:


pkg/yamltemplate/filetests/ytt-library/assert-equals-err.yml
```diff
@@ -0,0 +1,11 @@
+#@ load("@ytt:assert", "assert")
+#@ load("@ytt:json", "json")
+
+test_equals: #@ assert.equals("foo", "bar")
+
++++
+
+ERR: 
+- assert.equals: expected foo, but found bar
```
cari-lynn: This error could use a little more description as to why we were expecting 'foo'. Maybe a message with the the words not equal in it would help. For example: 'unequal: bar not equal to foo'.

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +26,20 @@ var (
 
 type assertModule struct{}
 
+func (b assertModule) Equals(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() != 2 {
+		return starlark.None, fmt.Errorf("expected two arguments")
+	}
+
+	expected, _ := core.NewStarlarkValue(args.Index(0)).AsString()
+	actual, _ := core.NewStarlarkValue(args.Index(1)).AsString()
```
cari-lynn: Before this is merged, we probably want to check these errors and report them.

pkg/yamltemplate/filetests/ytt-library/assert-equals.yml
```diff
@@ -0,0 +1,10 @@
+#@ load("@ytt:assert", "assert")
+#@ load("@ytt:json", "json")
+
+equals_works: #@ assert.equals("foo", "foo")
+
++++
+
+equals_works:
+- null
```
cari-lynn: Thinking about the signature of this function. Currently, when the two parameters are equal it returns null. Maybe we want to return True instead, to indicate success. This is a bit different than other equals functions since if the result is ever false, it will error out instead of return False. Still, maybe returning True could be helpful here.

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +26,20 @@ var (
 
 type assertModule struct{}
 
+func (b assertModule) Equals(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() != 2 {
+		return starlark.None, fmt.Errorf("expected two arguments")
+	}
+
+	expected, _ := core.NewStarlarkValue(args.Index(0)).AsString()
```
cari-lynn: going back at the motivating use case for this feature, [here](https://github.com/cloudfoundry/cf-for-k8s/blob/18b4f3f676a85e49e675120f5b720aab91ac3892/config/istio/gateway_test.star#L77-L81), the function allows types other than strings. We should expand this function to accept other types as well.

cari-lynn (commented):
This is a great addition. The direction here is looking good.

pkg/yamltemplate/filetests/ytt-library/assert-equals.yml
cari-lynn: I put more thought into this and realized that two return types (returning a tuple) are not needed, and rather than return a tuple or True, we can just return starlark.None. 

In the success case: return starlark.None. This is essentially a no-op.

In the error case: return the error. Returning an error from a Starlark builtin function, will stop execution and report the error. So there is no need to return an error for the user to check, since execution is stopped.

Then this results in this tests now only needs to check a single `null` value is returned.

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +26,20 @@ var (
...
+	actual, _ := core.NewStarlarkValue(args.Index(1)).AsString()
+	if expected != actual {
+		return starlark.None, fmt.Errorf("expected %s, but found %s", expected, actual)
+	}
+
+	return starlark.Tuple{starlark.None, starlark.None}, nil
```
cari-lynn: As a follow on to [this](https://github.com/vmware-tanzu/carvel-ytt/pull/555/files#r758812486) comment, this can return `starlark.None, nil`, instead of a tuple.

pkg/yamltemplate/filetests/ytt-library/assert-equals-err.yml
seemiller: changed to: `unequal: bar is not equal to the expected value foo`

pkg/yttlibrary/assert.go
seemiller: I removed the conversion to String and it removes the need for a return value here. This might also further the "motivating" case for this feature that you commented on below.

pkg/yttlibrary/assert.go
seemiller: I added a test for numbers. What other types could be passed in as arguments?

cari-lynn: It could get complex if we allowed all values such as maps, arrays, documents, etc. We do want to support comparing these types, however we would need to implement an algorithm to compare these nested structures which may be complex. So, to keep the scope of this story small I think we should only support scalar types for now. This is similar functionality to what currently exists with the function linked above. So that would be 
``` 
string, float64, int, int64, bool, nil
```
This can be accomplished by having a function that checks if the expected value is one of those types, and errors if not.

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +26,27 @@ var (
 
 type assertModule struct{}
 
+func (b assertModule) Equals(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() != 2 {
+		return starlark.None, fmt.Errorf("expected two arguments")
+	}
+
+	expected, err := core.NewStarlarkValue(args.Index(0)).AsGoValue()
```
cari-lynn: Since we don't want to allow lambda functions as arguments, we will need to explicitly check for them prior to trying to convert them to a go value (otherwise we get an 'error converting function to go value'). We could do a check similar to 

https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/yttlibrary/assert.go#L46-L49

But instead error if we get a `starlark.Callable` type.

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +26,27 @@ var (
...
+	if err != nil {
+		return starlark.None, err
+	}
+	actual, err := core.NewStarlarkValue(args.Index(1)).AsGoValue()
+	if err != nil {
+		return starlark.None, err
+	}
+
+	if expected != actual {
+		return starlark.None, fmt.Errorf("unequal: %s is not equal to the expected value %s", actual, expected)
```
cari-lynn: For `actual` and `expected` here, since the formatting for `int64`, `float64`, etc do not have easily readable string representations and we do not yet have a helper function available for this, we can add a little converter. Something like this that I whipped up below:

```golang
func scalarToString(val interface{}) string {
	switch typedVal := val.(type) {
	case float64:
		return strconv.FormatFloat(typedVal, 'f', -1, 64)
	case int64:
		return strconv.FormatInt(typedVal, 10)
	case bool:
		return strconv.FormatBool(typedVal)
	case nil:
		return "nil"
	}
	return fmt.Sprintf("%s", val)
}
```

pkg/yamltemplate/filetests/ytt-library/assert-equals.yml
```diff
@@ -0,0 +1,14 @@
+#@ load("@ytt:assert", "assert")
+#@ load("@ytt:json", "json")
```
cari-lynn: The unused json imports can be removed from these tests

pkg/yamltemplate/filetests/ytt-library/assert-equals-err-decimal.yml
```diff
@@ -0,0 +1,11 @@
+#@ load("@ytt:assert", "assert")
+#@ load("@ytt:json", "json")
+
+test_equals: #@ assert.equals(3.14159, 2.71828)
```
cari-lynn: Love these additional tests! Now that there are several files what do you think about placing all these assert tests into an assert directory?

pkg/yttlibrary/assert.go
cppforlife: > So, to keep the scope of this story small I think we should only support scalar types for now. 

unless we are planning to immediately follow up with a PR, i wouldnt mind seeing maps, arrays, yamlfragments etc in this PR.

cari-lynn: There was a conversation about this in [Slack ](https://kubernetes.slack.com/archives/CH8KCCKA5/p1638896293138700). We talked about encoding the value in yaml, which then could be compared and displayed in the error message.

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +27,30 @@ var (
...
 
+func (b assertModule) Equals(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() != 2 {
+		return starlark.None, fmt.Errorf("expected two arguments")
+	}
+
+	expected := args.Index(0)
+	actual := args.Index(1)
+	if _, notOk := expected.(starlark.Callable); notOk {
+		return starlark.None, fmt.Errorf("lambdas are not allowed")
```
cari-lynn: Elsewhere in this file we use the word 'function' instead of 'lambda', so  similar wording like: "Expected argument not to be a function" would be more consistent

> 		return starlark.None, fmt.Errorf("expected argument to be a function, but was %T", lambda)

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +27,30 @@ var (
...
 type assertModule struct{}
 
+func (b assertModule) Equals(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
+	if args.Len() != 2 {
+		return starlark.None, fmt.Errorf("expected two arguments")
+	}
+
+	expected := args.Index(0)
+	actual := args.Index(1)
+	if _, notOk := expected.(starlark.Callable); notOk {
```
cari-lynn, line 36: Let also check that the second argument is not a `starlark.Callable` as well

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +27,30 @@ var (
...
+		return starlark.None, fmt.Errorf("lambdas are not allowed")
+	}
+
+	expectedType := expected.Type()
+	actualType := actual.Type()
+	if expectedType != actualType {
+		return starlark.None, fmt.Errorf("arguments are of different types. expected %s, but got %s", expectedType, actualType)
+	}
+
+	if !reflect.DeepEqual(expected, actual) {
```
cari-lynn: This was covered elsewhere, reiterating here: `reflect.DeepEqual()` works for scalars, maps, arrays well, but it does not work for [`yamlFragments`](https://carvel.dev/ytt/docs/latest/lang-ref-yaml-fragment/#docs), so in order to support that use case as well we can encode the value in yaml, and compare them as strings that way.

pkg/yamltemplate/filetests/ytt-library/assert/assert-equals.yml
```diff
@@ -0,0 +1,18 @@
+#@ load("@ytt:assert", "assert")
+
+equals_works_strings: #@ assert.equals("foo", "foo")
+equals_works_numbers: #@ assert.equals(1, 1)
+equals_works_deciaml: #@ assert.equals(3.14159, 3.14159)
+equals_works_boolean: #@ assert.equals(True, True)
+equals_works_none: #@ assert.equals(None, None)
+equals_works_array: #@ assert.equals([1, 2], [1, 2])
+equals_works_map: #@ assert.equals({"foo": "bar"}, {"foo": "bar"})
++++
```
cari-lynn, line 35: 🎉  for adding maps and arrays!

Another test should be added for [`yamlFragments`](https://carvel.dev/ytt/docs/latest/lang-ref-yaml-fragment/#docs) both for success and failure.

seemiller: Got a test working for the yamlfragment failure case  - The TestYAMLTemplate function is _extremely_ picky about whitespace and it was exceedingly frustrating to get a test to pass.

This is what the error message looks like (or at least gets the test to pass)
```
- assert.equals: yamlfragments are not equal:
    in <toplevel>
      stdin:12 | test_equals: #@ assert.equals(yamlFunc(), moreYamlFunc())

    reason:
     Expected:
     foo: bar
     -----
     Actual:
     bar: foo
```

pkg/yttlibrary/assert.go
seemiller: Added checks for maps, arrays and yamlfragments.

seemiller:
I can rebase into one commit after any reviews or further changes are complete.

pkg/yamltemplate/filetests/ytt-library/assert/assert-equals.yml
cari-lynn: This is a good format. For consistency in this codebase let's follow the testing library we use elsewhere. They present the information in the format below. This does not have capitalization, no leading whitespace, and a newline to separate the two blocks:

https://github.com/stretchr/testify/blob/ab6dc3262822ed562480c19876b0257ace761e3e/assert/assertions.go#L346-L348

To introduce as few new concepts in the error message as possible, let's leave out the word `yamlFragment`.

cari-lynn (changes requested):
If the second argument is `yamlFragment` but the first is not, we get this error. We want to be able to present the `yamlFragment` as a string.

```
ytt: Error: 
- assert.equals: yamlfragment(*yamlmeta.Map) is not equal to the expected value {"map": "value"}
    in <toplevel>
      frag.yml:8 | test_equals: #@ assert.equals({"map":"value"}, yamlFunc())
```

pkg/yamltemplate/template_test.go
```diff
@@ -121,9 +121,9 @@ func TestYAMLTemplate(t *testing.T) {
 		t.Errorf("%s", errs[0].Error())
 	}
 
-	if len(selectedFileTestPath) > 0 {
-		t.Errorf("skipped tests")
-	}
+	//if len(selectedFileTestPath) > 0 {
+	//	t.Errorf("skipped tests")
+	//}
```
cari-lynn: This can be uncommented

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +28,69 @@ var (
...
+		}
+		actualYamlString = string(actualYaml)
+	}
+
+	if expected.Type() == "yamlfragment" && actual.Type() != "yamlfragment" {
+		return starlark.None, fmt.Errorf("Not equal:\n"+
+			"expected:\n%s\nactual:\n%s", expectedYamlString, actual.String())
+	} else if expected.Type() != "yamlfragment" && actual.Type() == "yamlfragment" {
+		return starlark.None, fmt.Errorf("Not equal:\n"+
+			"expected:\n%s\n\nactual:\n%s", expected.String(), actualYamlString)
```
cari-lynn: Nice work handing all of these cases. 

I think it would simplify the number of if statements here if these error cases were instead done in their respective if statements above. For example, in the if statement for: 
[`if expected.Type() == "yamlfragment" {`](https://github.com/vmware-tanzu/carvel-ytt/pull/555/files#diff-d07fbb786694723298677950720360973e65ec6c38222b5f708e839ef53281bbR48)

A check can be added for ` if actual.Type() != "yamlfragment"`, and then we could report this error, instead of doing it here. I Think that would make it easier to follow all of the possible outcomes of the function.

cari-lynn (changes requested):
Great work on this, I am really happy with the work you've done and the changes you've applied. It's been great working with you. As I have discussed with you elsewhere, I am going to be offline for some time so I am requesting @pivotaljohn  's help to review and approve this PR moving forward.

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +28,69 @@ var (
...
+		expectedStarlarkValue, err := core.NewStarlarkValue(expected).AsGoValue()
+		if err != nil {
+			return starlark.None, err
+		}
+		document := yamlmeta.Document{Value: expectedStarlarkValue}
+		expectedYaml, err := document.AsYAMLBytes()
+		if err != nil {
+			return starlark.None, err
+		}
+		expectedYamlString = string(expectedYaml)
```
pivotaljohn: This approach assumes that the contained values is not — itself — a YAML Fragment wrapping a DocSet (which is a legit possibility).

Turns out, there exists an operation in the system for converting from a YAML fragment to a string: [yamlModule.Encode()](https://github.com/vmware-tanzu/carvel-ytt/blob/fdbc6a0859e7a80d8aeea8e2e892c8b8c16fe57f/pkg/yttlibrary/yaml.go#L31). But what our code right here wants is the ability to invoke that exact functionality through a Go interface rather than a Starlark one — which doesn't currently exist.

But what if we extracted that functionality at the interface we _wished_ we had?

I'm thinking of something like this:
```go
// starlarkEncode adapts a call from Starlark to yamlModule.Encode()
func (b yamlModule) starlarkEncode(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
	if args.Len() != 1 {
		return starlark.None, fmt.Errorf("expected exactly one argument")
	}

	val, err := core.NewStarlarkValue(args.Index(0)).AsGoValue()
	if err != nil {
		return starlark.None, err
	}

	encoded, err := b.Encode(val)
	if err != nil {
		return starlark.None, nil
	}

	return starlark.String(encoded), nil
}

// Encode renders the provided input as a YAML-formatted string
func (b yamlModule) Encode(goValue interface{}) (string, error) {
	var docSet *yamlmeta.DocumentSet

	switch typedVal := goValue.(type) {
	case *yamlmeta.DocumentSet:
		docSet = typedVal
	case *yamlmeta.Document:
		// Documents should be part of DocumentSet by the time it makes it here
		panic("Unexpected document")
	default:
		docSet = &yamlmeta.DocumentSet{Items: []*yamlmeta.Document{{Value: typedVal}}}
	}

	valBs, err := docSet.AsBytes()
	if err != nil {
		return "", err
	}

	return string(valBs), nil
}
```

At _that_ point, we could simply invoke the _new_ yamlModule.Encode() and guarantee that our conversion here matches perfectly that supplied through `yaml.encode()` in the `ytt` Starlark world.

This would not only simplify this function, it would **reduce the possibility of bugs** 🐛  overall because we've reduced the total number of codepaths in the system. 👍🏻

pkg/yttlibrary/assert.go
```diff
@@ -25,6 +28,69 @@ var (
...
+			"expected:\n%s\n\nactual:\n%s", expected.String(), actualYamlString)
+	} else if expected.Type() == "yamlfragment" && actual.Type() == "yamlfragment" {
+		if expectedYamlString != actualYamlString {
+			return starlark.None, fmt.Errorf("Not equal:\n"+
+				"expected:\n%s\nactual:\n%s", expectedYamlString, actualYamlString)
+		}
+	} else {
+		if !reflect.DeepEqual(expected, actual) {
+			return starlark.None, fmt.Errorf("%s is not equal to the expected value %s", actual.String(), expected.String())
+		}
```
pivotaljohn: This error message format is different than the cases above. Is the difference making a difference?

If so, good stuff.  If not, let's make them the same.

When there are options to simplify (without going so far as to sacrifice other qualities), it's another fine way to avoid bugs.

pkg/yamltemplate/filetests/ytt-library/assert/assert-equals-err-args.yml
```diff
@@ -0,0 +1,11 @@
+#@ load("@ytt:assert", "assert")
+#@ load("@ytt:json", "json")
```
pivotaljohn:
```suggestion
```

This module is not used in this test.

Let's check the other test files as well to make sure they don't include unused code.

pivotaljohn (changes requested):
This is looking really good. It's been a bit of a windy journey to get to something this simple and useful. 👍🏻 

With a solid set of tests backing us up, we can enjoy the ability to consider refactoring confidently.

This review aims at making suggestions with the intent to simplify the implementation. I find the act of "simplifying" can often require a larger understanding of the context and codebase. My goal is to bring additional perspective that can make such simplifications possible.

I'm delighted to shoulder up with you to work through any of this thinking.

pkg/yamltemplate/template_test.go
seemiller: oops!

pivotaljohn (approved):
Paired w/ @seemiller.  👍🏻

PR #562, opened 2021-12-06
Error when schema any type node has nested schema annotations

gcheadle-vmware:


pkg/cmd/template/schema_author_test.go
```diff
@@ -288,11 +289,46 @@ schema.yml:
...
+Invalid schema
+==============
+
+@schema/... are not allowed within a fragment annotated '@schema/type any=True'
+schema.yml:
+    |
+  5 | #@schema/nullable
+  6 | #@schema/type any=False
+  7 | #@schema/default 1
+  8 |   bar: 7
```
pivotaljohn: mmmm.  Looks like my advice to consider reconstructing a line from a stringer of an annotation led to a not great place. I totally missed this case in 

The liability here is that to the extent that the portion of our error message that portends to quote the input source ... to the extent that it varies with the _actual_ source reduces the user's trust that the tool is faithfully handling their inputs.

Consider what this looks like when the data values are 3 or 4 levels indented:

```
  5 | #@schema/nullable
  6 | #@schema/type any=False
  7 | #@schema/default 1
  8 |         bar: 7
```

If you wanted to tackle this in _this_ PR, that would seem legit given that the "defect" came to light from the test we needed to write to get this feature in place. But if you wanted to tackle it in a separate PR, that's fine by me as well.

It's my gut read this is _just_ below the bar for something we'd want to release... but let's talk.

pkg/cmd/template/schema_author_test.go
```diff
@@ -288,11 +289,46 @@ schema.yml:
...
+  #@schema/default 1
+  bar: 7
+`
 
-		assertFails(t, filesToProcess, expectedErr2, opts)
+			expectedErr := `
+Invalid schema
+==============
+
+@schema/... are not allowed within a fragment annotated '@schema/type any=True'
```
pivotaljohn: I'm seeing a potential opportunity to interpret what "@schema/type any=True" means, in regular english.

Either in this title/summary, here... down in the "Expects", or a hint?

I've got the vague sense of not getting enough of the "Why" from the info in this error message.

pkg/schema/schema.go
```diff
@@ -262,6 +262,13 @@ func inferTypeFromValue(value interface{}, position *filepos.Position) (yamlmeta
 func valueTypeAllowsItemValue(explicitType yamlmeta.Type, itemValue interface{}, position *filepos.Position) error {
 	switch explicitType.(type) {
 	case *AnyType:
+		if node, ok := itemValue.(yamlmeta.Node); ok {
+			// search children for annotations
+			err := yamlmeta.Walk(node, checkForAnnotations{})
```
pivotaljohn, line 267: 💪🏻

pkg/cmd/template/schema_author_test.go
gcheadle-vmware: yep, I 100% agree. I noticed how bad it could get during some manual testing and the fix is currently in progress 👍

pkg/cmd/template/schema_author_test.go
gcheadle-vmware: This error message could defiantly be improved, let's sync up to chat about that.

pivotaljohn: > This error message could **defiantly** be improved _(emphasis added)_ ...

I can't tell if this was spellcheck or mistype... either way, I dig your moxie. 😉

pkg/cmd/template/schema_author_test.go
pivotaljohn: Exploratory testing. It's just so good. So good. 👍🏻

pivotaljohn (approved):
Thanks for jumping on the indentation part, @gcheadle-vmware.

It's clear from this PR and #552 that carrying source data all the way through the system has been quite challenging. Thank you for making this significant improvement in the UX, here. 🙏🏻 

...

I'm also feeling a creative tension to find a way to _rendez-vous_ with original inputs (rather attempt to carry them through).

For example:
- what if `schema.schemaError` and `schema.assertionFailure` were exported, and,
- `LibraryExecution.collectSchemaDocs()`, knowing what errors could be returned from `schema.NewDocumentSchemaEnvelope()`, would range over the `AssertionFailure`s, collect the `filepos.Position`
- `workspace.Library` (and crew) that could now take in one or more `filepos.Position` and return the corresponding source lines.
- that would be used to populate the source lines in the `AssertionFailure`s...

🤔 

This could be a "refactor" that could relieve `filepos.Position` of the responsibility of carrying source lines through library execution. I haven't tried it myself, so not sure if there's some 👿 in the details, here.

...

Back to _this_ PR, we've got the behavior we want (and it's pretty sweet!).

LGTM! 👍🏻

PR #564, opened 2021-12-10
add hint messages to use boolean and/or

vmunishwar:


pivotaljohn (changes requested):
This is a good first cut. 👍🏻 

I particularly appreciate you're "going for it" in the "... want primary expression" case. I believe that _most_ users will be carrying over the `&&` operator in the cases we're targeting. 🙏🏻 

There are two areas I think we need to evolve a bit more: test coverage and types of operands.

### Tests

I think there's enough risk here that warrants some tests:
- the data we're examining come from vendor'ed code (i.e. starlark-go). If/when we upgrade that dependency (e.g. for a CVE?) we'll want to know that it doesn't create a regression.
- certain cases (i.e. the "got X, want primary expression") appropriately have additional logic that raises the confidence we won't give a false positive (and mislead the user). While we can't cover all possible inputs, it's important to show some of our work in the form of a set of tests.

Concretely, I'm recommending we:
- [x] augment `error_hints_test.go` with the use-cases we're covering here.
- [x] add a second test suite in `error_hints_test.go` that specifically verifies that we do _not_ hint in situations that would be misleading (personally, I'm thinking covering the "got X, want primary expression" cases are the one's "at risk" here.)


### Scope: covering all operand types

The "unknown binary op: ..." messages are the perfect place to start with this feature. However, right now, we're covering one of many possible cases: where both operands are booleans.

In Starlark, the operands for the disjunction (i.e. `or`) and conjunction (i.e. `and`) operators need not be booleans ([ref](https://github.com/google/starlark-go/blob/master/doc/spec.md#or-and-and)). This characteristic combined with the fact that empty values are "falsey" are handy for short-circuiting style expressions.

For example:
```python
x = []
x and x[0]
```

where the second line guards against an "index out of range" error: `[]` is an empty list and thus is falsey, so the right-hand operand (`x[0]`) is never even evaluated.

It's conceivable a user would want to employ such idiom. Providing the hint here is just as valuable as it is when the operands are booleans.

I recommend we:
- [ ] convert to a "[No Condition" style `switch`](https://yourbasic.org/golang/switch-statement/#no-condition) and using [regexp.MatchString()](https://pkg.go.dev/regexp#MatchString) for this use-case.
- [ ] use a regular expression that matches just on these error messages but ignore the types of the operands:
    ```regex
    `unknown binary op: .+ | .+`g
    ```
    (playground: https://regex101.com/r/6aD8Sj/1)

Please feel free to reach out to me to discuss!

vmunishwar:
Thank you for the review comments. I have added the test scenarios as mentioned and committed the changes. Regarding the 'Scope: covering all operand types', as discussed, we will consider this later depending on the need to provide hint message for various operand types.

pkg/template/compiled_template_error.go
```diff
@@ -204,6 +204,26 @@ func (CompiledTemplateMultiError) hintMsg(errMsg string) string {
 		hintMsg = "use 'None' instead of 'none' to indicate no value"
 	case "unhandled index operation struct[string]":
 		hintMsg = "use getattr(...) to access struct field programmatically"
+	case "unknown binary op: bool | bool":
+		hintMsg = "use 'or' instead of '|' to combine boolean expressions"
+	case "unknown binary op: bool & bool":
+		hintMsg = "use 'and' instead of '&' to combine boolean expressions"
```
pivotaljohn, line 220: 👍🏻  All the hints you're adding here are written in the same voice (with the same typography) as the existing text.

pkg/template/compiled_template_error.go
```diff
@@ -187,7 +187,7 @@ func (CompiledTemplateMultiError) findClosestLine(ct *CompiledTemplate, posLine
 	}
 }
 
-func (CompiledTemplateMultiError) hintMsg(errMsg string) string {
+func (CompiledTemplateMultiError) hintMsg(errMsg string, err CompiledTemplateError) string {
 	hintMsg := ""
 	switch errMsg {
```
pivotaljohn: 👍🏻  Yup, we absolutely need to bring in additional context to ensure we're not giving hints in the wrong circumstances.

💡 At this point, looks like we can "just" take the error, itself, and read the message off of that:

```suggestion
func (CompiledTemplateMultiError) hintMsg(err CompiledTemplateError) string {
	hintMsg := ""
	switch err.Msg {
```

pkg/template/compiled_template_error.go
```diff
@@ -204,6 +204,26 @@ func (CompiledTemplateMultiError) hintMsg(errMsg string) string {
...
+	case "unknown binary op: bool & bool":
+		hintMsg = "use 'and' instead of '&' to combine boolean expressions"
+	case "got '&', want primary expression":
+		for _, pos := range err.Positions {
+			codeLine := pos.TemplateLine.Instruction.code
+			if strings.Contains(codeLine, "&&") {
+				hintMsg = "use 'and' instead of '&&' to combine boolean expressions"
+				break
+			}
+		}
```
pivotaljohn: 💡 If the body of this case and the subsequent one are genuinely the same logic (and not just incidentally similar), let's factor that into a helper function. This has the double effect of making _this_ function more readable _and_ reducing the total number of codepaths (which generally reduces the risk of bugs).

Just like `hintMsg()`, this new function it should be a method on `CompiledTemplateMultiError` so as to namespace it (and not be defining this as a package-level function).

pkg/template/error_hints_test.go
```diff
@@ -63,8 +63,72 @@ end`,
...
+		}
+
+		if err.Error() != cs.ErrMsg {
+			t.Fatalf("Expected error to match but did not; expected >>>%s<<< vs actual >>>%s<<<", cs.ErrMsg, err)
+		}
 	}
 
+}
+
+func TestNoHints(t *testing.T) {
```
pivotaljohn, line 123: 👍🏻  So-called "negative tests" are particularly useful here: giving use evidence that hints don't appear in contexts where they would be confusing and/or lead the user in the wrong direction.

pivotaljohn (changes requested):
This is solid.

Here are some "nips and tucks" designed to improve the maintainability of this code.

pkg/template/compiled_template_error.go
vmunishwar: Thanks for pointing out this change. I have updated the code.

pkg/template/compiled_template_error.go
vmunishwar: Got it. Made the appropriate changes. Thank you.

pkg/template/compiled_template_error.go
```diff
@@ -187,9 +187,20 @@ func (CompiledTemplateMultiError) findClosestLine(ct *CompiledTemplate, posLine
 	}
 }
 
-func (CompiledTemplateMultiError) hintMsg(errMsg string) string {
+// hintMsgHelper checks if the operator is present in the code line to set the hint message appropriately
+func (CompiledTemplateMultiError) hintMsgHelper(err CompiledTemplateError, op string) bool {
```
pivotaljohn: 👍🏻   The shape of this [extract method](https://refactoring.guru/extract-method) refactor sliced through to capture the lower-level details into here.

💡  for function/method names, we aim to reveal intention. The grammar that does this is almost always a verb or verb phrase.  One approach is to take the Godoc and come up with a succinct phrase that says the same thing (and for unexported functions, it can often mean you don't need to Godoc comment any more!)

pivotaljohn (approved):
With just one comment, the rest LGTM!

PR #567, opened 2021-12-19
Improve usability/maintainability of template tests

pivotaljohn:
To date, template tests (i.e. those test run via `TestYAMLTemplate`) can be somewhat awkward to author (especially for those contributing for the first time):
- expected error output can include hard-to-detect trailing whitespace (and some code editors trim trailing whitespace by default)
- many modern editors syntax highlight/lint files with `.yaml`/`.yml` extensions as YAML, creating visual noise.
- the suite has a custom test selection mechanism; if each test were a `go test` sub-test, then the standard test selection mechanism (i.e. `-run`) can be used.


gcheadle-vmware (approved):
This refactor simplifies the testing process, and documents important info about these test files. I remember having a hard time making sense of this code a while back, and I think your changes will improve future onboarding. LGTM 👍

hack/test-all.sh
```diff
@@ -10,4 +10,15 @@ fi
 
 go test ./... "$@"
 
+# error out if -run is given but no test is run
+if [[ "$@" == *"-run "* ]]; then
+  num_pkgs_with_tests=$( go test ./... "$@"  | grep "^\(ok  \|FAIL\)\tgithub.com" | grep -v "no test" | wc -l )
```
cari-lynn, line 15: 🥳

pkg/yamltemplate/template_test.go
```diff
@@ -24,12 +24,22 @@ import (
 
 var (
 	// Example usage:
-	//   ./hack/test.sh -run TestYAMLTemplate TestYAMLTemplate.filetest=yield-def.yml
-	selectedFileTestPath = kvArg("TestYAMLTemplate.filetest")
-	showTemplateCode     = kvArg("TestYAMLTemplate.code") // eg t|...
-	showErrs             = kvArg("TestYAMLTemplate.errs") // eg t|...
+	//   ./hack/test-all.sh -v -run TestYAMLTemplate/filetest/if.tpltest
+	//   ./hack/test-all.sh -v -run TestYAMLTemplate/filetest/if.tpltest TestYAMLTemplate.code=true
```
cari-lynn: Optional: since the path here works for `../filetest/..` _and_ `../filetests/..` we can change this to `filetests` to have it match the directory structure exactly and reduce confusion in the slight naming difference. 

```suggestion
	//   ./hack/test-all.sh -v -run TestYAMLTemplate/filetests/if.tpltest
	//   ./hack/test-all.sh -v -run TestYAMLTemplate/filetests/if.tpltest TestYAMLTemplate.code=true
```

cari-lynn (approved):
Thanks for adding so much documentation with this!

pkg/yamltemplate/template_test.go
pivotaljohn: Nice catch; thank you!

Fixed.

seemiller (approved):
Appreciate the refactor here, especially WRT to ignoring whitespace - that will save quite a bit of time and greatly ease creating/modifying tests in the future.

Changing the file extension away from `yml` is also a huge win!

Issue #418, opened 2021-06-04
Set the Data Values on a library a la `--data-values-file`


pivotaljohn:
**Describe the problem/challenge you have**
I want to be able to set the data values of a library with the same semantics as the `--data-values-file` flag from the command-line.

Specifically, I want the array values I pass in to _replace_ (not append) to the array values in the data values defined in the library.

**Describe the solution you'd like**
A function on the `@ytt:library` instance that accepts a `struct`, `dict`, `list`, or scalar that overlays those values in the same way they do with `--data-values-file`:

```yaml
--- #@ template.replace(library.get("libby").with_data_values_literals({"clients": [1,2,3]}).eval())
```

**Anything else you would like to add:**

To be explicit: there's significant value in being able to regard a `library.get()...eval()` as analogous to a separate invocation of `ytt`. One can envision "refactoring" what are literally multiple `ytt` invocations into a single invocation that folds together multiple `library.get()...eval()` calls. This remains a solution so long as the two APIs (i.e. `ytt`'s command-line and the `@ytt:library` are congruent).

_This_ request is in reaction to the introduction of the `--data-values-file` flag on the CLI (#381). It was further inspired by #563.

To Do:
- [x] #568
- [x] vmware-tanzu/carvel#323

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

cppforlife:
some api alternatives:
- version library.get method (e.g. library.get_v2)
- version library module (e.g. @ytt:library2)
- add keyword argument to with_data_values(..., plain=True)

pivotaljohn:
> some api alternatives:

Yes... better decisions through a greater diversity of options.

I had a visceral reaction to versioning. :) That's some serious interface complexity... would need to be worth it.

I'm fond of the additive `plain` keyword on _this_ version of `@ytt:library`. That said, I see how sweet it would be if `@ytt:library` mirrored the CLI. 🤔

pivotaljohn:
Based on results of #429, recommending as a priority ==> recommended

pivotaljohn:
@cppforlife: what about altering the behavior depending on the presence of schema?

[...]

PR #568, opened 2021-12-19
Allow plain merge of data values into a library

pivotaljohn:
- Add keyword argument "plain" to @ytt:library.with_data_values()
- Extract/reuse annotating method from template.DataValuesFile

Addresses #418

### DIFF
--- a/pkg/cmd/template/data_values_file.go
+++ b/pkg/cmd/template/data_values_file.go
@@ -4,13 +4,8 @@
 package template
 
 import (
-	"fmt"
-
-	"github.com/k14s/starlark-go/starlark"
-	"github.com/k14s/ytt/pkg/template"
 	"github.com/k14s/ytt/pkg/yamlmeta"
-	"github.com/k14s/ytt/pkg/yamltemplate"
-	yttoverlay "github.com/k14s/ytt/pkg/yttlibrary/overlay"
+	"github.com/k14s/ytt/pkg/yttlibrary/overlay"
 )
 
 type DataValuesFile struct {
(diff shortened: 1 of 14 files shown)


pkg/workspace/library_module.go
```diff
@@ -159,16 +160,33 @@ func (l *libraryValue) WithDataValues(thread *starlark.Thread, f *starlark.Built
...
+		return starlark.None, err
+	}
 
 	dataValues, err := core.NewStarlarkValue(args.Index(0)).AsGoValue()
 	if err != nil {
 		return starlark.None, err
 	}
 
-	valsYAML, err := datavalues.NewEnvelope(&yamlmeta.Document{
+	usePlainMerge, err := core.BoolArg(kwargs, "plain")
```
cppforlife, line 176: err check

pivotaljohn: ty.

pivotaljohn:
~~Pulling back to draft because I see an inconsistency:~~

~~To mirror the merge order of the root library, we need to hold separate the _plain_ data values from data value overlays so that the former can be applied after the latter.~~

~~That is, values from `.with_data_values(plain=True)` should be applied _last_ just like values from `--data-values-file` do.~~

_(edit: this is wrong thinking; cleared up below)_

pivotaljohn:
~~So, this gets even more interesting...~~

~~If we ensure that `.with_data_values(plain=True)` has the identical precedence that `--data-values-file` has, then we get weirdness like this:~~

```yaml
#@ lib = library.get("lib")
#@ lib = lib.with_data_values({"from": "plain yaml"}, plain=True)
#@ lib = lib.with_data_values({"from": "overlay"})
--- #@ template.replace(lib.eval())
```

~~would necessarily get you:~~

```yaml
---
from: plain yaml
```

~~Which feels really surprising from the template author's perspective: how is it that the subsequent `.with_data_values()` call is seemingly ignored?!?!~~

...

~~I think we can have both: a mirror of data values precedence between the root and private libraries _and_ keep an intuitive API for template authors...~~
~~if we — instead of adding the `plain` kwarg — introduce a new method: `.set_data_values()`.~~

~~The above would _then_ look like this:~~

```yaml
#@ lib = library.get("lib")
#@ lib = lib.set_data_values({"from": "plain yaml"})
#@ lib = lib.with_data_values({"from": "overlay"})
--- #@ template.replace(lib.eval())
```

_(edit: this is not necessarily; cleared up, below.)_

gcheadle-vmware:
After taking a look through these changes, I like the way you extracted and reused current code.

In response to the most recent comment, I'm having trouble trying to motivate why we need the identical precedence that `--data-values-file` has. When trying to approach these examples from an author's perspective, I would expect that when I programmatically set data values on a library, it would apply in the order specified. Thus I would expect 
```
---
from: overlay
```
for both examples, because as an author, that how I wrote it.

pivotaljohn:
> In response to the most recent comment, I'm having trouble trying to motivate why we need the identical precedence that --data-values-file has. 

Good question. The initial motivation was to try to keep the command-line and programmatic interfaces as similar (if not identical) as possible.

I have just realized a couple of mistakes in my thinking:
- I failed to see `.with_data_values()` as equivalent to all of the command-line flags (i.e. there are many command-line flags because _that_ interface has multiple sources; the programmatic interface really has one: some AST) 💡 
- I misread the order in which the command-line flags were being processed; I got it exactly opposite — I thought the plain DVs were applied last; in fact, they are applied _first_. 🤦🏻‍♂️ 

I'm gonna pull this back into draft and fix it.

pkg/cmd/template/data_values_file.go
```diff
@@ -24,43 +21,10 @@ func NewDataValuesFile(doc *yamlmeta.Document) DataValuesFile {
 func (f DataValuesFile) AsOverlay() (*yamlmeta.Document, error) {
 	doc := f.doc.DeepCopy()
 
-	if yamltemplate.HasTemplating(doc) {
-		return nil, fmt.Errorf("Expected to not find annotations inside data values file " +
-			"(hint: remove comments starting with '#@')")
+	err := overlay.AnnotateForPlainMerge(doc)
+	if err != nil {
+		return nil, fmt.Errorf(err.Error())
```
cppforlife: return err

cppforlife:
lgtm

pivotaljohn:
Note, conversations around this work inspired #575. More to come? 😁

PR #569, opened 2021-12-19
Detect binary use of | and & for any type operands

pivotaljohn:
Follow-up from #564.

Extends hints for cases where authors use `|`/`&` for logical disjunction/conjunction for non-boolean operands (e.g. for defaulting or guard expressions)


vmunishwar (approved):

vmunishwar:
Looks good to me, this will provide hint message for any type operands used with a binary operator.

PR #570, opened 2021-12-22
Reduce accidental complexity in handling of types and type names (post schema refactor)

pivotaljohn:
A series of refactors (each commit with a separate intention) that aims to reduce complexity around schema code.


gcheadle-vmware (approved):
I looked through each commit and I think this is a quality refactor that builds on the work done previously in #519. The change set looks good to me and I think the added comments are enough of an improvement to approve these changes. Some other notes that I took while going through the changes are included below:
This refactor...
- converges on the way we print the names of types 
- places ScalarType only in the schema package, since it doesn't make a ton of sense in yamlmeta.
- converges on coding standards like named returns 
- aligns data values processing by creating DV Schema Preprocessing, very nice!

cari-lynn (approved):
There were a lot of changes in this PR. I read through each commit and while I don't think I have fully digested every single change, this is a great improvement. I did not see any changes that raised concern. I am very happy that this brings schema closer to how data values are processed, and consistency to the patterns in the codebase syntactically and conceptually.

PR #571, opened 2021-12-22
Enable trivy scanning of ytt

joaopapereira:
- [x] Create a secret named SLACK_WEBHOOK_URL that points to a slack webhook or slack workflow.


pivotaljohn:
🙏🏻

Issue #551, opened 2021-11-22
OpenAPI keys are in a uniform order


cari-lynn:
# Why
When keys that take multiple lines are listed prior to keys that are only one line, it can be difficult to visually parse which indentation level corresponds to which key. 

# Current Behavior
## Some keys appear in different order than others
From this data values schema:
```yaml
#@data/values-schema
#@schema/desc "Network configuration values"
---
#@schema/desc "List of database connections"
db_conn:
#@schema/desc "A network entry"
- 
  #@schema/desc "The hostname"
  hostname: ""
  #@schema/desc "Port should be between 49152 through 65535"
  port: 0
  #@schema/desc "Timeout in minutes"
  timeout: 1.0
  #@schema/desc "Any type is allowed"
  #@schema/type any=True
  any_key: thing
  #@schema/desc "When not provided, the default is null"
  #@schema/nullable
  null_key: ""
```
`ytt -f . --data-values-schema-inspect -o openapi-v3`
the result is:
```yaml
type: object
additionalProperties: false
description: Network configuration values
properties:
   db_conn:
     type: array
     description: List of database connections
     items:
       type: object
       additionalProperties: false
       description: A network entry
       properties:
          hostname:
            type: string
            default: ""
            description: The hostname
         port:
           type: integer
           default: 0
           description: Port should be between 49152 through 65535
         timeout:
           type: number
           default: 1
           format: float
           description: Timeout in minutes
         any_key:
           nullable: true                             # <- Nullable first here
           default: thing
           description: Any type is allowed
         null_key:
           type: string
           default: null
           nullable: true                             # <- nullable third here
           description: When not provided, the default is null
     default: []
```

Desired behavior:
* `default` last
* `description` second to last
* `format` after `type`
* keys are in consistent order 
```yaml
type: object
additionalProperties: false
description: Network configuration values
properties:
   db_conn:
     type: array
     description: List of database connections
     items:
       type: object
       additionalProperties: false
       description: A network entry
       properties:
          hostname:
            type: string
            description: The hostname
            default: ""
         port:
           type: integer
           description: Port should be between 49152 through 65535
           default: 0
         timeout:
           type: number
           format: float
           description: Timeout in minutes
           default: 1
         any_key:
           nullable: true
           description: Any type is allowed
           default: thing
         null_key:
           type: string
           nullable: true
           description: When not provided, the default is null
           default: null
     default: []
```

gcheadle-vmware:
Released in ytt v0.39.0

PR #573, opened 2022-01-06
Generate OpenAPI title

vmunishwar:
Generates an Openapi title key from schema/title annotations
so that Configuration Consumers see a short explanation of the key
Openapi title key will appear as a first key except for null_key which will be taken care in [#551 ](https://github.com/vmware-tanzu/carvel-ytt/issues/551)


pkg/schema/schema.go
```diff
@@ -118,6 +118,11 @@ func getType(node yamlmeta.Node) (Type, error) {
 			typeOfValue.SetDescription(desc.description)
 		}
 	}
+	for _, ann := range docAnns {
+		if titleAnn, ok := ann.(*TitleAnnotation); ok {
+			typeOfValue.SetTitle(titleAnn.title)
+		}
+	}
```
pivotaljohn: We now have two cases within the same loop.

How about a single `switch` within the `for` with a `case` for each documentation annotation?

pkg/schema/openapi.go
```diff
@@ -57,11 +58,15 @@ func (o *OpenAPIDocument) calculateProperties(schemaVal interface{}) *yamlmeta.M
...
-			{Key: typeProp, Value: "object"},
-			{Key: "additionalProperties", Value: false},
-		}}
-
+		var property yamlmeta.Map
+		if typedValue.GetTitle() != "" {
+			item := []*yamlmeta.MapItem{
+				{Key: titleProp, Value: typedValue.GetTitle()},
+			}
+			property.Items = append(property.Items, item...)
```
pivotaljohn: `item` need not be a slice of `yamlmeta.MapItem`, we know we're adding exactly one. That makes declaring an array/slice "accidental complexity."

Let's make it a simple pointer to a map item, instead.

pkg/schema/annotations.go
```diff
@@ -297,6 +358,15 @@ func collectDocumentationAnnotations(node yamlmeta.Node) ([]Annotation, error) {
...
 	}
+	for _, annotation := range []template.AnnotationName{AnnotationTitle} {
+		ann, err := processOptionalAnnotation(node, annotation, nil)
+		if err != nil {
+			return nil, err
+		}
+		if ann != nil {
+			anns = append(anns, ann)
+		}
+	}
```
pivotaljohn: We've got duplicate code, here. The `for` loop is iterating over the annotation names.

I think you meant to add `AnnotationTitle` to the list of `AnnotationName`s.

pivotaljohn (changes requested):
Overall, looking good. 👍🏻 

A few tune-ups required.

gcheadle-vmware:
From my perspective, these changes look really good! You extend the existing patterns that are being used to add descriptions, and applied them to adding a title. Your error cases read concisely and are actionable. There were only the few things that John pointed out, that I agree would improve readability and simplicity. Great work!

pkg/schema/schema.go
vmunishwar: Thanks for the review. I have made the changes to the code.

pkg/schema/openapi.go
vmunishwar: Made the changes. The code looks simpler and consistent now.

pkg/schema/annotations.go
vmunishwar: Good catch. I have updated the code.

pkg/cmd/template/schema_inspect_test.go
```diff
@@ -624,6 +623,84 @@ components:
 	})
 }
 
+func TestSchemaInspect_annotation_adds_title_key(t *testing.T) {
```
cari-lynn: Nice job on these tests. For organization, let's move the subtests from ` TestSchemaInspect_annotation_adds_title_key` to the test above: `TestSchemaInspect_annotation_adds_key`. That test tests adding 'documenting' keys when an annotation that provides documentation information is present. As opposed to the other schema annotations we have (`@schema/default`, `@schema/type`, `@schema/nullable`) add constraints to the schema.

cari-lynn (changes requested):
Great changes. I did some manual testing as well using multiple schema annotations per node, and tuples as arguments, and all worked great.

pivotaljohn (approved):
After addressing the item that @cari-lynn brought up (so much love on this PR!), this all LGTM, @vmunishwar.

After you address her feedback and check back in with her, I'd say feel free to merge. 👍🏻

pkg/cmd/template/schema_inspect_test.go
vmunishwar: Thanks for the review. I have moved the subtest as suggested and committed the code.

cari-lynn (approved):

PR #580, opened 2022-01-12
Add Schema examples annotation

gcheadle-vmware:
- [x] include annotation lines in mismatched type error
- [x] collect mismatched type error all provided examples and return together


pkg/schema/annotations.go
```diff
@@ -402,6 +512,66 @@ func getTypeFromAnnotations(anns []Annotation, pos *filepos.Position) (Type, err
...
+	exampleDescription string
+	exampleYAML        interface{}
+}
+
+func setDocumentationFromAnns(docAnns []Annotation, typeOfValue Type) error {
+	var documentationInfo documentation
+	for _, a := range docAnns {
+		switch ann := a.(type) {
+		case *DescriptionAnnotation:
+			documentationInfo.description = documentationInfo.description + ann.description
```
cari-lynn: I am curious about this line. It looks like we are concatenating all the `description` annotations values together here into one string. But I was under the impression that only the first annotation will be used if there are multiple of the same type. Whats the story here?

cari-lynn (approved):
Great job covering all the possible test cases. I'm really happy with the code added here, and think the `description` struct is a good way to group these annotations. 

One piece for discussion:

There is two different syntax for adding an example WITHOUT a description.

`@schema/examples (‘value’,)`

 `@schema/examples (“”, “value”)`
Since if the description is `“”` it will be ignored.

I do like the readability of the second one, I can't imagine a use case where the description needs to be `""`. But I do wonder if it is a good idea to remove the string just because it is zero value.. 🤔

pkg/schema/annotations.go
gcheadle-vmware: That's a good point, I want to maintain the same behavior as before, so if we changed it to `documentationInfo.description = ann.description` would do the trick?

gcheadle-vmware:
> two different syntax for adding an example WITHOUT a description.
>
> @schema/examples (‘value’,)
>
> @schema/examples (“”, “value”)
> Since if the description is “” it will be ignored.

@cari-lynn 
I'm glad you pointed this out, I thought that it was not right to add an empty string as the value to the `x-example-description` key, so I treated the empty string as the default case. I could see a case where a GUI is displaying the example and it's description from an OpenAPI schema, and an author might need to always include the `x-example-description` key since the GUI checks for it. The author in that case may want the ability to include an empty string?

However the `x-example-description` key is [a custom extension to the OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#specification-extensions) so this GUI would have to be built specifically for the schema that `ytt` generates...

pivotaljohn:
> @cari-lynn I'm glad you pointed this out, I thought that it was not right to add an empty string as the value to the `x-example-description` key, so I treated the empty string as the default case. I could see a case where a GUI is displaying the example and it's description from an OpenAPI schema, and an author might need to always include the `x-example-description` key since the GUI checks for it. The author in that case may want the ability to include an empty string?

Oh, as having consumed one-to-many APIs where keys disappear when no value is specified, I really appreciate this thinking. Reminds me of the Go Proverb, "[Make the zero value useful](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=385s)".

> However the `x-example-description` key is [a custom extension to the OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.2.md#specification-extensions) so this GUI would have to be built specifically for the schema that `ytt` generates...

And it's the nature of maps/dictionaries/key-value pairs that such entries can easily/safely be ignored if not needed. 👍🏻

pkg/cmd/template/schema_author_test.go
```diff
@@ -756,6 +756,198 @@ schema.yml:
...
+
+syntax error in @schema/examples annotation
+schema.yml:
+    |
+  3 | #@schema/examples
+  4 | key: val
+    |
+
+    = found: missing value in @schema/examples (by schema.yml:3)
+    = expected: tuple with description and example
```
pivotaljohn:
```suggestion
    = expected: tuple containing description and example
```
nit.

pkg/cmd/template/schema_author_test.go
```diff
@@ -756,6 +756,198 @@ schema.yml:
...
+==============
+
+syntax error in @schema/examples annotation
+schema.yml:
+    |
+  3 | #@schema/examples "example value"
+  4 | key: val
+    |
+
+    = found: non tuple in @schema/examples (by schema.yml:3)
```
pivotaljohn: Do we have access to the type of the actual argument? We'll learn in the expectation that "tuple" is what we want. 

```suggestion
    = found: string for @schema/examples (at schema.yml:3)
```

nit: s/by/at/ ??

pkg/cmd/template/schema_author_test.go
```diff
@@ -756,6 +756,198 @@ schema.yml:
...
+
+syntax error in @schema/examples annotation
+schema.yml:
+    |
+  3 | #@schema/examples ()
+  4 | key: val
+    |
+
+    = found: empty tuple in @schema/examples (by schema.yml:3)
+    = expected: tuple with optional string description and required example value
```
pivotaljohn:
```suggestion
    = expected: 2-tuple containing description (string) and example value (of expected type)
```

pkg/cmd/template/schema_author_test.go
```diff
@@ -756,6 +756,198 @@ schema.yml:
...
+syntax error in @schema/examples annotation
+schema.yml:
+    |
+  3 | #@schema/examples ("key example", "val", "value")
+  4 | key: val
+    |
+
+    = found: 3 arguments in @schema/examples (by schema.yml:3)
+    = expected: no more than 2 arguments per tuple. e.g. @schema/examples ('description', exampleValue())
+`
```
pivotaljohn:
```suggestion
    = found: 3-tuple argument in @schema/examples (by schema.yml:3)
    = expected: 2-tuple containing description (string) and example value (of expected type)
`
```

nit: pedantically, it's the requisite one argument, but has the wrong cardinality.

pkg/cmd/template/schema_author_test.go
```diff
@@ -756,6 +756,198 @@ schema.yml:
...
+
+syntax error in @schema/examples annotation
+schema.yml:
+    |
+  3 | #@schema/examples (3.14, 5)
+  4 | key: 7.3
+    |
+
+    = found: Non-string value in @schema/examples (by schema.yml:3)
+    = expected: string
```
pivotaljohn:
```suggestion
    = found: float value for @schema/examples (at schema.yml:3)
    = expected: 2-tuple containing description (string) and example value (of expected type)
```

pkg/schema/annotations.go
```diff
@@ -56,6 +58,13 @@ type TitleAnnotation struct {
 	pos   *filepos.Position
 }
 
+// ExampleAnnotation documents an example of a node
+type ExampleAnnotation struct {
+	descriptions []string
+	examples     []interface{}
+	pos          *filepos.Position
+}
+
```
pivotaljohn: 🤔  Interesting. I had some expectation that there'd be a single slice of examples: a struct with description and value (i.e. mapping closely to what the user inputs)

Had you considered that approach and found it less favorable to storing the descriptions and examples separately?

pkg/cmd/template/schema_author_test.go
```diff
@@ -756,6 +756,198 @@ schema.yml:
...
+syntax error in @schema/examples annotation
+schema.yml:
+    |
+  3 | #@schema/examples "example value"
+  4 | key: val
+    |
+
+    = found: non tuple in @schema/examples (by schema.yml:3)
+    = expected: tuple with optional string description and required example value
+    = hint: use a trailing comma to construct tuple with a single value. e.g. ('example value',)
```
pivotaljohn: I believe this would be a non-empty description _with_ a `None` value...?

I think you mean to hint: `(, "example value")`

pkg/schema/annotations.go
```diff
@@ -249,6 +258,91 @@ func NewTitleAnnotation(ann template.NodeAnnotation, pos *filepos.Position) (*Ti
...
+			}
+		} else {
+			return nil, schemaAssertionError{
+				annPositions: []*filepos.Position{ann.Position},
+				position:     pos,
+				description:  fmt.Sprintf("syntax error in @%v annotation", AnnotationExamples),
+				expected:     fmt.Sprintf("tuple with optional string description and required example value"),
+				found:        fmt.Sprintf("non tuple in @%v (by %v)", AnnotationExamples, ann.Position.AsCompactString()),
+				hints:        []string{"use a trailing comma to construct tuple with a single value. e.g. ('example value',)"},
+			}
```
pivotaljohn: If we negated the check for whether this is a tuple and pulled this up above the switch, it would pull in this whole body of code in by one indentation.

I think of checks like that (known also as "guard clauses") as a kind of pre-condition ... that once execution passes that point, the reader can safely assume the condition has been met: fewer things to keep track of in my head as I'm grokking the code.

pkg/schema/annotations.go
```diff
@@ -249,6 +258,91 @@ func NewTitleAnnotation(ann template.NodeAnnotation, pos *filepos.Position) (*Ti
...
+				}
+			default:
+				return nil, schemaAssertionError{
+					annPositions: []*filepos.Position{ann.Position},
+					position:     pos,
+					description:  fmt.Sprintf("syntax error in @%v annotation", AnnotationExamples),
+					expected:     fmt.Sprintf("no more than 2 arguments per tuple. e.g. @%v ('description', exampleValue())", AnnotationExamples),
+					found:        fmt.Sprintf("%v arguments in @%v (by %v)", len(exampleTuple), AnnotationExamples, ann.Position.AsCompactString()),
+				}
+			}
```
pivotaljohn: 🤔 💭 
I wonder how this would read if all of the error conditions were captured in cases and the `default` is the success case...

```go
switch {
  case len(exampleTuple) == 0: // error
  case len(exampleTuple) == 1: // error
  case len(exampleTuple) > 2:  // error
  default: // success
}
```
And there's no rule that says that you _have_ to have a `default` case.... it might visually help if that code just follows the `switch` entirely?!?!

_(related to next comment...)_

pkg/schema/annotations.go
```diff
@@ -469,6 +579,69 @@ func getTypeFromAnnotations(anns []Annotation, pos *filepos.Position) (Type, err
...
+}
+
+func checkExamplesValue(ann *ExampleAnnotation, typeOfValue Type) error {
+	var typeCheck TypeCheck
+	for _, ex := range ann.examples {
+		if node, ok := ex.(yamlmeta.Node); ok {
+			defaultValue := node.DeepCopyAsNode()
+			chk := typeOfValue.AssignTypeTo(defaultValue)
+			if !chk.HasViolations() {
+				chk = CheckDocument(defaultValue)
```
pivotaljohn, line 613: Perhaps `CheckDocument()` needs a rename...?

ideas:
- `CheckNodes()` — hint that this is a tree?
- `CheckAST()` — explicitly call out the tree? (eeeehh... not sure how widely we use AST in places other than yamlmeta....)
- `CheckNode()` — is it implied that we mean both the Node and children?

pkg/schema/type.go
```diff
@@ -21,10 +21,10 @@ type Type interface {
 	GetDefinitionPosition() *filepos.Position
 
 	GetDescription() string
-	SetDescription(string)
-	String() string
-	SetTitle(string)
 	GetTitle() string
+	GetExample() (string, interface{})
+	SetDocumentation(documentation)
```
pivotaljohn: This signature will make it not possible for code outside of `schema` to be able to set documentation...

... have you already considered providing setters for `Title` and `Description` and then maybe something like `AddExample(..)`? or `SetExamples(...)`?

Related:
- Today, we can only use the first example in OpenAPI v3.0 output
- But when we turn around and support OpenAPI v3.1 (and that _will_ happen), we'll want to be able to store multiple examples.
- Also, we'd love to be able to support _multiple_ examples in `ytt` Schema, today; no need to limit _that_ part of these system.
- Consider what happens when we push the limitation right up to the seam where it's introduced (i.e. when building the OpenAPI output).

pivotaljohn (changes requested):

pkg/cmd/template/schema_author_test.go
cari-lynn: (I looked into this a bit so im commenting what I found) I thought `('example value',)` was odd too. But I believe the trailing comma is the correct syntax. That makes it a single-element tuple. I believe the leading comma syntax isn't supported. 
 https://wiki.python.org/moin/TupleSyntax#:~:text=Multiple%20Element%20Tuples&text=Multiple%2Delement%20tuples%20may%20be,be%20enclosed%20in%20parentheses%2C%20e.g.

cari-lynn:
> I thought that it was not right to add an empty string as the value to the x-example-description key, so I treated the empty string as the default case

I can see adding the `x-example-description` key adding noise to an already dense schema and not a lot of value. 

> And it's the nature of maps/dictionaries/key-value pairs that such entries can easily/safely be ignored if not needed. 👍🏻

Forgive me for being thick-skulled, do you mean that the empty string in the case of `("", "value")` can be ignored if not needed or do you mean that OpenAPI schema _consumers_ can ignore `x-example-description` if not needed?

For users that want to create a tuple without a description we could encourage users to use the `('example value',)` syntax, while having the `('', 'example value')` syntax behave as predicted and add `x-example-description: ""`.

cari-lynn:
One more thing I was experimenting with is using a yamlFragment in the annotation to clean things up:

`#@schema/examples func()`

But I believe there is no way to specify a tuple in yaml.. which makes abstracting this to a function not possible using yamlFragements.

pivotaljohn:
> > GC: I thought that it was not right to add an empty string as the value to the x-example-description key, so I treated the empty string as the default case
> 
> CLD: I can see adding the `x-example-description` key adding noise to an already dense schema and not a lot of value.

I agree that if its not typically used, it's more noise than signal. But by making it optional (i.e. sometimes present, sometimes not), makes that "API" a bit more complicated to use (noted, below).
 
> > JSR: And it's the nature of maps/dictionaries/key-value pairs that such entries can easily/safely be ignored if not needed. 👍🏻
> 
> CLD: Forgive me for being thick-skulled, do you mean that the empty string in the case of `("", "value")` can be ignored if not needed or do you mean that OpenAPI schema _consumers_ can ignore `x-example-description` if not needed?

The latter: that it's (usually) easy to ignore an entry in a map. I'm drawing an analogy with JSON APIs: that it makes for an easier interface to program to when every time there's an `example:` property, it _always_ has an `x-example-description`... even if an empty string. It's analogous to a field that's zero vs. not present/defined.

> For users that want to create a tuple without a description we could encourage users to use the `('example value',)` syntax, while having the `('', 'example value')` syntax behave as predicted and add `x-example-description: ""`.

Wouldn't that mean that _sometimes_ the first element in the tuple is the `example value` and other-times it would be the second element in the tuple?

pkg/schema/annotations.go
pivotaljohn: handled in a subsequent PR. 👍🏻

gcheadle-vmware:
After syncing with John, I decided to go with a simpler design for the annotation, where the description and example are always required. If the user does not want to include a description, they can just provide an empty string and the `x-example-description` key will just have an empty string.

pkg/cmd/template/schema_author_test.go
gcheadle-vmware: See comment below about disallowing this for now.

cari-lynn:
Commenting here to wrap up the threads I started. 
> [pivotalJohn] it always has an x-example-description... even if an empty string

Sounds good. I agree with the decisions above 👍  If we hear from users that this is noisy, in the future we can add a flag or syntax to omit this.

> [me] But I believe there is no way to specify a tuple in yaml.. which makes abstracting this to a function not possible using yamlFragements.

I am satisfied with being able to use a syntax to declare the `example` in a yamlFragment, and not necessarily being able to declare the entire tuple in a yamlFragment. Eg:
`#@schema/example ("host", hostExample())`

Thank you!!

gcheadle-vmware:
@pivotaljohn, @cari-lynn - changes made to type interface to reflect the conversation earlier today.

pkg/cmd/template/schema_author_test.go
```diff
@@ -756,6 +756,196 @@ schema.yml:
...
+
+schema.yml:
+    |
+  3 | #@schema/examples ("Zero value", 0)
+  4 | enabled: false
+    |
+
+    = found: integer (by schema.yml:3)
+    = expected: boolean (by schema.yml:4)
+    = hint: is the default value set using @schema/default?
```
pivotaljohn, line 942: Just note: the hint is misplaced and is fixed in #582. This is a new instance of the hint and will need to be removed after _that_ PR is merged.

pivotaljohn (approved):
Looks great, @gcheadle-vmware.

Note the minor interaction with #582 — there's a hint in here that (should) go away when _it_ merges.

Will leave it to you to decide which order you want these merged to minimize the effort on your part. Ping me if you want a buddy in that.

PR #582, opened 2022-01-13
Include default annotation position when default value is wrong type

gcheadle-vmware:
- removes `hint: is the default value set with @schema/default` from type mismatch error messages
- checks the value of a default annotation during the `getValue()` operation of constructing a schema. We know that schema is build depth first, so any child that also needs to be type checked will have already been typed and processed by schema. Previously, if the default annotation had nested values, we deferred the type check to data values pre-processing, which lost all annotation position information.  

*Note:* This PR can be merged separately, but is related to #580

### DIFF
--- a/pkg/cmd/template/schema_author_test.go
+++ b/pkg/cmd/template/schema_author_test.go
@@ -493,12 +493,12 @@ Invalid schema - @schema/default is wrong type
 
 schema.yml:
     |
+  3 | #@schema/default 1
   4 | foo: a string
     |
 
-    = found: integer
+    = found: integer (at schema.yml:3)
     = expected: string (by schema.yml:4)
-    = hint: is the default value set using @schema/default?
 `
 
 				filesToProcess := files.NewSortedFiles([]*files.File{
@@ -521,12 +521,12 @@ Invalid schema - @schema/default is wrong type
 
 schema.yml:
     |
+  3 | #@schema/default [{"item": 1}]
   4 | map: thing
     |
 
-    = found: array
+    = found: array (at schema.yml:3)
     = expected: string (by schema.yml:4)
-    = hint: is the default value set using @schema/default?
 `
 
 				filesToProcess := files.NewSortedFiles([]*files.File{
@@ -939,7 +939,6 @@ schema.yml:
 
     = found: integer (by schema.yml:3)
     = expected: boolean (by schema.yml:4)
-    = hint: is the default value set using @schema/default?
 `
 
 			filesToProcess := files.NewSortedFiles([]*files.File{
--- a/pkg/schema/check.go
+++ b/pkg/schema/check.go
@@ -9,14 +9,14 @@ import (
 	"github.com/k14s/ytt/pkg/yamlmeta"
 )
 
-// CheckDocument attempts type check of `doc`.
+// CheckNode attempts type check of root node and its children.
 //
-// If `d` has "schema/type" metadata (typically attached using SetType()), `d` is checked against that schema, recursively.
+// If `n` has "schema/type" metadata (typically attached using SetType()), `n` is checked against that schema, recursively.
 // `chk` contains all the type violations found in the check.
-func CheckDocument(doc yamlmeta.Node) TypeCheck {
+func CheckNode(n yamlmeta.Node) TypeCheck {
 	checker := newTypeChecker()
 
-	err := yamlmeta.Walk(doc, checker)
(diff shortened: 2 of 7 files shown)


pkg/cmd/template/schema_author_test.go
```diff
@@ -493,12 +493,12 @@ Invalid schema - @schema/default is wrong type
 
 schema.yml:
     |
+  3 | #@schema/default 1
   4 | foo: a string
     |
 
-    = found: integer
+    = found: integer (by schema.yml:3)
```
pivotaljohn:
```suggestion
    = found: integer (at schema.yml:3)
```

🤷🏻 

My case for "at" vs "by" is that "by" makes more sense for the "expected" situation... "string is expected _by_ <file:pos>" vs. "an integer was found _at_ <file:pos>"

pkg/cmd/template/schema_author_test.go
```diff
@@ -521,12 +521,12 @@ Invalid schema - @schema/default is wrong type
...
 schema.yml:
     |
+  3 | #@schema/default [{"item": 1}]
   4 | map: thing
     |
 
-    = found: array
+    = found: array (by schema.yml:3)
     = expected: string (by schema.yml:4)
-    = hint: is the default value set using @schema/default?
```
pivotaljohn, line 529: Thank you, hints, you did your service. Also, delighted to see us evolving passed this form. 👍🏻

pkg/schema/check.go
```diff
@@ -9,11 +9,11 @@ import (
 	"github.com/k14s/ytt/pkg/yamlmeta"
 )
 
-// CheckDocument attempts type check of `doc`.
+// CheckNode attempts type check of root node and its children.
 //
 // If `d` has "schema/type" metadata (typically attached using SetType()), `d` is checked against that schema, recursively.
 // `chk` contains all the type violations found in the check.
-func CheckDocument(doc *yamlmeta.Document) TypeCheck {
+func CheckNode(doc yamlmeta.Node) TypeCheck {
```
pivotaljohn: HA! I reviewed #580 first and made this suggestion. I'm sure you'll point my attention back here. 😆  Good stuff.

pkg/schema/schema.go
```diff
@@ -153,12 +153,27 @@ func getValueFromAnn(defaultAnn *DefaultAnnotation, t Type) (interface{}, error)
 	defaultValue := defaultAnn.Val()
 	if node, ok := defaultValue.(yamlmeta.Node); ok {
 		defaultValue = node.DeepCopyAsInterface()
-		typeCheck = t.AssignTypeTo(defaultValue.(yamlmeta.Node))
+		val := defaultValue.(yamlmeta.Node)
+		typeCheck = t.AssignTypeTo(val)
+		if !typeCheck.HasViolations() {
+			typeCheck = CheckNode(val)
+		}
```
pivotaljohn: Does it make sense to use `node.DeepCopy()`, here? That method returns a `yamlmeta.Node` already... no need to cast!

pivotaljohn (approved):
There are a few tweaks to suggest.

But all-in-all, this is a wonderful improvement. 💚

pkg/schema/schema.go
gcheadle-vmware: When changing the code to the following:
```
val := node.DeepCopyAsNode()
typeCheck = t.AssignTypeTo(val)
```
The following tests fail:
TestSchemaInspect_exports_an_OpenAPI_doc/including_explicitly_set_default_values
TestSchemaInspect_exports_an_OpenAPI_doc/including_nullable_values_with_defaults

I have not investigated this further, but I would be interested if @pivotaljohn or @cari-lynn had any idea why this is happening?

Seems to not be adding the default `ree` key when the default value came from `@schema/default` ann.

pivotaljohn: The tests are failing because we need to assign the copy _back_ to `defaultValue` (it's what's getting merged in).

I now realize that it's not as simple as just returning a `Node` because the static type of `defaultValue` is `interface{}`.

One might argue that the following is _slightly_ better, given that `Node.DeepCopyAsNode()` guarantees that it returns a `Node`. 🤷🏻 
```go
defaultNode := node.DeepCopyAsNode()
typeCheck = t.AssignTypeTo(defaultNode)
if !typeCheck.HasViolations() {
	typeCheck = CheckNode(defaultNode)
}
defaultValue = defaultNode
```

PR #595, opened 2022-01-22
Order open api schema keys

gcheadle-vmware:


pkg/schema/openapi.go
```diff
@@ -55,64 +90,82 @@ func (o *OpenAPIDocument) calculateProperties(schemaVal interface{}) *yamlmeta.M
...
-		property.Items = append(property.Items, &yamlmeta.MapItem{Key: typeProp, Value: "array"})
-		property.Items = append(property.Items, &yamlmeta.MapItem{Key: "items", Value: properties})
-		property.Items = append(property.Items, &yamlmeta.MapItem{Key: defaultProp, Value: typedValue.GetDefaultValue()})
-		return &property
+		sort.Sort(items)
+		return &yamlmeta.Map{Items: items}
 	case *ScalarType:
+		var items openAPIKeys
+		items = append(items, collectDocumentation(typedValue)...)
+		items = append(items, &yamlmeta.MapItem{Key: defaultProp, Value: typedValue.GetDefaultValue()})
```
cari-lynn, line 120: Love how much easier these cases are to read now that we don't have to think about ordering explicitly.

pkg/schema/openapi.go
```diff
@@ -5,21 +5,56 @@ package schema
...
 
 // keys used when generating an OpenAPI Document
 const (
+	titleProp              = "title"
 	typeProp               = "type"
-	defaultProp            = "default"
+	additionalPropsProp    = "additionalProperties"
+	formatProp             = "format"
 	nullableProp           = "nullable"
+	deprecatedProp         = "deprecated"
```
cari-lynn: There is another story in flight to add deprecated, so it's not yet tested even though it's here. I know there has been a lot of collaboration on this, but I think we should take this key out, and add it in the story for `@schema/deprecated`. 

For example, it is possible that this PR get accepted, and we have this key for `deprecated` here. Then pretend some new information comes to light and we no longer want to move forward with the `@schema/deprecated` feature at all. Then this code will never be used. While it is unlikely, I think it's good practice to wait.

cari-lynn (commented):
Looks good, I just have one comment 👍  Love the use of the sorter interface to do this.

pkg/schema/openapi.go
gcheadle-vmware: Makes sense, I will remove it.

PR #596, opened 2022-01-28
Openapi schema/deprecated annotation

vmunishwar:


gcheadle-vmware (commented):
These changes look good to me, good job with this story from start to finish!

cari-lynn (changes requested):
The existing code LGTM and is great work, but[ we are missing](https://kubernetes.slack.com/archives/CH8KCCKA5/p1643226033163800) a test and error message for disallowing the `@schema/deprecated` on a schema document.

The error should mention that annotating a document as deprecated is disallowed, and instead encourage users to annotate individual keys.

vmunishwar:
> The existing code LGTM and is great work, but[ we are missing](https://kubernetes.slack.com/archives/CH8KCCKA5/p1643226033163800) a test and error message for disallowing the `@schema/deprecated` on a schema document.
> 
> The error should mention that annotating a document as deprecated is disallowed, and instead encourage users to annotate individual keys.

Thanks for the comments, Cari! I have added the test case and error message for disallowing `@schema/deprecated` on a schema document.

pkg/schema/annotations.go
```diff
@@ -512,6 +574,20 @@ func processOptionalAnnotation(node yamlmeta.Node, optionalAnnotation template.A
 				return nil, err
 			}
 			return defaultAnn, nil
+		case AnnotationDeprecated:
+			if _, ok := node.(*yamlmeta.Document); ok {
+				return nil, schemaAssertionError{
+					description:  fmt.Sprintf("@%v not supported on a document", AnnotationDeprecated),
```
cari-lynn: To be consistant with other errors (see pkg/schema/annotations.go:559)
```suggestion
					description:  fmt.Sprintf("Invalid schema - @%v not supported on a %s", AnnotationDeprecated, yamlmeta.TypeName(node)),
```

pkg/schema/type.go
```diff
@@ -26,6 +26,8 @@ type Type interface {
 	SetTitle(string)
 	GetExamples() []Example
 	SetExamples([]Example)
+	GetDeprecated() bool
+	SetDeprecated(bool)
```
cari-lynn: Since this is our API if users are integrating with ytt programmatically, users may need to know two things:
1. _is_ the value deprecated
2. what the value for the deprecated string is (even though we do not yet use this, we still can expose this info. A integrator may want this info before we get around to prioritizing the story to use it, so lets make it available)

Lets change the Type interface to expose these things:
```
type Type interface {
...
   IsDeprecated() bool
   SetDeprecated(string)             <- takes the string instead of a bool
   GetDeprecationNotice() string
...
}
```

cari-lynn (changes requested):

pkg/schema/annotations.go
vmunishwar: Thanks for the comment! I had removed "Invalid schema" to avoid duplicate error message. Made remaining changes.

pkg/schema/type.go
cari-lynn: After talking about this over zoom, we dicided to change the interface to:
```
IsDeprecated() (bool, string)
SetDeprecated(bool, string)
```

vmunishwar: Thanks Cari! I have made the changes as discussed.

pkg/schema/type.go
```diff
@@ -26,6 +26,8 @@ type Type interface {
 	SetTitle(string)
 	GetExamples() []Example
 	SetExamples([]Example)
+	IsDeprecated() (bool, string)
+	SetDeprecated(deprecated bool, notice string)
```
cari-lynn:
```suggestion
	SetDeprecated(bool, string)
```

We can remove the names here

pkg/schema/type.go
```diff
@@ -482,6 +484,85 @@ func (n *NullType) SetExamples(exs []Example) {
...
 }
 
+// IsDeprecated provides deprecated field information
+func (t *DocumentType) IsDeprecated() (bool, string) {
+	return false, ""
+}
+
+// IsDeprecated provides deprecated field information
+func (m *MapType) IsDeprecated() (bool, string) {
+	return m.documentation.deprecated, ""
```
cari-lynn: For each of these lets return the deprecated notice as well, so that consumers of the api can use it.

```suggestion
	return m.documentation.deprecated, a.documentation.deprecationNotice
```

cari-lynn (changes requested):
Just a couple minor comments 👍

pkg/schema/type.go
vmunishwar: Done. Thanks Cari!

cari-lynn (approved):
Great work! Thanks for addressing all my comments 🎆

PR #606, opened 2022-02-16
Update Go Module name to "github.com/vmware-tanzu/carvel-ytt"

gcheadle-vmware:


gcheadle-vmware:
directory `examples/integrating-with-ytt/internal-templating/` imports ytt v0.38.0 with module: `github.com/k14s/ytt v0.38.0`. Since v0.40.0 will be the first release with the new module name, this dependency will not need to be updated.

cari-lynn (approved):
This looks good to me, it's my understanding that the previous import paths: `github.com/k14s/ytt` will continue working with version `v0.39.0` or earlier, and then to update to `v0.40.0` the import path will need to be changed to `github.com/vmware-tanzu/carvel-ytt`. So integrators shouldn't have any interruptions.

(not required as part of this PR) After we cut `v0.40.0` with the new module, we should update the `examples/integrating-with-ytt/internal-templating/` directory since [it mentions](examples/integrating-with-ytt/internal-templating/) that the module will be renamed.

pivotaljohn (approved):
LGTM.

PR #609, opened 2022-02-18
Allow `:` in `--data-values-file` path

cari-lynn:
UPDATE 3/23:

## What
Prior to this PR we were using just `:` to detect a library reference, but this PR introduces using a prefix of `@`, and suffix of `:` to detect it in all flags. This doesn't change any behavior since this syntax was already required. This change avoids ytt confusing the `:` that can come from a prefix of `@libraryReference:`, and a `:` that comes from a file path.

## How
These flags are unaffected by this change:
```shell
--data-values-env
--data-values-env-yaml
--data-value
--data-value-yaml
--data-value-file
```
These flags have a flag value in the format `@lib:key=value`, which is why they are not changed, there is no change needed to allow `:` in the `value` because it is not currently ambiguous.

This is the only flag affected: `--data-values-file`. This is affected because it does not have a key and it is ambiguous where a library reference and a file path value are separated.

It is in the format `@lib:value`. `:` can appear in the `@lib:` and anywhere in the `value`. So this PR makes a change to allow the `:` to appear anywhere in the `value`.

In a case breakdown for `--data-values-file` is:
* case 1: No library reference. Examples: “path.yml”, “@path.yml”, “c:path.yml”, “:::::path:::::.yml”, “pa@lib:th.yml”
* case 2: Library reference. Examples: “@~libref:path.yml”, “@lib@lib1:”, “@lib:@lib1:path.yml <- o.o
* case 3: Error. Examples: “@:path.yml”, “@:”


pkg/cmd/template/data_values_flags.go
```diff
@@ -288,11 +289,18 @@ func (DataValuesFlags) libraryRefAndKey(key string) (string, string, error) {
...
+		}
+		// the LibrarySep is part of the path
+		return "", key, nil
+	case 3:
+		if strings.HasPrefix(keyPieces[0], ref.LibrarySep) {
+			// the second LibrarySep is part of the path
+			return keyPieces[0], keyPieces[1] + libraryKeySep + keyPieces[2], nil
+		}
 	}
+	return "", "", fmt.Errorf("Expected at most one library-key separator '%s' in '%s'", libraryKeySep, key)
```
gcheadle-vmware: This seems to contradict the fact that you have `case 3` in your switch statement. Currently, this code reads to me as:
- It's okay to have two ":"s if the whole string is prefixed with "@". Everything after the first ":" is the path to the file.
- otherwise, two or more ":"s is an error.

On a Mac, there would certainly be an error later on about trying to read the file with a weird name, but I wanted to point out that the bullets presented in this comment did not match my assumption that multiple ":"s would only be considered in the windows case.

gcheadle-vmware (commented):
This change set took a while for me to digest, and honestly the test is difficult to attempt manually without a windows machine. The implementation was surprising to me since the issue mentions the `libraryKeySep`, but that const is not altered. After some time, I was able to build a mental model for the switch statement:
- case 1: "path.yml" - no libref case
- case 2: "@libref:path.yml"  - normal case, so confirm by checking prefix (@)
- case 2: "c://Users/path.yml" - windows case, return back full string since no prefix (@)
- case 3: "@lib:c://Users/path.yml" - windows case, repair unintentional split on full path

These cases make sense to me and the implementation tracks, and even though there are many ways to supply data values via flags, I think your change will catch all cases. 

I do wonder if a Mac user that has multiple ":" in an argument, like trying to (incorrectly) ref two libraries with a single data value file: `@lib:@otherLib:path`. I believe this arg would return a path like "@otherLib:path" for the file location, instead of reporting an err (err would happen when attempting to read file from system). 🤔

pkg/cmd/template/data_values_flags.go
cari-lynn: I updated the PR to allow 0+ `:` characters in any file path. I also learned that file names in unix can contain `:` so we should also allow them.

pkg/cmd/template/data_values_flags.go
```diff
@@ -135,15 +137,15 @@ func (s *DataValuesFlags) AsOverlays(strict bool) ([]*datavalues.Envelope, []*da
...
+func (s *DataValuesFlags) file(fullPath string, strict bool) ([]*datavalues.Envelope, error) {
+	libRef, path, err := s.libraryRefAndKey(fullPath)
 	if err != nil {
 		return nil, err
 	}
 
 	contents, err := s.readFile(path)
 	if err != nil {
-		return nil, fmt.Errorf("Reading file '%s': %s", path, err)
+		return nil, fmt.Errorf("Unable to read file '%s' from flag value '%s': %s", path, err)
```
gcheadle-vmware: There are only two arguments here so the error message is showing me "MISSING" after the colon.

gcheadle-vmware:
@cari-lynn, I took another look at these changes, and I think they look great. By improving the "Unable to read file error" you are improving the ytt experience across operating systems. I also appreciate how you've done research to inform the decisions being made.

pkg/cmd/template/data_values_flags.go
gcheadle-vmware: FYI `./hack/test-all.sh` did not pass for me when I pulled and ran it locally

cari-lynn: oops, this was a bad merge

pkg/cmd/template/data_values_flags.go
```diff
@@ -274,26 +276,36 @@ func (s *DataValuesFlags) kvFile(kv string) (*datavalues.Envelope, error) {
...
+	}
+	return libRef, key, nil
+}
 
-	case 2:
-		if len(keyPieces[0]) == 0 {
-			return "", "", fmt.Errorf("Expected library ref to not be empty")
+// libraryRefAndString separates a library reference and a string.
+// A library reference starts with ref.LibrarySep, and ends with the first occurrence of libraryKeySep.
+func (DataValuesFlags) libraryRefAndString(str string) (string, string, error) {
```
pivotaljohn:
```suggestion
func (DataValuesFlags) libraryRefAndRemainder(arg string) (string, string, error) {
```

Part one: the method name....

Trying to think of a more concrete name for the second return value... (burned through a timebox over here)

Ideally, it would be a domain term ("library ref" is a domain term; what the rest of this argument called?):
- in the one case, it's the key/value... what is that?
  - Setting
  - Assignment
  - er... KeyValuePair
- in the another case, it's the path to a file that contains key/values (`--data-values-file`)
  - path
  - (sortof) KeyValuePairs

But nothin' gellin'... so perhaps the name of the role of the value in calculations:
- Remainder
- TheRest :)

I test these ideas out by looking at the _calling_ spot to see if as a reader, I have a chance of not having to read the body of the called function to get the gist of what's going on.


Part two: the parameter name...
It's more than the key, it's the whole argument. So... `arg`?

pkg/cmd/template/data_values_flags.go
```diff
@@ -274,26 +276,36 @@ func (s *DataValuesFlags) kvFile(kv string) (*datavalues.Envelope, error) {
 	return datavalues.NewEnvelopeWithLibRef(overlay, libRef)
 }
 
+// libraryRefAndKey separates a library reference and a key and validates that no libraryKeySep exist in the key.
+// libraryKeySep is disallowed in data value flag keys.
 func (DataValuesFlags) libraryRefAndKey(key string) (string, string, error) {
```
pivotaljohn:
```suggestion
func (DataValuesFlags) libraryRefAndKey(arg string) (string, string, error) {
```
Input is the whole argument, so `arg`?
2nd output is the key 👍🏻 

(not part of your changeset, but inspired by the comment below)

pivotaljohn (approved):
Left a couple of naming suggestions.

Otherwise, looks great!

pkg/cmd/template/data_values_flags.go
cari-lynn: `libraryRefAndRemainder` is more clear 👍 , making that change also made it easier to update the godoc to be more descriptive.

gcheadle-vmware (approved):
LTGM!

PR #617, opened 2022-03-01
Accept stdin for data-values-file flag

vmunishwar:


pkg/cmd/template/data_values_flags.go
```diff
@@ -1,10 +1,11 @@
-// Copyright 2020 VMware, Inc.
+// Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
 package template
 
 import (
 	"fmt"
+	"github.com/vmware-tanzu/carvel-ytt/pkg/files"
```
gcheadle-vmware: nit: we like to order the imports alphabetically, and separated. We put the standard imports (like fmt, strings) first, and the custom imports last, separated by a blank line. For example:
```
"fmt"
"io/ioutil"
"os"
"strings"

"github.com/vmware-tanzu/carvel-ytt/pkg/files"
```

pkg/files/stdin.go
```diff
@@ -0,0 +1,21 @@
...
+
+package files
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+)
+
+var hasStdinBeenRead bool
```
gcheadle-vmware, line 12: I like the simplicity of this solution. It relies on this file being the only way we read from standard input, but since that is in fact the case, a simple bool will let you know if this function has been called.

pkg/files/stdin.go
```diff
@@ -0,0 +1,21 @@
...
+
+//ReadStdin only read stdin once
+func ReadStdin() ([]byte, error) {
+	if hasStdinBeenRead {
+		return nil, fmt.Errorf("Standard input has been already read once (has the '-' argument been used in more than one flags?)")
```
gcheadle-vmware: nit: reword error to "Standard input has already been read, has the '-' argument been used in more than one flag?"

gcheadle-vmware (approved):
Overall this change looks really good. You use the systems that are already in place to your advantage, and that results in a small but effective change set🥇 . 

I did a small amount of manual testing and was able to use std-in for the (out of scope) flag `--data-value-file`, which takes key=value pairs, where a key is a path to a data value, and the value is a path to a file which is read as a string:
❌ `cat val.txt | ./ytt [...] --data-value-file=-` where `cat val.txt` was "string=new".
❌ `cat val.txt | ./ytt [...] --data-value-file -` where `cat val.txt` was "string=new".
✅ `cat val.txt | ./ytt [...] --data-value-file key.otherKey=-`  where `cat val.txt` was "new".

I'm not sure what the exact desired behavior for this flag is, plus it is out of scope for this story, but I thought it would be good to mention that your solution did in-fact extend to include the other data values flags.

pkg/cmd/template/data_values_flags.go
vmunishwar: Thanks for letting me know about this ordering. Have made the changes accordingly.

test/e2e/e2e_test.go
```diff
@@ -461,7 +489,36 @@ func runYtt(t *testing.T, files testInputFiles, stdinFileName string, flags yttF
...
+	stdError := bytes.NewBufferString("")
+	command.Stderr = stdError
+	command.Env = append(command.Env, envs...)
+
+	_, err := command.Output()
+	require.Error(t, err, stdError.String())
+
+	return stdError.String()
+
+}
```
pivotaljohn: Duplicating a function that _almost_ but _doesn't quite_ fit is a terrific first move in situations like this.

...

Also, given that test code isn't _itself_ covered, we tend to get better results by leaving that test subsystem a bit more "DAMP" (i.e. less Do not Repeat Yourself). 👍🏻 

So, the question arises at this point: should the common code between `runYtt()` and `runYttExpectingError()` be extracted?
In my mind, we're trying to balance "correctness" with "maintainability."
- Correctness: we can be more confident that our test code is correct when there are fewer levels of indirection — it's more straight-forward to read.
- Maintainability: the more modular our code (including minimizing duplication), the less effort it takes to affect changes.

In this very specific case, I think the core question I ask myself is:

_How fragmented are the differences between the two functions (i.e. are there differences throughout? or are they all in one place?)_

pkg/files/stdin.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+)
+
+var hasStdinBeenRead bool
+
+//ReadStdin only read stdin once
```
pivotaljohn: nit: typically we leave a space between the comment token(s) and the start of the comment.

pivotaljohn (approved):
Looks great, Varsha!

There are two suggestion in this review, below.

Once you make any corresponding edits (if any), feel free to merge. 👍🏻

test/e2e/e2e_test.go
vmunishwar: Thanks John for the detailed review. I have refactored the code to minimize the duplication.

cari-lynn (approved):
Love to see this small and precise change set 👍 

Just wanted to call it out that there is no way to use the `ytt --data-values-file @lib:-` syntax to specify the data value is for a library. This feature still holds its own without that, and we can add support for that at a later time I believe.

PR #620, opened 2022-03-02
add @assert/validate for standard Data Values files

gcheadle-vmware:
interesting points of conversation:
- Errors can get long and repetitive if the user has an assert.fail
  - A: place error list on "-"s, use as much text from the Validation as possible, investigate later if I can remove double fail.
- starlark thread for running the starlark functions that are assertions:
  - should we make a new thread?
    - if yes, what should the name be, and should there be any imports/packages included?
      - Yes, the name should be related to the process that is being completed. We are leaving the option to include imports and globals.
    - if no, which thread should be included otherwise?
- if an assertion returns a falsy value (0, "", [], None) we should fail?
  - A: non-None, falsy returns from an assertion will be considered a violation.

### DIFF
--- a/pkg/cmd/template/cmd.go
+++ b/pkg/cmd/template/cmd.go
@@ -150,6 +150,11 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
 		return Output{Err: err}
 	}
 
+	err = rootLibraryExecution.ValidateValues(values)
+	if err != nil {
+		return Output{Err: err}
+	}
+
 	libraryValues = append(libraryValues, libraryValuesOverlays...)
 
 	if o.DataValuesFlags.Inspect {
--- a/pkg/cmd/template/cmd.go
+++ b/pkg/cmd/template/cmd.go
@@ -150,11 +150,6 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
 		return Output{Err: err}
 	}
 
-	err = rootLibraryExecution.ValidateValues(values)
-	if err != nil {
-		return Output{Err: err}
-	}
-
 	libraryValues = append(libraryValues, libraryValuesOverlays...)
 
 	if o.DataValuesFlags.Inspect {
--- a/pkg/cmd/template/cmd_assert_validate_test.go
+++ b/pkg/cmd/template/cmd_assert_validate_test.go
@@ -66,7 +66,7 @@ my_map:
 		assertSucceedsDocSet(t, filesToProcess, expected, opts)
 	})
 
-	t.Run("fails with only data values files", func(t *testing.T) {
+	t.Run("assert fails with only data values files", func(t *testing.T) {
 		t.Run("on multiple data values", func(t *testing.T) {
 			opts := cmdtpl.NewOptions()
 			dataValuesYAML := `#@ load("@ytt:assert", "assert")
--- a/pkg/cmd/template/cmd_assert_validate_test.go
+++ b/pkg/cmd/template/cmd_assert_validate_test.go
@@ -1,0 +1,274 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package template_test
+
+import (
+	cmdtpl "github.com/vmware-tanzu/carvel-ytt/pkg/cmd/template"
+	"github.com/vmware-tanzu/carvel-ytt/pkg/files"
+
+	"testing"
+)
+
+func TestAssertValidateOnDataValues(t *testing.T) {
+	t.Run("succeeds with data values files using --data-values-inspect", func(t *testing.T) {
+		opts := cmdtpl.NewOptions()
+		opts.DataValuesFlags.Inspect = true
+		dataValuesYAML := `#@ load("@ytt:assert", "assert")
+
+#@data/values
+#@assert/validate ("a non empty data values", lambda v: True if v else assert.fail("data values was empty"))
+---
+#@assert/validate ("a map with more than 3 elements", lambda v: True if len(v) > 3 else assert.fail("length of map was less than 3"))
+my_map:
+  #@assert/validate ("a non-empty string", lambda v: True if len(v) > 0 else assert.fail("length of string was 0"))
+  string: server.example.com
+  #@assert/validate ("an int over 9000", lambda v: True if v > 9000 else assert.fail("int was less than 9000"))
+  int: 54321
+  #@assert/validate ("a float less than pi", lambda v: True if v < 3.1415 else assert.fail("float was more than 3.1415"))
+  float: 2.3
+  #@assert/validate ("bool evaluating to true", lambda v:  v)
+  bool: true
+  #@assert/validate ("a null value", lambda v: True if v == None else assert.fail("value was not null"))
+  nil: null
+  #@assert/validate ("an array with less than or exactly 10 items", lambda v: True if len(v) <= 10 else assert.fail("array was more than 10 items"))
+  my_array:
+  #@assert/validate ("a non-empty string", lambda v: True if len(v) > 0 else assert.fail("length of string was 0"))
+  - abc
+  #@assert/validate ("an int over 9000", lambda v: True if v > 9000 else assert.fail("int was less than 9000"))
+  - 12345
+  #@assert/validate ("a float less than pi", lambda v: True if v < 3.1415 else assert.fail("float was more than 3.1415"))
+  - 3.14
+  #@assert/validate ("bool evaluating to true", lambda v:  v)
+  - true
+  #@assert/validate ("a null value", lambda v: True if v == None else assert.fail("value was not null"))
+  - null
+`
+
+		expected := `my_map:
+  string: server.example.com
+  int: 54321
+  float: 2.3
+  bool: true
+  nil: null
+  my_array:
+  - abc
+  - 12345
+  - 3.14
+  - true
+  - null
+`
+
+		filesToProcess := files.NewSortedFiles([]*files.File{
+			files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(dataValuesYAML))),
+		})
+
+		assertSucceedsDocSet(t, filesToProcess, expected, opts)
+	})
+
+	t.Run("fails with only data values files", func(t *testing.T) {
+		t.Run("on multiple data values", func(t *testing.T) {
+			opts := cmdtpl.NewOptions()
+			dataValuesYAML := `#@ load("@ytt:assert", "assert")
+
+#@data/values
+#@assert/validate ("a non empty data values", lambda v: True if v else assert.fail("data values was empty"))
+---
+#@assert/validate ("a map with less than 3 elements", lambda v: True if len(v) < 3 else assert.fail("length of map was more than or equal to 3"))
+my_map:
+  #@assert/validate ("a non-empty string", lambda v: True if len(v) > 0 else assert.fail("length of string was 0"))
+  string: ""
+  #@assert/validate ("an int over 9000", lambda v: True if v > 9000 else assert.fail("int was less than 9000"))
+  int: 2
+  #@assert/validate ("a float less than pi", lambda v: True if v < 3.1415 else assert.fail("float was more than 3.1415"))
+  float: 20.3
+  #@assert/validate ("bool evaluating to true", lambda v:  v)
+  bool: false
+  #@assert/validate ("a null value", lambda v: True if v == None else assert.fail("value was not null"))
+  nil: anything else
+  #@assert/validate ("an array with more than or exactly 10 items", lambda v: True if len(v) >= 10 else assert.fail("array was less than 10 items"))
+  my_array:
+  #@assert/validate ("a non-empty string", lambda v: True if len(v) > 0 else assert.fail("length of string was 0"))
+  - ""
+  #@assert/validate ("an int over 9000", lambda v: True if v > 9000 else assert.fail("int was less than 9000"))
+  - 123
+  #@assert/validate ("a float less than pi", lambda v: True if v < 3.1415 else assert.fail("float was more than 3.1415"))
+  - 3.14159
+  #@assert/validate ("bool evaluating to true", lambda v:  v)
+  - false
+  #@assert/validate ("a null value", lambda v: True if v == None else assert.fail("value was not null"))
+  - not null
+`
+
+			expectedErr := `One or more data values were invalid
+schema.yml:7 requires a valid value: "a map with less than 3 elements"; assert.fail: fail: length of map was more than or equal to 3 (by schema.yml:6)
+schema.yml:9 requires a valid value: "a non-empty string"; assert.fail: fail: length of string was 0 (by schema.yml:8)
+schema.yml:11 requires a valid value: "an int over 9000"; assert.fail: fail: int was less than 9000 (by schema.yml:10)
+schema.yml:13 requires a valid value: "a float less than pi"; assert.fail: fail: float was more than 3.1415 (by schema.yml:12)
+schema.yml:15 requires a valid value: "bool evaluating to true" (by schema.yml:14)
+schema.yml:17 requires a valid value: "a null value"; assert.fail: fail: value was not null (by schema.yml:16)
+schema.yml:19 requires a valid value: "an array with more than or exactly 10 items"; assert.fail: fail: array was less than 10 items (by schema.yml:18)
+schema.yml:21 requires a valid value: "a non-empty string"; assert.fail: fail: length of string was 0 (by schema.yml:20)
+schema.yml:23 requires a valid value: "an int over 9000"; assert.fail: fail: int was less than 9000 (by schema.yml:22)
+schema.yml:25 requires a valid value: "a float less than pi"; assert.fail: fail: float was more than 3.1415 (by schema.yml:24)
+schema.yml:27 requires a valid value: "bool evaluating to true" (by schema.yml:26)
+schema.yml:29 requires a valid value: "a null value"; assert.fail: fail: value was not null (by schema.yml:28)
+`
+
+			filesToProcess := files.NewSortedFiles([]*files.File{
+				files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(dataValuesYAML))),
+			})
+
+			assertFails(t, filesToProcess, expectedErr, opts)
+		})
+		t.Run("on a document", func(t *testing.T) {
+			opts := cmdtpl.NewOptions()
+			dataValuesYAML := `#@ load("@ytt:assert", "assert")
+
+#@data/values
+#@assert/validate ("a non empty data values", lambda v: True if v else assert.fail("data values was empty"))
+---
+`
+
+			expectedErr := `One or more data values were invalid
+schema.yml:5 requires a valid value: "a non empty data values"; assert.fail: fail: data values was empty (by schema.yml:4)
+`
+
+			filesToProcess := files.NewSortedFiles([]*files.File{
+				files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(dataValuesYAML))),
+			})
+
+			assertFails(t, filesToProcess, expectedErr, opts)
+		})
+	})
+
+	t.Run("fails when @assert/validate is not well-formed", func(t *testing.T) {
+		t.Run("is empty", func(t *testing.T) {
+			opts := cmdtpl.NewOptions()
+			dataValuesYAML := `#@data/values
+---
+#@assert/validate 
+foo: bar
+`
+
+			expectedErr := `Expected @assert/validate to have validation 2-tuple as argument(s), but found no arguments (by schema.yml:3)`
+
+			filesToProcess := files.NewSortedFiles([]*files.File{
+				files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(dataValuesYAML))),
+			})
+
+			assertFails(t, filesToProcess, expectedErr, opts)
+		})
+		t.Run("is not a tuple", func(t *testing.T) {
+			opts := cmdtpl.NewOptions()
+			dataValuesYAML := `#@data/values
+---
+#@assert/validate "string"
+foo: bar
+`
+
+			expectedErr := `Expected @assert/validate to have validation 2-tuple as argument(s), but found: "string" (by schema.yml:3)`
+
+			filesToProcess := files.NewSortedFiles([]*files.File{
+				files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(dataValuesYAML))),
+			})
+
+			assertFails(t, filesToProcess, expectedErr, opts)
+		})
+		t.Run("is an empty tuple", func(t *testing.T) {
+			opts := cmdtpl.NewOptions()
+			dataValuesYAML := `#@data/values
+---
+#@assert/validate ()
+foo: bar
+`
+
+			expectedErr := `Expected @assert/validate 2-tuple, but found tuple with length 0 (by schema.yml:3)`
+
+			filesToProcess := files.NewSortedFiles([]*files.File{
+				files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(dataValuesYAML))),
+			})
+
+			assertFails(t, filesToProcess, expectedErr, opts)
+		})
+		t.Run("has incorrect number of args in a validation", func(t *testing.T) {
+			opts := cmdtpl.NewOptions()
+			dataValuesYAML := `#@data/values
+---
+#@assert/validate (lambda v: v,)
+foo: bar
+`
+
+			expectedErr := `Expected @assert/validate 2-tuple, but found tuple with length 1 (by schema.yml:3)`
+
+			filesToProcess := files.NewSortedFiles([]*files.File{
+				files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(dataValuesYAML))),
+			})
+
+			assertFails(t, filesToProcess, expectedErr, opts)
+
+			dataValuesYAML2 := `#@data/values
+---
+#@assert/validate ("some validation", lambda v: v, "a message if valid", "a message if fail")
+foo: bar
+`
+			expectedErr = `Expected @assert/validate 2-tuple, but found tuple with length 4 (by schema.yml:3)`
+
+			filesToProcess = files.NewSortedFiles([]*files.File{
+				files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(dataValuesYAML2))),
+			})
+
+			assertFails(t, filesToProcess, expectedErr, opts)
+
+		})
+		t.Run("is a key=value pair", func(t *testing.T) {
+			opts := cmdtpl.NewOptions()
+			dataValuesYAML := `#@data/values
+---
+#@assert/validate foo=True
+foo: bar
+`
+
+			expectedErr := `Expected @assert/validate to have validation 2-tuple as argument(s), but found keyword argument (by schema.yml:3)`
+
+			filesToProcess := files.NewSortedFiles([]*files.File{
+				files.MustNewFileFromSource(files.NewBytesSource("schema.yml", []byte(dataValuesYAML))),
+			})
+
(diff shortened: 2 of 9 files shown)


gcheadle-vmware:
Linter seems to be failing due to a `const` variable not being commented... this is not consistent with when I've added new `const` vars in the past. Not sure what is going on here. Otherwise, this PR is ready for review. @jtigger

pkg/cmd/template/cmd.go
```diff
@@ -150,6 +150,11 @@ func (o *Options) RunWithFiles(in Input, ui ui.UI) Output {
 		return Output{Err: err}
 	}
 
+	err = rootLibraryExecution.ValidateValues(values)
```
pivotaljohn: Genuine question:
- under what conditions would we want to allow invalid values to be returned from `LibraryExecution.Values()`?
- would we want to validate values multiple times?

Trying to figure out if `ValidateValues()` should be part of `Values()`... 🤔

pkg/workspace/assert.go
```diff
@@ -0,0 +1,167 @@
...
+const (
+	AnnotationAssertValidate template.AnnotationName = "assert/validate"
+)
+
+// ProcessAndRunAssertions takes a root Node and traverses the tree checking for assert annotations.
+// Assertions are processed and executed using the value of the annotated node as the parameter to the assertions.
+//
+// When the assertions have violations, the errors are collected and returned, prefixed by the errMsg context string.
+// Otherwise, returns nil.
+func ProcessAndRunAssertions(n yamlmeta.Node, errMsg string) error {
```
pivotaljohn: Passing in `errMsg` only to use it to prepend to the returned value has something like [Feature Envy](https://refactoring.guru/smells/feature-envy) to me...

It seems like `ProcessAndRunAssertions()` need not have this additional responsibility.

Irrationally, it makes invocations feel strange to me: why am I passing in a failure message?

Suggestion: at the calling site, if there's an error, then prepend this message _there_. 

Thoughts?

pkg/workspace/assert.go
```diff
@@ -0,0 +1,167 @@
...
+
+func newAssertChecker(threadName string) *assertChecker {
+	return &assertChecker{thread: &starlark.Thread{Name: threadName}, violations: []error{}}
+}
+
+func (a *assertChecker) hasViolations() bool {
+	if len(a.violations) > 0 {
+		return true
+	}
+	return false
```
pivotaljohn:
```suggestion
	return len(a.violations) > 0
```

Simpler.

pkg/workspace/assert.go
```diff
@@ -0,0 +1,167 @@
...
+
+	return nil
+}
+
+func newValidationsFromAssertValidateAnnotation(annotation template.NodeAnnotation, n yamlmeta.Node, thread *starlark.Thread) ([]Validation, error) {
+	var validations []Validation
+	validationPosition := createAssertAnnotationPosition(annotation.Position, n)
+
+	if len(annotation.Kwargs) != 0 {
+		return nil, fmt.Errorf("Expected @%v to have validation 2-tuple as argument(s), but found keyword argument (by %v)", AnnotationAssertValidate, validationPosition.AsCompactString())
```
pivotaljohn: nit.

`%v` is a more general verb than `%s`... it added just a smidge of "cognitive load" for me to realize that you really do just want the string value given.

Use `%s` here if you really mean "insert the string value here"? 🤔

pkg/workspace/assert.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// Validate runs the function from the Validation with the node's value as arguments.
+//
+// Returns an error if the function returns an error, or if the resulting value from the function is falsy.
+// Otherwise, returns nil.
+func (v Validation) Validate(node yamlmeta.Node) error {
+	var nodeValue starlark.Value
+	switch values := node.GetValues(); len(values) {
+	case 0: // run validation with no arguments
```
pivotaljohn: Are there cases when we expect there to be no values?

If not, this might be cause for PANIC!!!!!! 😱

pkg/workspace/assert.go
```diff
@@ -0,0 +1,167 @@
...
+	case 1:
+		nodeValue = yamltemplate.NewGoValueWithYAML(values[0]).AsStarlarkValue()
+	default:
+		nodeValue = yamltemplate.NewGoValueWithYAML(values).AsStarlarkValue()
```
pivotaljohn: I don't expect that we've necessarily written test to cover this... but what are we envision is happening here? 

That is: under what conditions would this branch be taken?  And how does such input arrive at the validators?

p.s. If we don't have answers to those questions right now, I think that's actually okay... but in that case, I think `panic()`'ing would be better — it would prevent the possibility that users encounter a behavior that we wouldn't want them to rely on... panic'ing reserves that branch for us to do something better informed, later...

pkg/workspace/assert.go
```diff
@@ -0,0 +1,167 @@
...
+		})
+	}
+
+	return validations, nil
+}
+
+// Validation is an assertion from an @assert/validate annotation.
+// One @assert/validate annotation can have multiple Validations.
+type Validation struct {
+	thread   *starlark.Thread
```
pivotaljohn: I guess time (and research) will tell whether we want a thread per Validation... and whether it should be state _on_ that validation or be something constructed each time the validation is run... or what. 🤔 

Do you have any insights right now on thread handling?

pkg/workspace/assert.go
```diff
@@ -0,0 +1,167 @@
...
+	msg      string
+	f        starlark.Callable
```
pivotaljohn: `f` feels a bit terse to me, here.

There's some distance between the declaration of the variable from its usage.

We could call this `assertion`?

pkg/workspace/assert.go
```diff
@@ -0,0 +1,167 @@
...
+			position: validationPosition,
+		})
+	}
+
+	return validations, nil
+}
+
+// Validation is an assertion from an @assert/validate annotation.
+// One @assert/validate annotation can have multiple Validations.
+type Validation struct {
```
pivotaljohn: I think what we have here is a `Rule`???

From: https://github.com/vmware-tanzu/carvel/pull/331/files#diff-a81faf6d33e82e22eee1083ee484a23eda8a928b046116a2cb516e58086984b8R48-R53

> - **rule** — the combination of: 
>     - a textual description of what constitutes a valid value;
>     - a function that asserts the rule against an actual value;
>     - a violation message template, used to report invalid values.
>     - **custom rule** — a rule based on a function (either one written by a user or provided in a `ytt` module).
>     - **named rule** — a rule provided with `ytt` expressed through a keyword argument to an `@schema/validation` or `@assert/validate`.

vs.

> - **validation** — the binding of one or more "rule"s to a "node".

ℹ️  what words we use is absolutely up for debate... 👍🏻   What I'm pushing for is [ubiquitous language](https://thedomaindrivendesign.io/developing-the-ubiquitous-language/).

pkg/workspace/assert.go
```diff
@@ -0,0 +1,167 @@
...
+		nodeValue = yamltemplate.NewGoValueWithYAML(values).AsStarlarkValue()
+	}
+
+	result, err := starlark.Call(v.thread, v.f, starlark.Tuple{nodeValue}, []starlark.Tuple{})
+	if err != nil {
+		return fmt.Errorf("%v requires a valid value: %v; %v (by %v)", node.GetPosition().AsCompactString(), v.msg, err.Error(), v.position.AsCompactString())
+	}
+	// From proposal: "if all rules pass (either returns True or None), then the value is valid."
+	// I believe a None result would return an error
+	if !result.Truth() {
```
pivotaljohn: 🤔 

I really dig the simplicity of "Falsey is faliure, Truthy is success."

_However_, it's really common for assertion functions to actually return nil/null/None upon success.

For example:

https://github.com/vmware-tanzu/carvel-ytt/blob/b50fd584133ce57a147c9fa050bd4acea684a259/pkg/yttlibrary/assert.go#L75-L86

pivotaljohn (changes requested):
This is a partial review; I was unable to digest the annotation position calculation stuff in time.

But I wanted to provided feedback ASAP.

I'll come back with another review on that other part. In the meantime, some stuff for us to chew on...

pkg/cmd/template/cmd.go
gcheadle-vmware: After taking a step back and reevaluating the change set, I think that it would make sense for `Values()` to return a valid set. I think I made `ValidateValues()` separate at the beginning of the implementation to get some distance from the Data Values Pre-Processing that we know and love.

pkg/workspace/assert.go
gcheadle-vmware: Here I noticed that `ProcessAndRunAssertions()` would likely in the future be called with more than just data values, so there was an effort to make that error msg composable when assertions fail. I agree that we should be able to compose this err message at the call site 👍

pkg/workspace/assert.go
gcheadle-vmware: `node.GetValues()` couldn't return an empty list of array or map items whose length could be zero because that would mean the map or array was annotated with `@assert/validate`, which you can not do. I think a change could be made so that `@assert/validate` errors/panics if we find it on a map or array. With a guard in place, I think it would be right to panic for cases other than `len(values)==1`.

pkg/workspace/assert.go
gcheadle-vmware: see comment above, tldr: I agree.

pkg/workspace/assert.go
gcheadle-vmware: The insights that I gained from looking at other thread creations in the code suggested that the thread be created higher up in the code path (currently created in ProcessAndRunAssertions). The thread is currently injected all the way down to a Validation, however I also experimented with it as a parameter to the Validate function: `Validation.Validate(Node, starlark.Thread)`. The thread would then be supplied to the validate function via the `AssertChecker`. Thoughts?

pkg/workspace/assert.go
gcheadle-vmware: Thank you for this 🙌  I didn't catch that I had some of these terms incorrect/swapped.

pkg/workspace/assert.go
gcheadle-vmware: Great point, I'm not sure where I got the Falsey as failure for assertions from, after reviewing the proposal, I see here what the desired behavior is:
[from proposal](https://github.com/vmware-tanzu/carvel/blob/aa19de3af4626f70765b8543bea60ca4ff5ca9c2/proposals/ytt/004-schema-validation/README.md#assertvalidate):
- if all rules pass (either returns True or None), then the value is valid.
- if a rule returns False (not None) or assert.fail()s, then the valid is invalid.

pkg/cmd/template/cmd.go
pivotaljohn: Makes sense to me... and reads like a great way to start in with stuff... I think it's generally easier to integrate in than extract out... so. 👍🏻

pkg/workspace/assert.go
pivotaljohn: mmm... yeah... and that might result in duplicate code... it's worth waiting to see how that looks.  Sometimes, a little bit of duplication is easier to maintain.

pkg/workspace/assert.go
pivotaljohn: > With a guard in place, I think it would be right to panic for cases other than len(values)==1.

My intuition says this is a good starting place.

pkg/workspace/assert.go
pivotaljohn: Yup yup...

and a fine-tuning:

> if a rule returns Falsey (except None) or assert.fail()s, then the valid is invalid.

pkg/workspace/assert.go
pivotaljohn: Prior art suggests to me that a new thread is needed for each evaluation:
https://github.com/vmware-tanzu/carvel-ytt/blob/b50fd584133ce57a147c9fa050bd4acea684a259/pkg/workspace/template_loader.go#L197

So, from what I can tell, the more relevant part of Thread is its ability to carry context along the flow-of-execution (via `SetLocal()`/`Local()`).

Would we ever want to _share_ context in this way across assertion calls (or from a higher level process to assertion calls...)?
Conversely, do we want to specifically _prevent_ sharing context in this way across assertion calls?

🤔

pkg/filepos/position.go
```diff
@@ -122,3 +123,32 @@ func (p *Position) IsNextTo(otherPostion *Position) bool {
...
+
+	return annPos
+}
+
+// Meta is a representation of comment's source, contains comment as a string, and original position.
+// Used to populate
+type Meta struct {
+	Data     string
+	Position *Position
+}
```
pivotaljohn, line 154: In my head, the responsibility of the `filepos` package is to provide the domain concept of "Position" in `ytt`.

A "position" in `ytt` uniquely identifies a single line of text. It also provides the means of presenting that "coordinate" to the end-user in various formats.
More recently, it's also provided a means of rendez-vous'ing back with the literal source line (currently, caching it).

Currently, this package is utterly unaware of how this concept is used.

This change is about an algorithm _about_ position, but I'm dubious as to whether it belongs _in_ `filepos`.

With _this_ change, we're expanding the conceptual scope of `filepos` to include ideas like "Node" and "Comment" and "Meta". Those concepts _have_ homes already.

Is there an even better home for this logic?

pivotaljohn: On second thought... this shaped code reads like a kind of "repair" — it's populating a value that "should have" been there from the start...??

... are we discovering that _this_ feature requires more of our YAML parser?

Namely, that _it_ should be populating _even annotations_ with position?

pivotaljohn:
> Linter seems to be failing due to a `const` variable not being commented... this is not consistent with when I've added new `const` vars in the past. Not sure what is going on here. Otherwise, this PR is ready for review. @jtigger

It's my recollection that all exported identifiers (functions, types, variables ... including constants) "ought" to have GoDoc.  In a case like this, this string's role may seem blatantly obvious, but I think that just means such documentation is for the very uninitiated.

🤷🏻

pkg/workspace/assert.go
gcheadle-vmware: Since I cannot produce a reason to either explicitly include or exclude context from the TemplateLoader, I think it would be best to include the Loader when creating a new thread, so that we have the option later to include or not include the context that the Loader contains.

pkg/filepos/position.go
gcheadle-vmware: Pulling this (annotation positions) work into another PR.

pkg/cmd/template/cmd_assert_validate_test.go
```diff
@@ -0,0 +1,273 @@
...
+		})
+		t.Run("has incorrect function arg in a validation", func(t *testing.T) {
+			opts := cmdtpl.NewOptions()
+			dataValuesYAML := `#@data/values
+---
+#@assert/validate ("some validation", True)
+foo: bar
+`
+
+			expectedErr := `Invalid @assert/validate annotation - expected @assert/validate to have an assertion function as the second item in the 2-tuple, but found type: bool (by schema.yml:3)`
```
pivotaljohn: Love these error messages. ❤️ 

In fact, many folks _are going to_ fumble their way through this syntax... these guards are going to be well-used, I predict. 💪🏻 

nit: suggesting a tweak.

```suggestion
			expectedErr := `Invalid @assert/validate annotation - expected second item in the 2-tuple to be an assertion function, but was a bool (at schema.yml:3)`
```

pkg/filepos/position.go
pivotaljohn: 👍🏻

pkg/workspace/assert.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+type assertChecker struct {
+	thread     *starlark.Thread
+	violations []error
+}
+
+func newAssertChecker(threadName string, _ *TemplateLoader) *assertChecker {
+	return &assertChecker{thread: &starlark.Thread{Name: threadName}, violations: []error{}}
+}
```
pivotaljohn: If I was tracking correctly, we talked about the `TemplateLoader` param being forward-looking... that we'll likely need it at some point?

I think we should defer anything like this until we actually see the need arise.

I believe that there's even reason to think that it may not be needed for invoking a function: Starlark function values contain (all?) of their execution context and likely won't be loading other templates.

```suggestion
func newAssertChecker(threadName string) *assertChecker {
	return &assertChecker{thread: &starlark.Thread{Name: threadName}, violations: []error{}}
}
```

pkg/workspace/assert.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil
+	}
+
+	err := yamlmeta.Walk(n, checker)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
```
pivotaljohn: The function requires the caller to construct the checker.

In a previous revision, it _took on_ that responsibility... what motivated relocating that responsibility to the caller(s)?

I think passing in the name of the scenario/task (which we set as the name of the thread) — in this case "assert-data-values" — seems to fit, given that we want to run validations in a number of places. 👍🏻 

When I naively think of what I'd expect the signature of this function to be, I assume:
- the Node(tree)
- some configuration about the validation process itself (so far, we've identified the need to give it a name... in the future, we might provide _other_ options. 🤷🏻 )
- and it would return the slice of violations.

Perhaps we're want for a value object that holds the violations that has that `HasViolations()` method on it? Perhaps we extract that out of the `assertChecker` — it would then _produce_ such a value rather than _be_ that value... 🤔

pkg/workspace/assert.go
```diff
@@ -0,0 +1,162 @@
...
+		values := node.GetValues()
+		nodeValue = yamltemplate.NewGoValueWithYAML(values[0]).AsStarlarkValue()
+	}
+
+	result, err := starlark.Call(thread, r.assertion, starlark.Tuple{nodeValue}, []starlark.Tuple{})
+	if err != nil {
+		return fmt.Errorf("%s requires %s; %s (by %s)", node.GetPosition().AsCompactString(), r.msg, err.Error(), r.position.AsCompactString())
+	}
+
+	// in order to pass, the assertion must return True or None
```
pivotaljohn, line 172: nice comment. 👌🏻

pkg/workspace/assert.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// Visit if `node` is annotated with `@assert/validate` (AnnotationAssertValidate).
+// Checks, validates, and runs the validation Rules, any violations from running the assertions are collected.
+//
+// This visitor returns and error if any assertion is not well-formed,
+// otherwise, returns nil.
+func (a *assertChecker) Visit(node yamlmeta.Node) error {
+	nodeAnnotations := template.NewAnnotations(node)
+	if nodeAnnotations.Has(AnnotationAssertValidate) {
```
pivotaljohn: Consider: negating this if and turning it into a guard clause. It would reduce nesting and make the function a tad more readable.

pkg/workspace/assert.go
```diff
@@ -0,0 +1,162 @@
...
+		switch node.(type) {
+		case *yamlmeta.DocumentSet, *yamlmeta.Array, *yamlmeta.Map:
+			return fmt.Errorf("Invalid @%s annotation - not supported on %s at %s", AnnotationAssertValidate, yamlmeta.TypeName(node), node.GetPosition().AsCompactString())
+		default:
+			rules, syntaxErr := newRulesFromAssertValidateAnnotation(nodeAnnotations[AnnotationAssertValidate], node)
```
pivotaljohn: I dig what you did there: conveying that the error here would be a syntax problem by naming the `err` to `syntaxErr`.

It broke up the monotony for me of `if err != nil {...}`.

Got me thinking about the pros and cons. 🤔   Great experimentation/creativity, here. 💡

pkg/workspace/library_execution.go
```diff
@@ -86,7 +87,34 @@ func (ll *LibraryExecution) Values(valuesOverlays []*datavalues.Envelope, schema
...
+	}
+
+	return values, libValues, ll.validateValues(values, loader)
+}
+
+// validateValues runs validations from @assert/validate annotations in Data Values for the current library.
+//
+// Returns an error if the arguments to an @assert/validate are invalid,
+// otherwise, checks the assertionChecker for violations, and returns nil if there are no violations.
+func (ll *LibraryExecution) validateValues(values *datavalues.Envelope, loader *TemplateLoader) error {
```
pivotaljohn: Keep an eye open for the possibility that we'll want to wrap the validation process in a struct of its own (like we do with `DataValuesSchemaPreProcessing`, `DataValuesPreProcessing`, and `OverlayPostProcessing`). 

Among other benefits is that it provides an easy home for helper functions and a handy place to put state/inject dependencies.

Could be a bit early, but wanted to note that existing pattern.

pkg/cmd/template/cmd_assert_validate_test.go
```diff
@@ -0,0 +1,273 @@
...
+  #@assert/validate ("a float less than pi", lambda v: True if v < 3.1415 else assert.fail("float was more than 3.1415"))
+  - 3.14159
+  #@assert/validate ("bool evaluating to true", lambda v:  v)
+  - false
+  #@assert/validate ("a null value", lambda v: True if v == None else assert.fail("value was not null"))
+  - not null
+`
+
+			expectedErr := `One or more data values were invalid:
+- schema.yml:6 requires "a map with less than 3 elements"; assert.fail: fail: length of map was more than or equal to 3 (by schema.yml:5)
```
pivotaljohn:
```suggestion
- my_map (at schema.yml:6) requires "a map with less than 3 elements"; assert.fail: fail: length of map was more than or equal to 3 (by schema.yml:5)
```

(see suggestion at `Rule.Validate()` for a possible approach)

pkg/workspace/assert.go
```diff
@@ -0,0 +1,162 @@
...
+		panic(fmt.Sprintf("@%s annotation at %s - not supported on %s at %s", AnnotationAssertValidate, r.position.AsCompactString(), yamlmeta.TypeName(node), node.GetPosition().AsCompactString()))
+	default:
+		values := node.GetValues()
+		nodeValue = yamltemplate.NewGoValueWithYAML(values[0]).AsStarlarkValue()
+	}
+
+	result, err := starlark.Call(thread, r.assertion, starlark.Tuple{nodeValue}, []starlark.Tuple{})
+	if err != nil {
+		return fmt.Errorf("%s requires %s; %s (by %s)", node.GetPosition().AsCompactString(), r.msg, err.Error(), r.position.AsCompactString())
+	}
```
pivotaljohn: Because we want to provided a user-friendly identifier/key of the offending value, I think we _need_ to handle all the Node types.

Here's one approach:

```suggestion
func (r Rule) Validate(node yamlmeta.Node, thread *starlark.Thread) error {
	var key string
	var nodeValue starlark.Value
	switch typedNode := node.(type) {
	case *yamlmeta.DocumentSet, *yamlmeta.Array, *yamlmeta.Map:
		panic(fmt.Sprintf("@%s annotation at %s - not supported on %s at %s", AnnotationAssertValidate, r.position.AsCompactString(), yamlmeta.TypeName(node), node.GetPosition().AsCompactString()))
	case *yamlmeta.MapItem:
		key = fmt.Sprintf("%s", typedNode.Key)
		nodeValue = yamltemplate.NewGoValueWithYAML(typedNode.Value).AsStarlarkValue()
	case *yamlmeta.ArrayItem:
		key = "array item"
		nodeValue = yamltemplate.NewGoValueWithYAML(typedNode.Value).AsStarlarkValue()
	case *yamlmeta.Document:
		key = "document"
		nodeValue = yamltemplate.NewGoValueWithYAML(typedNode.Value).AsStarlarkValue()
	}

	result, err := starlark.Call(thread, r.assertion, starlark.Tuple{nodeValue}, []starlark.Tuple{})
	if err != nil {
		return fmt.Errorf("%s (%s) requires %s; %s (by %s)", key, node.GetPosition().AsCompactString(), r.msg, err.Error(), r.position.AsCompactString())
	}

```

pivotaljohn (changes requested):

pkg/workspace/assert.go
gcheadle-vmware: What motivated this change was that I wanted `ProcessAndRunValidations()` to be able to accept different kinds of checkers. Right now there is just one Validation annotation, but I was thinking that if we needed to process a subset of Validation annotations, then accepting a Visitor interface would be the way to do it.

gcheadle-vmware: Seems like I may be assuming too much about the future in this case, I agree that the responsibility should be with `ProcessAndRunValidations()` to construct the checker.

pkg/workspace/assert.go
```diff
@@ -22,30 +22,50 @@ const (
...
+// HasViolations indicates whether this AssertCheck contains any violations.
+func (ac *AssertCheck) HasViolations() bool {
+	return len(ac.Violations) > 0
 }
 
-func newAssertChecker(threadName string, _ *TemplateLoader) *assertChecker {
-	return &assertChecker{thread: &starlark.Thread{Name: threadName}, violations: []error{}}
+type assertChecker struct {
+	thread *starlark.Thread
+	AssertCheck
```
pivotaljohn: Is this exported? What's our intent?

pivotaljohn (approved):
Reviewed again, synchronously.

With the private library test and the tweaks discussed, LGTM!

pkg/cmd/template/cmd_assert_validate_test.go
```diff
@@ -0,0 +1,273 @@
...
+package template_test
+
+import (
+	cmdtpl "github.com/vmware-tanzu/carvel-ytt/pkg/cmd/template"
+	"github.com/vmware-tanzu/carvel-ytt/pkg/files"
+
+	"testing"
+)
+
+func TestAssertValidateOnDataValues(t *testing.T) {
```
cari-lynn: Subtests can
- Make tests easier to read by reading the top level test and the sub-test name as one sentence.
-  Separate and group success and failure tests, and say why they succeed or fail in the name.
An idea I am not married to is something like:
```
func TestAssertValidateOnDataValuesSucceeds(t *testing.T) {
  t.Run("when all assertions pass", func(t *testing.T) {...}
}
func TestAssertValidateOnDataValuesFails(t *testing.T) {
  t.Run("when assertions fail", func(t *testing.T) {...}
  t.Run("when assertions fail on the document", func(t *testing.T) {...}
}
```

The content of the tests look great to me.

pkg/workspace/assert.go
```diff
@@ -0,0 +1,190 @@
...
+// Declare @assert/... annotation names
+const (
+	AnnotationAssertValidate template.AnnotationName = "assert/validate"
+)
+
+// ProcessAndRunValidations takes a root Node, and threadName, and traverses the tree checking for assert annotations.
+// Validations are processed and executed using the value of the annotated node as the parameter to the assertions.
+//
+// When the assertions have violations, the errors are collected and returned in an AssertCheck.
+// Otherwise, returns empty AssertCheck and nil.
```
cari-lynn, line 24: Love these detailed godocs ⭐

pkg/filepos/position.go
```diff
@@ -122,3 +123,32 @@ func (p *Position) IsNextTo(otherPostion *Position) bool {
...
+	for _, c := range nodeComments {
+		if c.Position.IsKnown() && c.Position.AsIntString() == fmt.Sprintf("%d", annPos.LineNum()) {
+			lineString = fmt.Sprintf("%v#%s", strings.Repeat(" ", leftPadding), c.Data)
+		}
+	}
+
+	annPos.SetFile(nodePosition.GetFile())
+	annPos.SetLine(lineString)
+
+	return annPos
```
cari-lynn, line 146: [Thinking out loud here] I wonder if these changes to add annotation positions would benefit from or are risky enough for a test. I know we have some file position tests in `template_test.go`.

cari-lynn (commented):
I'm really happy with this change set and the code that I see is clean and dry.

I would like to see a test for `_ytt_lib` libraries if that is part of this work.

pkg/cmd/template/cmd_assert_validate_test.go
gcheadle-vmware: Did some reformatting 👍

PR #631, opened 2022-03-17
plumb annotation positions through compilation 

gcheadle-vmware:
- added annotations map to Nodes data structure 
  - used to retain position info through template compilation,
- set line of comment's position to source line rather than the comment's contents
- copy comment's position into Node Annotation
- removed repetitive `MetaOpts` and `MetasOpts`
- removed code that is now obsolete that was retroactively adding source lines to annotations


pkg/template/meta.go
```diff
@@ -17,15 +18,32 @@ const (
...
+}
+
+func (a *Annotations) AddAnnotation(tag NodeTag, ann Annotation) {
+	if _, found := a.tagToAnnotationsMap[tag]; !found {
+		a.tagToAnnotationsMap[tag] = NodeAnnotations{}
+	}
+	a.tagToAnnotationsMap[tag][ann.Name] = NodeAnnotation{Position: ann.Position}
+}
+
+func (a *Annotations) FindAnnotation(tag NodeTag, annName AnnotationName) (NodeAnnotation, bool) {
```
pivotaljohn: 🤔  At this point, the function no longer justifies the level of indirection.

Suggestion: inline it.

💭  although, if you merge `Annotations` and `Nodes` in the Evaluation Context (see below), then this would most definitely be needed.

pkg/template/evaluation_ctx.go
```diff
@@ -5,19 +5,17 @@ package template
...
 	// Should not import template specific packages here (like yamlmeta)
 )
 
 type EvaluationCtx struct {
 	dialect EvaluationCtxDialect
 
-	nodes     *Nodes
-	ancestors Ancestors
+	nodes       *Nodes
+	annotations *Annotations
```
pivotaljohn: On reflection, given that `nodes` and `annotations` have the same lifecycle and both are collections of source entities, we can reduce some complexity in merging these two.

That is: put the `Annotations` _inside_ of `Nodes`.

Something like:

```go
type Nodes struct {
	id               int
	tagToNode        map[NodeTag]EvaluationNode
	childToParentTag map[NodeTag]NodeTag
	annotations      map[NodeTag]map[AnnotationName]NodeAnnotation
}
```
(updating the `NewNodes()` constructor, of course.)

move over the methods:
- `Nodes.AddAnnotation(NodeTag, AnnotationName)`
- `Nodes.FindAnnotation(NodeTag, AnnotationName)`


Perhaps `Nodes` is still a good name? (stretching the definition of a node to include its annotations).

pivotaljohn (commented):
After sleeping on it, some thoughts.

pkg/template/meta.go
```diff
@@ -39,22 +36,26 @@ type MetaOpts struct {
 	IgnoreUnknown bool
 }
 
-// NewAnnotationFromString constructs an Annotation from a given string.
+// NewAnnotationFromComment constructs an Annotation from a given comment.
 //
 // if opts.IgnoreUnknown is true and the annotation is unknown, it is returned as a comment.
 // if opts.IgnoreUnknown is false and the annotation is unknown, returns an error.
-func NewAnnotationFromString(data string, opts MetaOpts) (Annotation, error) {
+func NewAnnotationFromComment(comment *yamlmeta.Comment, opts MetaOpts) (Annotation, error) {
```
gcheadle-vmware: Changing this param into a `yamlmeta.Comment` imports `yamlmeta` into the `template` package. We don't import it anywhere else in the `template` package, so maybe we include the position as an extra param instead?

pivotaljohn: Good catch. Yeah, that breaks our intended layering: `template` should be dialect agnostic. 👍🏻 

The breakdown you suggest breaks the explicit suggestion that `template.Annotation` gets its position from a `yamlmeta.Comment`. So, that's at least a better place to be.

pkg/yamltemplate/metas.go
```diff
@@ -36,16 +36,12 @@ type CommentAndAnnotation struct {
...
-}
-
 // NewTemplateAnnotationFromYAMLComment parses "comment" into template.Annotation.
 //
 // nodePos is the position of the node to which "comment" is attached (this is important to differentiate between
 // @template/code and @template/value).
-func NewTemplateAnnotationFromYAMLComment(comment *yamlmeta.Comment, nodePos *filepos.Position, opts MetasOpts) (template.Annotation, error) {
-	ann, err := template.NewAnnotationFromString(comment.Data, template.MetaOpts{IgnoreUnknown: opts.IgnoreUnknown})
+func NewTemplateAnnotationFromYAMLComment(comment *yamlmeta.Comment, nodePos *filepos.Position, opts template.MetaOpts) (template.Annotation, error) {
+	ann, err := template.NewAnnotationFromComment(comment.Data, comment.Position, opts) //template.MetaOpts{IgnoreUnknown: opts.IgnoreUnknown})
```
cppforlife: comment?

pkg/template/meta.go
```diff
@@ -39,22 +38,25 @@ type MetaOpts struct {
...
 }
 
-// NewAnnotationFromString constructs an Annotation from a given string.
+// NewAnnotationFromComment constructs an Annotation from a string and position from a Comment.
 //
 // if opts.IgnoreUnknown is true and the annotation is unknown, it is returned as a comment.
 // if opts.IgnoreUnknown is false and the annotation is unknown, returns an error.
-func NewAnnotationFromString(data string, opts MetaOpts) (Annotation, error) {
+func NewAnnotationFromComment(data string, position *filepos.Position, opts MetaOpts) (Annotation, error) {
+	annotationPosition := position.DeepCopy()
```
cppforlife: annotationPosition => position

pkg/template/nodes.go
```diff
@@ -50,6 +59,22 @@ func (n *Nodes) FindNode(tag NodeTag) (EvaluationNode, bool) {
...
 }
 
+// AddAnnotation creates an entry in the annotations map of Nodes.
+// The entry is NodeAnnotation with only the annotation's position,
+// indexed by NodeTag and AnnotationName.
+func (n *Nodes) AddAnnotation(tag NodeTag, ann Annotation) {
+	if _, found := n.annotations[tag]; !found {
+		n.annotations[tag] = NodeAnnotations{}
+	}
+	n.annotations[tag][ann.Name] = NodeAnnotation{Position: ann.Position}
```
cppforlife, line 69: im not sure i understand why we are using NodeAnnotation here. It seems like purpose of AddAnnotation is to store position mapped to a name of the annotation. shouldnt it be something like name => Position? so i would argue that it should be AddAnnotationPosition/FindAnnotationPosition.

additionally whats tripping me up is that annotation name is technically not a unique key, since user can specify multiple annotations on the same node -- should we be failing somewhere for that? may be we should be using index of the annotation for that node.

pkg/yamlmeta/parser.go
```diff
@@ -199,7 +199,7 @@ func (p *Parser) assignComments(val interface{}, comments []yaml.Comment, lineCo
 	for _, comment := range comments {
 		comment := &Comment{
 			Data:     comment.Data,
-			Position: p.newPosition(comment.Line, lineCorrection, comment.Data),
+			Position: p.newPosition(comment.Line, lineCorrection, lines[comment.Line]),
```
cppforlife, line 202: why do we need to be switching to lines. seems like it introduces more possibility that we ll screw something up.

pkg/yamltemplate/metas.go
```diff
@@ -36,16 +36,12 @@ type CommentAndAnnotation struct {
 	Annotation *template.Annotation
 }
 
-type MetasOpts struct {
```
cppforlife, line 39: for the record, the reason why i had MetasOps in this package is because yamltemplate may in theory have its own opts, and hence shouldnt have to expose underlying impl (template package's MetasOpts).

pkg/template/nodes.go
pivotaljohn: > im not sure i understand why we are using NodeAnnotation here. It seems like purpose of AddAnnotation is to store position mapped to a name of the annotation. shouldnt it be something like name => Position? so i would argue that it should be AddAnnotationPosition/FindAnnotationPosition.

Your read is spot on.

We've been thinking of it slightly broader: that there are two parts of annotation data:
- the bits that are evaluated (args and kwargs)
- the bits that are not (position)

The stuff in that second category would... categorically... be carried over. So, even though it's one field, today, that felt more like an implementation detail rather than a conceptual distinction.

But yes, YAGNI — the only field being carried over is a position.

So, our current thinking is that the risk of accidentally depending on some new field in the source annotation is outweighed by the design conveying that it's annotation data making it easier to grok what's going on, here.

How does that show up on _your_ scorecard?


> additionally whats tripping me up is that annotation name is technically not a unique key, since user can specify multiple annotations on the same node -- should we be failing somewhere for that? may be we should be using index of the annotation for that node

We didn't want to introduce a breaking change with this work. Today, [the last annotation in, wins](https://carvel.dev/ytt/#gist:https://gist.github.com/pivotaljohn/5c5b8d0ed1bb47e5cd5204b54514f8cc).

This probably goes on the ytt v1 wish list.

pkg/yamlmeta/parser.go
pivotaljohn: That field on `Position` is mean to be a cache of the source line.

What's the risk you see?  Note that all the other elements's positions cache the source line as well (we just missed one when introducing this cache).

pkg/yamltemplate/metas.go
pivotaljohn: Roger that.
And thank you for the context.

We removed it (for now) because:
- to date, there has been no drift, and;
- `yamltemplate` is a concretion of `template` — the dependency is natural.

It was an abstraction that wasn't (yet) carrying its weight.

pkg/template/meta.go
pivotaljohn: ~name's taken... I think you mean make it shorter (or shorten the parameter name as well).~

@gcheadle-vmware covered this: shadowing seems appropriate, here. 👍🏻

pkg/yamltemplate/metas.go
pivotaljohn: Thank you! (oops!)

PR #632, opened 2022-03-17
return correct error for go api example

AyWa:
Fix a small issue of the example. When I little play around with the example, I had an issue where the template output was empty without any errors. Because the wrong return value was use


pivotaljohn:
Oops! Nice catch, @AyWa.

We typically include running examples as part of our build; clearly missed a step.

---

Not sure why your agreement was rejected. Please let us know if we can help in any way.

AyWa:
@pivotaljohn I submitted again the cla. Lets see if it is approve this time 🤔

PR #638, opened 2022-03-22
Explore: ytt being able to emit comments

pivotaljohn:
This is a PR meant for exploration: what would be the consequences of ytt emitting comments?

Inspired by: https://hackmd.io/IqHKk9LgSkadtrhV0-qyjg


examples/concourse-overlay/expected.txt
```diff
@@ -1,10 +1,10 @@
 resources:
```
cppforlife, line 1: non-indented arrays is more or less a given with yaml formatting. would not be good to change that for our end users imho.

pivotaljohn: > non-indented arrays is more or less a given with yaml formatting

From what I've seen, there is no clear consensus on the indentation of sequences:
- from [the spec](https://yaml.org/spec/1.2-old/spec.html#:~:text=The%20amount%20of%20indentation%20is%20a%20presentation%20detail%20and%20must%20not%20be%20used%20to%20convey%20content%20information):
  >  "The amount of indentation is a [presentation detail](https://yaml.org/spec/1.2-old/spec.html#presentation/detail/) and must not be used to convey [content](https://yaml.org/spec/1.2-old/spec.html#content//) information.

  and [also notes](https://yaml.org/spec/1.2.2/#:~:text=Since%20people%20perceive%20the%20%E2%80%9C%2D%E2%80%9D%20indicator%20as%20indentation%2C%20nested%20block%20sequences%20may%20be%20indented%20by%20one%20less%20space%20to%20compensate%2C):

  > "Since people perceive the “-” indicator as [indentation](https://yaml.org/spec/1.2.2/#indentation-spaces), nested [block sequences](https://yaml.org/spec/1.2.2/#block-sequences) may be [indented](https://yaml.org/spec/1.2.2/#indentation-spaces) by one less [space](https://yaml.org/spec/1.2.2/#white-space-characters) to compensate,"
- certain communities express indentation in their guides:
  - https://developers.home-assistant.io/docs/documenting/yaml-style-guide/#block-style-sequences
  - https://github.com/flathub/flathub/wiki/YAML-Style-Guide
  - https://symfony.com/doc/current/components/yaml/yaml_format.html#:~:text=You%20can%20nest%20sequences%20and%20mappings%20as%20you%20like%3A
- [eslint](https://ota-meshi.github.io/eslint-plugin-yml/rules/indent.html) lints non-indented sequences as an error.
- the YAML library author [weighs in](https://github.com/go-yaml/yaml/issues/661#issuecomment-733677290) as well:
  > The formatting in v2 was inconsistent in several ways with many edge cases. In v3 we've changed it to simply make it not special: every depth level is indented by the space amount of spacing, regardless of its type.
  >
  > I understand that some people will be used to the old formatting, but on this basis alone it'd be hard to evolve anything.

---

> ... would not be good to change that for our end users imho...

I'm not sure.

On the one hand, the output of an upgraded version would not be _syntactically_ identical, but it would _semantically_; so it's arguable.
One path is to give the user the option: if you want to output comments, sequences are indented; otherwise no change.

On the other hand, folks might be diff'ing text to determine if there are changes.
I'm unsure how common this is.

I'm currently thinking that the added value that comments can bring would outweigh any inconvenience.

What informs _your_ humble opinion, here? What do you see?

pivotaljohn:
May depend on https://github.com/go-yaml/yaml/issues/749 getting addressed.

From [#carvel](https://kubernetes.slack.com/archives/CH8KCCKA5/p1648545789088679?thread_ts=1648544620.582159&cid=CH8KCCKA5)

> Yeah I think it is just that the first sperator gets removed as you can see from the example (line 2 of app.yml).  I was hoping there was a way to allow this as yamllint we have checks if the --- is at the start (edited)

pivotaljohn:
Keeping the branch, but closing this PR, for now.

tvkit:
Allowing _eslint_ directives would be nice:
`# eslint yml/sort-keys: warn`

PR #639, opened 2022-03-23
remove cobra dependency from pkg/cmd/template package

cppforlife:
ytt library users do not want to bring in unnecessary dependencies (such as cobra). 

per request from #carvel slack: https://kubernetes.slack.com/archives/CH8KCCKA5/p1647894140376479


examples/integrating-with-ytt/internal-templating/go.mod
```diff
@@ -2,13 +2,11 @@ module example_internal_templating
 
 go 1.17
 
-require github.com/k14s/ytt v0.38.0
+replace github.com/vmware-tanzu/carvel-ytt => ../../../
```
pivotaljohn, line 6: What motivates this `replace` directive?

As an example, I assume it's more useful when it's self-contained.

pkg/cmd/template/cmd.go
```diff
@@ -55,26 +54,20 @@ func NewOptions() *Options {
...
-func NewCmd(o *Options) *cobra.Command {
-	cmd := &cobra.Command{
-		Use:     "template",
-		Aliases: []string{"t", "tpl"},
-		Short:   "Process YAML templates (deprecated; use top-level command -- e.g. `ytt -f-` instead of `ytt template -f-`)",
-		RunE:    func(c *cobra.Command, args []string) error { return o.Run() },
-	}
-	cmd.Flags().BoolVar(&o.IgnoreUnknownComments, "ignore-unknown-comments", false,
+// BindFlags registers top level template flags for template command.
+func (o *Options) BindFlags(cmdFlags CmdFlags) {
```
pivotaljohn:
```suggestion
// BindFlags registers template flags for template command.
func (o *Options) BindFlags(cmdFlags CmdFlags) {
```

nit: while this function technically delegates to contained structs for binding their flags, it's whole responsibility is to all flags.

pkg/yttlibrary/all.go
```diff
@@ -7,18 +7,34 @@ import (
...
 	"github.com/vmware-tanzu/carvel-ytt/pkg/yttlibrary/overlay"
 )
 
+var registeredExts []*starlarkstruct.Module
+
+// DoNotUseRegisterExt allows to registry ytt library extensions.
+// It's important to allow only specific module names to be registered
+// as we do not want end-users of ytt to extend @ytt library.
+// (They should define their own library for that.)
+func DoNotUseRegisterExt(mod *starlarkstruct.Module) {
```
pivotaljohn:
```suggestion
// RegisterExt adds "mod" to the standard set of ytt library modules as an extension.
// An "extension" is a Starlark module that has external Go dependencies
// (as opposed to other Starlark modules in the ytt library that either have no dependencies or depend on the Go standard lib)
// This enables those using ytt as a Go module to opt-in (rather than be forced) to accept such dependencies.
// Only Carvel-maintained extensions can be registered; this reserves the `@ytt:` namespace.
// Integrators who want to write their own extensions should construct their own library.
func RegisterExt(mod *starlarkstruct.Module) {
```

Given that it contains mechanism that implements the policy, it feels a bit "overkill" to _also_ prepend the name of the function with `DoNotUse...`.
Add explicit definition of what an "extension" is; providing a well-defined concept for this overall mechanism (registering).
Also suggesting additional context for usage by maintainers.

pkg/yttlibrary/all.go
```diff
@@ -47,7 +62,17 @@ func NewAPI(replaceNodeFunc tplcore.StarlarkFunc, dataMod DataModule,
 		"version": VersionAPI,
 
 		"library": libraryMod,
-	}}
+	}
+
+	for _, ext := range registeredExts {
+		// Double check that we are not overriding predefined library
```
pivotaljohn, line 72: nice! 👍🏻

pkg/yttlibraryext/toml/toml.go
```diff
@@ -14,21 +14,26 @@ import (
...
+			"decode": starlark.NewBuiltin("toml.decode", core.ErrWrapper(tomlModule{}.Decode)),
 		},
 	}
+
+	// TOMLAPI contains the definition of the @ytt:toml module
+	TOMLAPI = starlark.StringDict{"toml": tomlMod}
 )
 
+func init() {
+	yttlibrary.DoNotUseRegisterExt(tomlMod)
```
pivotaljohn:
```suggestion
	yttlibrary.RegisterExt(tomlMod)
```
(see rationale at definition)

pivotaljohn (changes requested):
Minor feedback: all renaming and comment improvements.

Otherwise, looks great.

pkg/yttlibrary/all.go
pivotaljohn: @cppforlife noted:
> the reason why ive added DoNotUse is because i want to strongly discourage anybody from using it.
> it's really a "private" method to ytt project, but go doesnt have a way to define it.

Yes. The menu of "channels" to "signal" here include:
- the method name
- Go Doc
- error messages.

I'm suggesting that the latter two are sufficient to send this signal.
Further that we're "denormalizing" the method namespace with this kind of prefix. (tho, TBH, if we would adopt something like `ModulePrivate_`, `Private` or just `_` as a prefix, I'd be _more_ okay with it — such a convention would easily apply for both integrators and contributors without (much) of a story to tell).

pivotaljohn: We spoke about this.

For now, we'll adopt the changes I suggested.

At the same time, @cppforlife has some best-laid plans for providing for a proper library extensions feature for ytt. So, this will likely evolve through that work.

examples/integrating-with-ytt/internal-templating/go.mod
pivotaljohn, line 6:
```suggestion
replace github.com/vmware-tanzu/carvel-ytt => ../../../   // ensure example works with this copy of ytt; remove before use.
```

pivotaljohn, line 6:
```suggestion
replace github.com/vmware-tanzu/carvel-ytt => ../../../   // ensure example works with this copy of ytt; remove before use.
```

pivotaljohn (approved):

Issue #646, opened 2022-04-11
Make symlink error message more helpful


cari-lynn:
**Describe the problem/challenge you have**
Using symlinks like in Kubernetes to mount a configmap or secret into a container causes a error unless using `--allow-symlink-destination`. This error message is confusing as to why symlinks aren't allowed by default. 

The current error: 
```
$ ytt -f /opt/xxx/config/workshop.yaml
ytt: Error: Checking symlink file '/opt/xxx/config/workshop.yaml':
  Expected symlink file '/opt/xxx/config/workshop.yaml' -> '/opt/xxx/config/..2022_04_09_00_13_38.3873050756/workshop.yaml' to be allowed, but was not
```

**Describe the solution you'd like**
A more clear error that describes how to solve the problem:
```
$ ytt -f /opt/xxx/config/workshop.yaml
ytt: Error: Checking symlink file '/opt/xxx/config/workshop.yaml':
  Expected symlink file '/opt/xxx/config/workshop.yaml' -> '/opt/xxx/config/..2022_04_09_00_13_38.3873050756/workshop.yaml' to be allowed, but was not. Symlinks are disallowed as a security feature, use `--allow-symlink-destination` flags to override
```

**Anything else you would like to add:**
Originating issue: #643 

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

cppforlife:
this is a good example for using `hint`:

```
$ ytt -f /opt/xxx/config/workshop.yaml
ytt: Error: Checking symlink file '/opt/xxx/config/workshop.yaml':
  Expected symlink file '/opt/xxx/config/workshop.yaml' -> '/opt/xxx/config/..2022_04_09_00_13_38.3873050756/workshop.yaml' to be allowed, but was not (hint: symlinks are disallowed as a security feature, use '--allow-symlink-destination' flags to override)
```

vicmarbev:
I would like to work on it!

cari-lynn:
@vicmarbev That's great to hear! We will keep an eye out for a PR from you. Let us know if you have any questions here or in the Kubernetes Slack in [#carvel](https://kubernetes.slack.com/archives/CH8KCCKA5) channel. :)

cppforlife:
merged!

PR #647, opened 2022-04-14
Add hint to symlink error message

vicmarbev:
As specified in issue #646, add a hint to the symlink error message.


cppforlife:
@vicmarbev thanks!

PR #652, opened 2022-04-19
Bug fix: decode TOML array of tables

### LINKED ISSUE #630: @ytt:toml.decode() fails to marshal TOML array of table values
reported: https://kubernetes.slack.com/archives/CH8KCCKA5/p1647400515157369
minimal example: https://carvel.dev/ytt/#gist:https://gist.github.com/pivotaljohn/81df98b1024374ca04ad30ad987cc70b
TOML docs:  https://toml.io/en/v1.0.0#array-of-tableshttps://toml.io/en/v1.0.0#array-of-tables

**Environment:**

- ytt version (use `ytt --version`): v0.40.1

---
Vote on this request

gcheadle-vmware:
fixes: https://github.com/vmware-tanzu/carvel-ytt/issues/630

added `case []map[string]interface{}:` to conversion process.

### DIFF
--- a/pkg/orderedmap/convert.go
+++ b/pkg/orderedmap/convert.go
@@ -74,7 +74,13 @@ func (c Conversion) fromUnorderedMaps(object interface{}) interface{} {
 			typedObj[i] = c.fromUnorderedMaps(item)
 		}
 		return typedObj
-
+	case []map[string]interface{}:
+		resultArray := make([]interface{}, len(typedObj))
+		for i, item := range typedObj {
+			resultMap := c.fromUnorderedMaps(item)
+			resultArray[i] = resultMap
+		}
+		return resultArray
 	default:
 		return typedObj
 	}
--- a/pkg/orderedmap/convert.go
+++ b/pkg/orderedmap/convert.go
@@ -74,11 +74,14 @@ func (c Conversion) fromUnorderedMaps(object interface{}) interface{} {
 			typedObj[i] = c.fromUnorderedMaps(item)
 		}
 		return typedObj
+
+	// some sources (e.g. toml library) yield specific type of slices.
+	// slices in Go are not covariant, so each flavor of slice must have its own case, here.
+	// process these exactly the same way we do for generic slices (prior case)
 	case []map[string]interface{}:
 		resultArray := make([]interface{}, len(typedObj))
 		for i, item := range typedObj {
-			resultMap := c.fromUnorderedMaps(item)
-			resultArray[i] = resultMap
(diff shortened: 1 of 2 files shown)


pkg/orderedmap/convert.go
```diff
@@ -74,7 +74,13 @@ func (c Conversion) fromUnorderedMaps(object interface{}) interface{} {
 			typedObj[i] = c.fromUnorderedMaps(item)
 		}
 		return typedObj
-
+	case []map[string]interface{}:
```
gcheadle-vmware, line 81: Having the case `map[interface{}]interface{}:` earlier in this function makes me think that we should cover the  similar case: `[]map[interface{}]interface{}:`. However, I could not construct a TOML map with non 'string' map keys. 

Ultimately, I was left with this question: 
- "Why is the input of an array of maps not covered by the `[]interface{}` case?"

gcheadle-vmware: Another way to ask the question:
- "Why does `toml.decode` return different types of slices rather than always returning `[]interface{}`" 

Found others hitting up against the `toml.Unmarshal` inconsistencies: https://github.com/BurntSushi/toml/issues/169

cppforlife, line 81: nit (style): other cases have newline before/after them.

pkg/orderedmap/convert.go
```diff
@@ -74,7 +74,13 @@ func (c Conversion) fromUnorderedMaps(object interface{}) interface{} {
...
+		resultArray := make([]interface{}, len(typedObj))
+		for i, item := range typedObj {
+			resultMap := c.fromUnorderedMaps(item)
```
cppforlife: nit: do we need local variable here?

pkg/orderedmap/convert.go
pivotaljohn, line 81:
```suggestion

	// some sources (e.g. toml library) yield specific type of slices.
	// slices in Go are not covariant, so each flavor of slice must have its own case, here.
	// process these exactly the same way we do for generic slices (prior case)
	case []map[string]interface{}:
```

For now: Leave ourselves a note about why we have two cases with the same body.
Soon: we'll eventually upgrade to 1.18. At that point, we might be able to make this collection generic. 😁

pivotaljohn (approved):
LGTM.

In addition to @cppforlife's formatting nits, I'm suggesting a note to explain both where this specific type of slice is coming from and why we can't just combine all the cases... for now.

Thanks for tackling this one, @gcheadle-vmware.

PR #653, opened 2022-04-21
Data value schema validation

vmunishwar:


pkg/schema/schema.go
```diff
@@ -145,6 +162,20 @@ func getValue(node yamlmeta.Node, t Type) (interface{}, error) {
...
+func getValidations(node yamlmeta.Node) ([]assertions.Rule, error) {
+	ann, err := processOptionalAnnotation(node, AnnotationValidation, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	validationAnn, ok := ann.(*ValidationAnnotation)
+	if !ok {
+		return nil, nil
+	}
```
pivotaljohn: If I have to do a type check to determine if the annotation is present, it's a kind of complexity that I want to make sure I'm getting something for...

And when I review `processOptionalAnnotation()`, I see _some_ common code, but perhaps not enough to justify needing to make moves this ☝🏻 .

Suggestion: extract out of `processOptionalAnnotation()` the logic for handling `AnnotationValidation` into its own function that returns an `AnnotationValidation`... which would make this chunk unnecessary.

pkg/schema/type.go
```diff
@@ -564,6 +576,55 @@ func (n *NullType) SetDeprecated(deprecated bool, notice string) {
 	n.documentation.deprecated = deprecated
 }
 
+// GetValidations provides validations from @schema/validation for a node
+func (t *DocumentType) GetValidations() []assertions.Rule {
+	if t.validations != nil {
+		return t.validations
+	}
+	return nil
```
pivotaljohn:
```suggestion
	return t.validations
```
if `t.validations` is _in fact_ `nil`, then `nil` will be returned.

(applies to the other getters, below)

pkg/workspace/data_values_pre_processing.go
```diff
@@ -67,6 +67,8 @@ func (pp DataValuesPreProcessing) apply(files []*FileInLibrary) (*datavalues.Env
 		if len(typeCheck.Violations) > 0 {
 			return nil, nil, schema.NewSchemaError("One or more data values were invalid", typeCheck.Violations...)
 		}
+		//Walk updates node's validations meta from Node's assigned type
```
pivotaljohn: nit: for consistency...

```suggestion
		// Walk updates node's validations meta from Node's assigned type
```

pkg/assertions/yamlmeta.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package assertions
+
+import "github.com/vmware-tanzu/carvel-ytt/pkg/yamlmeta"
+
+// AddValidations appends validation Rules to node's validations metadata, later retrieved via GetValidations().
+func AddValidations(node yamlmeta.Node, rules []Rule) {
+	metas := node.GetMeta("validations")
```
pivotaljohn: Magic string: let's extract to a constant for that added confidence that it's the same. 👍🏻

pkg/assertions/yamlmeta.go
```diff
@@ -0,0 +1,29 @@
...
+	if currRules, ok := metas.([]Rule); ok {
```
pivotaljohn: I see how `ok` is `false` when `metas` is `nil`. That's pretty slick.

At the same time, there's the (should never happen) situation where `metas` is both _not `nil`_ and _not `[]Rule`_.

Given the role that `ytt` plays — a often low-level integrated piece in a larger mass of infrastructure automation — it pays to be _as boring as possible_ in order to avoid any chance of strange behaviors or unexpected errors.
For me, it took a bit to parse all of what's intended here, given that a failed type assertion is playing the role of a nil-check, here.

Suggestion: for readability and reliability's sake, let's err on the side of "boring code":
- split the lazy initialization (i.e. nil check) from the type assertion.
- on failed type assertion, panic.

_(I clearly didn't read through this closely enough, prior. Apologies for the late call, here.)_

pivotaljohn (changes requested):
Some tweaks in this review.

While we're at it, let's squash some of the git history, here.

pkg/assertions/yamlmeta.go
```diff
@@ -3,25 +3,35 @@
...
 func SetValidations(node yamlmeta.Node, rules []Rule) {
-	node.SetMeta("validations", rules)
+	node.SetMeta(validations, rules)
 }
 
 // GetValidations retrieves validation Rules from node metadata, set previously via SetValidations().
 func GetValidations(node yamlmeta.Node) []Rule {
-	metas := node.GetMeta("validations")
+	metas := node.GetMeta(validations)
 	if rules, ok := metas.([]Rule); ok {
```
gcheadle-vmware: 🤔  wondering if we should not return `nil` on a bad cast and panic if `validations` was set with something other than `[]Rule`?

gcheadle-vmware: Discussion: Return the direct cast and let go take care of the panic. cast the value without checking.

PR #661, opened 2022-05-02
Bug fix: have data.list("/") read from root of library

### LINKED ISSUE #645: Values file `data.list("/")` should read from root library
**What steps did you take:**
When `data.list("/")` within a data values file, it should list the files relative to the root library. Playground reproduction example: https://carvel.dev/ytt/#gist:https://gist.github.com/cari-lynn/8b7c194fd28bcab0ec80e716eecc161f

**What happened:**
It will list no files at all.

**What did you expect:**
I expect that `data.list("/")` will list all files relative to the root library

`data.list("")` without the `/` will list files relative to the current directory just as expected. This problem just shows up when using the `/`.

I believe this happens because we don't pass down the root library during data values pre processing. But I believe this could be changed.

https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/workspace/data_values_pre_processing.go#L119-L121

[ytt Docs](https://carvel.dev/ytt/docs/v0.40.0/lang-ref-ytt/#data)

[originating Slack thread in #carvel](https://kubernetes.slack.com/archives/CH8KCCKA5/p1649442906868389)

**Environment:**

- ytt version (use `ytt --version`): v0.40.1
- OS (e.g. from `/etc/os-release`):

---
Vote on this request

gcheadle-vmware:


pkg/cmd/template/cmd_test.go
```diff
@@ -210,6 +210,94 @@ data: #@ data.read("/funcs/data")`)
 	assert.Equal(t, expectedYAMLTplData, string(file.Bytes()))
 }
 
+func TestDataListRelativeToRootInDataValuesAndSchema(t *testing.T) {
```
pivotaljohn: These tests give me _a lot_ of confidence that our "absolute path" loading works in both schema and data values -flavored templates. 👍🏻  Thank you for investing in this.

As I look around at the test files in `pkg/cmd/template` ... their ranks are growing. `cmd_test.go` is almost a KLOC.

If you haven't already thought about this, I'm wondering if these tests would be more "at home" in one of these test files:
- `cmd_root_library_test.go`
- `cmd_data_values_test.go`

pkg/workspace/data_values_schema_pre_processing.go
```diff
@@ -97,7 +98,7 @@ func (pp DataValuesSchemaPreProcessing) collectSchemaDocs(schemaFiles []*FileInL
 }
 
 func (pp DataValuesSchemaPreProcessing) extractSchemaDocs(schemaFile *FileInLibrary) ([]*yamlmeta.Document, error) {
-	libraryCtx := LibraryExecutionContext{Current: schemaFile.Library, Root: NewRootLibrary(nil)}
+	libraryCtx := LibraryExecutionContext{Current: schemaFile.Library, Root: pp.rootLibrary}
```
pivotaljohn, line 101: How about how minimizing the differences between Data Values and Data Values Schema processing makes this sweetly obvious that making the exact same change in both flows is the right move? ⭐

pivotaljohn (approved):
LGTM. Thank you for the thorough approach, here.

Comments are just for your consideration.

If/When you're happy, good to merge.

PR #662, opened 2022-05-04
Data value schema disable validation

pivotaljohn:
Adds:
- `--data-values-disable-validations` flag to command-line interface
- `data_values_disable_validations` keyword argument to the `@ytt:library.instance.eval()` function.


pivotaljohn:
I have questions:
- Should all these components be aware of whether Data Validations are being skipped or not?
- There's two verbs being used, here: "disable validations" and "skip validations". We aim to employ [Ubiquitous Language](https://thedomaindrivendesign.io/developing-the-ubiquitous-language/#:~:text=Characteristics%20of%20the%20Ubiquitous%20Language) so as to avoid the accidental complexity that is "translating" from one set of terms to another. Is this a violation of that principle?
- Even when validations are disabled, if they are malformed (i.e. there's a syntax error), it still results in a templating error. Is that okay?

gcheadle-vmware:
> Even when validations are disabled, if they are malformed (i.e. there's a syntax error), it still results in a templating error. Is that okay?

I think that if an annotation is not valid starlark, then yes it should be reported as a syntax error. If there is an error about the order or number of arguments in a tuple, then no, I don't think there should be an error, which is consistent with how I am reading the code.

Currently half way through processing this PR and I will follow up tomorrow.

pkg/cmd/template/assert_validate_test.go
```diff
@@ -371,3 +371,149 @@ values: #@ data.values
...
+			opts := cmdtpl.NewOptions()
+			opts.DataValuesFlags.SkipValidation = true
+			configYAML := `
+#@ load("@ytt:template", "template")
+#@ load("@ytt:library", "library")
+
+--- #@ template.replace(library.get("lib").eval())
+
+#! even if validations are explicitly enabled...
+--- #@ template.replace(library.get("lib").eval(data_values_disable_validation=True))
```
gcheadle-vmware: The comment above this line makes me think that this library evaluation will have validations disabled through the command line flag (set on 398), but enabled through the `eval` parameter. Maybe I'm missing something, but the default value for `data_values_disable_validation` is false, which is the same thing as being enabled. So explicitly enabling validations via the flag doesn't make sense to me. 🤔

pkg/feature/flags.go
```diff
@@ -0,0 +1,82 @@
...
+
+    feature.Flags().Enable(<feature-constant>)
+
+To then circuit-break functionality behind a feature flag.
+
+    if feature.Flags().IsEnabled(<feature-constant>) {
+        ...
+    }
+*/
+package feature
```
cari-lynn: 👏 great explaination

pkg/feature/flags.go
```diff
@@ -0,0 +1,82 @@
...
+
+import "fmt"
+
+// Names of features that can be toggled
+const (
+	Noop        = "noop"
```
cari-lynn: What's the intention behind `noop`?

pkg/workspace/library_module.go
```diff
@@ -18,6 +18,10 @@ import (
 	"github.com/vmware-tanzu/carvel-ytt/pkg/yttlibrary/overlay"
 )
 
+// LibraryModule is the definition of the ytt-supplied Starlark module `@ytt:library`
+//
+// This module produces library instances (see Get()). Each such instance is handed the same configuration as the
+// library execution which is spawning the new instance. (see NewLibraryModule())
```
cari-lynn:
```suggestion
// LibraryModule is the definition of the ytt-supplied Starlark module `@ytt:library`
//
// This module produces library instances (see Get()). The configuration is copied from the 
// library execution to the new instance of a library module (see NewLibraryModule())
```

pkg/workspace/library_module.go
```diff
@@ -237,7 +246,23 @@ func (l *libraryValue) Eval(thread *starlark.Thread, f *starlark.Builtin,
...
+	skipDataValuesValidation, err := core.BoolArg(kwargs, "data_values_disable_validation")
+	if err != nil {
+		return starlark.None, err
+	}
+
+	libraryExecutionFactory := l.libraryExecutionFactory
+	if skipDataValuesValidation {
+		libraryExecutionFactory = l.libraryExecutionFactory.ThatSkipsDataValuesValidations()
+	}
+	libraryExecution := libraryExecutionFactory.New(l.libraryCtx)
```
gcheadle-vmware: When evaluating a library we call `libraryValues`, which will validate data values annotated with `@assert/validate`. 

Have we considered including the option to disable validations for the other library functions that call `libraryValues`?

cmd/ytt/ytt.go
```diff
@@ -24,3 +27,12 @@ func main() {
 		os.Exit(1)
 	}
 }
+
+func enableExperimentalFeatures() {
+	if experiments, isSet := os.LookupEnv("YTTEXPERIMENTS"); isSet {
+		for _, experiment := range strings.Split(experiments, ",") {
+			feature.Flags().Enable(experiment)
+			fmt.Fprintf(os.Stderr, "Experimental feature %q enabled.\n", experiment)
```
cari-lynn: I am not sure on this, but we should validate with kapp-controller if writing to stderr somehow interferes with how they report a successful template step. (This may not be an issue at all, just covering our bases here)

cari-lynn (approved):
> Even when validations are disabled, if they are malformed (i.e. there's a syntax error), it still results in a templating error. Is that okay?

We have a lot of checks in place (feature flag, and env variable) to make sure users are sure they want to enable this functionality, but I'm thinking of a possible use case..

if you have written your validations on ytt 0.41.0, but then in the following versions the validation syntax changes, and you use a newer version, say 0.43.0 for example, and disable validations because you just want to test it, you’ll still see syntax errors. This could be worse if you're using templates you don’t own because annotations can’t be changed with overlays yet. 

You'd still need to of course have the feature flag and environment variable set, and I assume we will have warned users there may be breaking changes, but I still think that it may be better to not error when syntax is wrong when `--disable-data-values-schema-validations` is set, since they aren't being used at all.

gcheadle-vmware (commented):
> "disable validations" and "skip validations"

I was slightly confused at first, but when I saw that the flag and the library eval parameter were both named "data values disable validations", it became clear. 

I am having trouble seeing exactly how setting the flag `--data-values-disable-validations=true` is passed down to all private libraries (I think it's via the libraryExecutionFactory). I see in the tests that "when --data-values-disable-validations=true, data values are NOT validated for any library, regardless". But with that statement, I'm wondering if we should be disabling validations when we GET the library, rather than when we evaluate it?

pkg/cmd/template/data_values_flags.go
```diff
@@ -55,6 +56,7 @@ func (s *DataValuesFlags) Set(cmdFlags CmdFlags) {
 	cmdFlags.StringArrayVar(&s.FromFiles, "data-values-file", nil, "Set multiple data values via a YAML file (format: /file/path.yml) (can be specified multiple times)")
 
 	cmdFlags.BoolVar(&s.Inspect, "data-values-inspect", false, "Calculate the final data values (applying any overlays) and display that result")
+	cmdFlags.BoolVar(&s.SkipValidation, "data-values-disable-validation", false, "Do not attempt to validate data values (warning: might allow invalid values)")
```
cppforlife: data-values-disable-validation => data-values-validation=true (enabled by default)

should this be called `data-values-schema-validation`? is validation part of schema or can i also apply validations without schema?

pkg/feature/flags.go
pivotaljohn: 🙇🏻

pkg/feature/flags.go
pivotaljohn: Honestly?  It was born out of testing — needing to have a "feature" that could be verified enabled or disabled.

But then as I used it, thought it would be nice to differentiate between `YTTEXPERIMENTS=` containing an empty string, which currently panics....  (which as I wrote that sentence don't like that as an experience... 🤔 ).

...

Yeah, I'm gonna document this as "intended for testing" thing... _and_ **not** panic when `YTTEXPERIMENTS=` is an empty string.

cmd/ytt/ytt.go
pivotaljohn: Great question!

I'll follow-up.

pkg/cmd/template/assert_validate_test.go
pivotaljohn: > I think it's via the libraryExecutionFactory

That's _exactly_ how that works. ☝🏻 

Also, sharing what we covered, in person: that (as @cppforlife also suggested, below), it's way easier to understand how flags work when they are expressed in the positive. I'm using the double-negative, here, and it's confusing.

I'll do that rename. 👍🏻

pkg/workspace/library_module.go
pivotaljohn: Very good catch. You're absolutely right. This flag should live on `@ytt:library.get()` not `@ytt:library.eval()`; otherwise, the author could receive data or results that are invalid.

🙌🏻

pkg/cmd/template/data_values_flags.go
pivotaljohn: > data-values-disable-validation => data-values-validation=true (enabled by default)

👍🏻  (note conversation above)

pivotaljohn: > should this be called data-values-schema-validation? is validation part of schema or can i also apply validations without schema?

This flag was meant to short-circuit validating data values, regardless of the source of the validation rules.

And yes, you can annotate `@assert/validate` in the case where you want to write validations without having to specify Schema.

cmd/ytt/ytt.go
pivotaljohn: It does not. 👍🏻 

https://kubernetes.slack.com/archives/CH8KCCKA5/p1652123416054149

pivotaljohn:
_(see also conversation in #carvel about the appropriate scope of this flag: https://kubernetes.slack.com/archives/CH8KCCKA5/p1652295963803589)_

PR #666, opened 2022-05-11
Setting buildid is no longer necessary for reproducible builds

benmoss:
This was fixed in go 1.14 https://github.com/golang/go/commit/aa680c0c49b55722a72ad3772e590cd2f9af541d


pivotaljohn:
🙏🏻

PR #667, opened 2022-05-11
add Validations when and when_null_skip kwargs

gcheadle-vmware:
The diff is hard to read so I'm open to going over the changes together.

A refactor caused the experiment check to be moved so I moved it back to where I felt it belonged. I also added an experiment check when processing schema validations.

### DIFF
--- a/pkg/validations/assert.go
+++ b/pkg/validations/assert.go
@@ -14,9 +14,9 @@ import (
 
 // Declare @assert/... annotation names
 const (
-	AnnotationAssertValidate template.AnnotationName = "assert/validate"
-	ValidationKwargWhen      string                  = "when"
-	ValidationKwargWhenNull  string                  = "when_null_skip"
+	AnnotationAssertValidate    template.AnnotationName = "assert/validate"
+	ValidationKwargWhen         string                  = "when"
+	ValidationKwargWhenNullSkip string                  = "when_null_skip"
 )
 
 // ProcessAssertValidateAnns checks Assert annotations on data values and stores them on a Node as Validations.
@@ -31,7 +31,7 @@ func ProcessAssertValidateAnns(rootNode yamlmeta.Node) error {
 type convertAssertAnnsToValidations struct{}
 
 // Visit if `node` is annotated with `@assert/validate` (AnnotationAssertValidate).
-// Checks annotation, and stores validation Rules on Node's validations meta.
+// Checks annotation, and stores the validation on Node's validations meta.
 //
 // This visitor returns and error if any assert annotation is not well-formed,
 // otherwise, returns nil.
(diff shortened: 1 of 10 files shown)


pkg/validations/assert.go
```diff
@@ -7,71 +7,31 @@ import (
...
-func (ac *AssertCheck) HasViolations() bool {
-	return len(ac.Violations) > 0
+	return yamlmeta.Walk(rootNode, &convertAssertAnnsToValidations{})
 }
 
 type convertAssertAnnsToValidations struct{}
 
 // Visit if `node` is annotated with `@assert/validate` (AnnotationAssertValidate).
-// Checks annotation, and stores validation Rules on Node's validations meta.
+// Checks annotation, and stores the validation on Node's validations meta.
```
pivotaljohn, line 34: When refactoring, it's easy to miss the attendant documentation; but here, you didn't. 🙏🏻 

It's attention to these details that makes our codebase _that_ much more approachable.

pkg/validations/validate.go
```diff
@@ -12,47 +12,174 @@ import (
...
+	for _, r := range v.rules {
+		result, err := starlark.Call(thread, r.assertion, starlark.Tuple{nodeValue}, []starlark.Tuple{})
+		if err != nil {
+			failures = append(failures, fmt.Errorf("%s (%s) requires %q; %s (by %s)", key, node.GetPosition().AsCompactString(), r.msg, err.Error(), v.position.AsCompactString()))
+		} else if _, ok := result.(starlark.NoneType); !ok {
+			// in order to pass, the assertion must return True or None
+			if !result.Truth() {
+				failures = append(failures, fmt.Errorf("%s (%s) requires %q (by %s)", key, node.GetPosition().AsCompactString(), r.msg, v.position.AsCompactString()))
+			}
+		}
```
pivotaljohn:
```suggestion
			_, isNone := result.(starlark.NoneType)
			isTrue := bool(result.Truth())
			if !(isNone || isTrue) {
				failures = append(failures, fmt.Errorf("%s (%s) requires %q (by %s)", key, node.GetPosition().AsCompactString(), r.msg, v.position.AsCompactString()))
			}
		}
```

Even after talking it through, this hunk tripped my internal threshold for cognitive load.

Suggesting we decompose this just slightly to make more explicit what it means for a rule to pass: it's either `None` or `True`-(thy)

pkg/validations/validate.go
```diff
@@ -12,47 +12,174 @@ import (
...
+
+		switch typedResult := result.(type) {
+		// if None or True, then execute rules
+		case starlark.NoneType:
+			return true, nil
+		case starlark.Bool:
+			return bool(typedResult), nil
+		default:
+			return false, fmt.Errorf("when kwarg function must reutrn a bool or None type")
+		}
```
pivotaljohn:
```suggestion
		_, isNone := result.(starlark.NoneType)
		isTrue := bool(result.Truth())
		return isNone || isTrue, nil
```

I'm not seeing a problem in allowing even the value for `when=` to be truth-y (rather than strictly `True`).

Is there?

pkg/workspace/data_values_pre_processing.go
```diff
@@ -114,6 +112,10 @@ func (pp DataValuesPreProcessing) typeAndCheck(dataValuesDoc *yamlmeta.Document)
 	if len(chk.Violations) > 0 {
 		return chk
 	}
+
+	// updates node's validations meta from Node's assigned type
+	yamlmeta.Walk(dataValuesDoc, schema.AssignSchemaValidations{})
```
pivotaljohn:
```suggestion
	_ = yamlmeta.Walk(dataValuesDoc, schema.AssignSchemaValidations{})
```

Let's ignore _explicitly_ to differentiate from cases where we forget to handle errors.

pivotaljohn (changes requested):
Some readability improvements suggested.

Notably, that `when=` (I'm fairly certain) ought to consider any truthy value as a pass, not just literal Bool

pkg/validations/validate.go
gcheadle-vmware: Based on the proposal, there is no difference between the argument to a `when` kwarg and an assertion, so we should handle the return result similarly. 👍

PR #670, opened 2022-05-14
Support loading Data Values files in the same ways "ordinary" files are loaded

### LINKED ISSUE #657: data-values-file should accept an entire directory
I keep my values files in a directory, which I'm very sure contains only values files.  I don't have a lot of control over the naming of the files in that folder.

I understand that currently 
`ytt --data-value-file /values`
does not work, because it expects:
`ytt --data-value-file /values/file.yaml --data-value-file /values/file2.yaml --data-value-file /values/file3.yaml`
It would be nice if it did accept the directory, and just implicitly applied a filemark to everything in that directory to then be a data value file.

Request originated in this [#carvel thread](https://kubernetes.slack.com/archives/CH8KCCKA5/p1651167583289939).

---
Vote on this request

pivotaljohn:
Reusing the strategy by which a filepath from `--file` is used to resolve to the contents of the file(s) for loading Data Values.

All types of inputs are now supported (and handled identically) across `--file`, `--data-value-file`, and `--data-values-file`: 🎉 
- ordinary file
- filesystem directories
- named pipes
- HTTP/HTTPS URLs
- standard input

### DIFF
--- a/pkg/cmd/template/data_values_flags.go
+++ b/pkg/cmd/template/data_values_flags.go
@@ -5,7 +5,6 @@ package template
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"strings"
 
@@ -40,8 +39,10 @@ type DataValuesFlags struct {
 	InspectSchema  bool
 	SkipValidation bool
 
-	EnvironFunc  func() []string
-	ReadFileFunc func(string) ([]byte, error)
+	EnvironFunc   func() []string
+	ReadFilesFunc func(paths string) ([]*files.File, error)
+
+	*files.SymlinkAllowOpts
 }
 
 // Set registers data values ingestion flags and wires-up those flags up to this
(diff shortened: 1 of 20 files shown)


pkg/cmd/template/data_values_flags.go
```diff
@@ -149,34 +152,43 @@ func (s *DataValuesFlags) file(fullPath string, strict bool) ([]*datavalues.Enve
...
+	dvFiles, err := s.asFiles(path)
 	if err != nil {
-		return nil, fmt.Errorf("Unmarshaling YAML data values file '%s': %s", path, err)
+		return nil, fmt.Errorf("Find files '%s': %s", path, err)
 	}
 
 	var result []*datavalues.Envelope
+	for _, dvFile := range dvFiles {
+		// skip non-YAML files that are contained within the specified directory
+		if strings.Contains(dvFile.RelativePath(), "/") && !(dvFile.Type() == files.TypeYAML) {
```
cppforlife: > strings.Contains(dvFile.RelativePath(), "/")

whats this bit about?

pkg/cmd/template/data_values_flags.go
```diff
@@ -40,8 +39,10 @@ type DataValuesFlags struct {
 	InspectSchema  bool
 	SkipValidation bool
 
-	EnvironFunc  func() []string
-	ReadFileFunc func(string) ([]byte, error)
+	EnvironFunc   func() []string
+	ReadFilesFunc func(paths string) ([]*files.File, error)
+
+	files.SymlinkAllowOpts
```
cppforlife: since users control the "values pipe" i dont think we should be restricting usage of symlinks for data value ingestion.

pivotaljohn: I had included this more with an understanding of protection/confidence...

That is, safe defaults with the option to do riskier things if you indicate you accept that risk.

We're walking a directory tree that is providing data values... shouldn't we ensure that if we're traversing a symlink that the user intends it?

Also, if our intention is to support traversing directories, we've got more work to do here.

pkg/cmd/template/data_values_flags.go
pivotaljohn: If the user supplies a non-YAML file directly as the parameter for the flag, we don't want to ignore it. It's only _implied_ files (i.e. contained within a subdirectory) that we want to ignore.

pivotaljohn: ... although, I had intended to check if `RelativePath()` was returning a normalized path or if we needed to grab the OS-specific path separator token, here.

I'll do that.

pkg/cmd/template/data_values_flags.go
pivotaljohn: _(discussed in [Slack 🧵  in #carvel](https://kubernetes.slack.com/archives/CH8KCCKA5/p1653053084082349))_

pkg/cmd/template/data_values_flags.go
pivotaljohn: _(discussed in [Slack :thread: in #carvel](https://kubernetes.slack.com/archives/CH8KCCKA5/p1653056007351229?thread_ts=1653053084.082349&cid=CH8KCCKA5))_

pivotaljohn: Added `File.IsImplied()` and make check OS-aware.

pkg/cmd/template/data_values_flags.go
pivotaljohn: Resolution: keep for now.

gcheadle-vmware (approved):
I've enjoyed watching this feature evolve into something that uses existing mechanisms in ytt of reading files from a Dir. Nothing stood out to me that was alarming + naming & godocs really helped me understand these changes. 🎊

PR #673, opened 2022-05-19
follow-up: include when_null_skip=True on Nullable Schema types

gcheadle-vmware:


pkg/validations/validate.go
```diff
@@ -106,18 +106,27 @@ func (v NodeValidation) Validate(node yamlmeta.Node, thread *starlark.Thread) []
...
+	if v.kwargs.whenNullSkip == nil {
+		t := true
+		v.kwargs.whenNullSkip = &t
+	}
+}
+
 // shouldValidate uses validationKwargs and the node's value to run checks on the value. If the value satisfies the checks,
 // then the NodeValidation's rules should execute, otherwise the rules will be skipped.
 func (v validationKwargs) shouldValidate(value starlark.Value, thread *starlark.Thread) (bool, error) {
-	if v.whenNullSkip {
```
pivotaljohn, line 120: If we hoist the type check for `NoneType` and reuse it, I think it reduces the overall cognitive load:
- I think the first `if` statement can now be combined more easily and stay readable.
  - the parens are not necessary for the compiler, but the humans can more easily parse that the nil-check and value check go together with them in place.
- no need to do the type check twice, so net fewer lines... 


```go
	_, isNone := value.(starlark.NoneType)
	if isNone && (v.whenNullSkip != nil && *v.whenNullSkip) {
		return false, nil
	}

	if v.when != nil {
		result, err := starlark.Call(thread, *v.when, starlark.Tuple{value}, []starlark.Tuple{})
		if err != nil {
			return false, err
		}

		return isNone || isTrue, nil
	}

```

pkg/validations/validate.go
```diff
@@ -29,7 +29,7 @@ type rule struct {
 // validationKwargs represent the optional keyword arguments and their values in a validation annotation.
 type validationKwargs struct {
 	when         *starlark.Callable
-	whenNullSkip bool
+	whenNullSkip *bool
```
pivotaljohn:
```suggestion
	whenNullSkip *bool  // default: true if value is nullable
```

Explicitly define the meaning of `nil`?

pivotaljohn (approved):
Two minor observations on readability. Otherwise, LGTM; merge at will.

Issue #642, opened 2022-04-04
Require consumer to supply a Data Value (string, scalar, maps)


pivotaljohn:
As a library/package author,
Some subset of Data Values I define, I can't possibly know what the values ought to be.
Or I want to force the end-user to supply a value.

To handle cases like this ☝🏻 ,
I want to be able to use out-of-the-box validation rules that generally cover this situation.

👉🏻  In effect, address the [Required Input](https://github.com/vmware-tanzu/carvel/blob/004-schema-validation/proposals/ytt/004-schema-validation/README.md#use-case-required-input) use case.

---
An approach:
- to `@ytt:assert` add `min_len()`, `max_len()`, `min()`, `max()`, and `not_null()`
- on `@assert/validate`, implement `min_len=`, `max_len=`, `min=`, `max=`, `not_null=`
- to `@schema/validation`, add `min_len=`, `max_len=`, `min=`, `max=`, `not_null=`

Consider:
- how should cross type comparisons work?
  - Starlark defines length (e.g. [`len()`](https://github.com/google/starlark-go/blob/master/doc/spec.md#len))
    - `yamlmeta.Map`s and `yamlmeta.Array`s ought to be considered a sequence.
  - Starlark also defines behavior of min/max over sequences (in [`min()`]((https://github.com/google/starlark-go/blob/master/doc/spec.md#min)).

Note: defer the implementation of `#@schema/validation-defaults-for-strings` for this story.

---

## Implementation
- #677 
- vmware-tanzu/carvel#466

PR #677, opened 2022-05-25
Validations require input via kwargs

gcheadle-vmware:
Contributes to #642 

## Feature Work

- [x] (@vmunishwar) allow assertion functions to be intuitively combined in boolean expressions — assertion functions return `True`, not `None`
- [x] (@pivotaljohn) support concise use of assertion functions in custom rules (e.g. `("custom def message", assert.min(1))` instead of `("custom def message", assert.min(1).check)`) — allow not just Callables, but also objects that have an `check` attribute that is a Callable.
- [x] implement `assert.not_null(v)` short-hand.
- [x] (@vmunishwar) store `min_len` and `max_len` as `starlark.Int`s (details: https://github.com/vmware-tanzu/carvel-ytt/pull/677#discussion_r910447266)
  - [x] remove comment (https://github.com/vmware-tanzu/carvel-ytt/pull/677#discussion_r910448304)
- [x] put new `@ytt:assert` functions into "validation" experiment to preserve the "right" to change the API.
  - [x] update docs for each such function with the same warning we have for `@schema/validation`
- [x] verify that duplicate validation kwargs are detected and result in an error.
- [x] switch the order of `fail()` and `True` for existing rules — to make the code more intuitive/readable.
- [x] `one_not_null()`/`one_not_null=`
  - [x] accept sequence of keys or bool for `assert.one_not_null()` and `one_not_null=`
  - [x] schema specific tests (`one_not_null=`)
  - [x] report helpful error message when value(s) is/are not a map.
- [x] Update GoDoc for `Max()`/`Min()`/`MaxLength()`/`MinLength()`
- [x] exploratory test edge-cases for inputs to validation kwargs.
  - when two validations rules contradict (e.g. min/max where max < min; min_len/max_len where max_len < min_len; etc.)
  - 0 cases
  - assert.<assertion-function>() with a misspelled name
  - where min=/max= are given a value that has no minimum or maximum defined, does the error message make sense? (e.g. None?)

## Fit and Finish

- [x] are there "TODO"s left in the code? Let's implement or remove.


pkg/validations/validate.go
```diff
@@ -131,8 +132,10 @@ func (v *NodeValidation) DefaultNullSkipTrue() {
 // shouldValidate uses validationKwargs and the node's value to run checks on the value. If the value satisfies the checks,
 // then the NodeValidation's rules should execute, otherwise the rules will be skipped.
 func (v validationKwargs) shouldValidate(value starlark.Value, thread *starlark.Thread) (bool, error) {
-	_, isNone := value.(starlark.NoneType)
-	if isNone && (v.whenNullSkip != nil && *v.whenNullSkip) {
+	_, isNull := value.(starlark.NoneType)
+	nullAllowed := !v.notNull
+	whenNullSkip := v.whenNullSkip != nil && *v.whenNullSkip
+	if isNull && nullAllowed && whenNullSkip {
```
pivotaljohn, line 138: To keep this `if` from getting out of hand, captured each intention in a variable.

pkg/yttlibrary/assert.go
```diff
@@ -261,21 +265,14 @@ func (b assertModule) Max(thread *starlark.Thread, f *starlark.Builtin, args sta
...
 func (b assertModule) NotNull(thread *starlark.Thread, f *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
-	if len(args) == 0 {
-		return starlark.None, fmt.Errorf("expected at least one argument.")
-	}
-	result, err := starlark.Call(thread, NewAssertNotNull(), args, []starlark.Tuple{})
-	if err != nil {
-		return starlark.None, err
-	}
-	return result, nil
+	return NewAssertNotNull(), nil
```
pivotaljohn: `assert.not_null()` is special: it does not take any parameters, so we just hand back the "assertion object", here. 👍🏻

pkg/validations/assert.go
```diff
@@ -110,6 +119,31 @@ func newValidationKwargs(kwargs []starlark.Tuple, annPos *filepos.Position) (val
 			}
 			b := bool(v)
 			processedKwargs.whenNullSkip = &b
+		case ValidationKwargMinLength:
+			v, err := starlark.NumberToInt(value[1])
```
gcheadle-vmware: Should we document the rounding behavior of floats? or just document it as an int?

pkg/validations/assert.go
```diff
@@ -110,6 +119,31 @@ func newValidationKwargs(kwargs []starlark.Tuple, annPos *filepos.Position) (val
 			}
 			b := bool(v)
 			processedKwargs.whenNullSkip = &b
+		case ValidationKwargMinLength:
```
gcheadle-vmware: If we consider someone providing multiple `min_len` kwargs, would it make sense to take the greater minimum value since they are comparable numbers?

However, it is consistent if we always took the last one, which is how all these kwargs are implemented.

pkg/validations/validate.go
```diff
@@ -30,6 +32,12 @@ type rule struct {
 type validationKwargs struct {
 	when         *starlark.Callable
 	whenNullSkip *bool // default: nil if kwarg is not set, True if value is Nullable
+	minLength    int   // 0 len("") == 0, this always passes
+	maxLength    *int
+	min          starlark.Value
+	max          starlark.Value
+	notNull      bool
+	// *int , default=nil possible_values={&1, &2, &-3, ..}
```
gcheadle-vmware: should this comment be on line 36?

pkg/validations/validate.go
```diff
@@ -125,8 +133,10 @@ func (v *NodeValidation) DefaultNullSkipTrue() {
 // shouldValidate uses validationKwargs and the node's value to run checks on the value. If the value satisfies the checks,
 // then the NodeValidation's rules should execute, otherwise the rules will be skipped.
 func (v validationKwargs) shouldValidate(value starlark.Value, thread *starlark.Thread) (bool, error) {
-	_, isNone := value.(starlark.NoneType)
-	if isNone && (v.whenNullSkip != nil && *v.whenNullSkip) {
+	_, isNull := value.(starlark.NoneType)
+	nullAllowed := !v.notNull
```
gcheadle-vmware: I have to remind myself that `v.notNull` is false by default in order to understand the if statement below.

pkg/validations/validate.go
```diff
@@ -137,12 +147,53 @@ func (v validationKwargs) shouldValidate(value starlark.Value, thread *starlark.
...
+		return isNull || isTrue, nil
 	}
 	// if no kwargs then execute rules
 	return true, nil
 }
 
+func (v validationKwargs) convertToRules() []rule {
+	var rules []rule
+	if minLen := v.minLength; minLen > 0 {
+		a := yttlibrary.NewAssertMinLen(core.NewGoValue(minLen).AsStarlarkValue())
```
gcheadle-vmware: I'm wondering if `minLen` and `maxLen` could just remain as their starlark types? It looks like we are converting it from a `starlark.Int` to an `int` when we construct `validationKwargs`, and then back to a starlark value here. We could save some sweat if we kept it as a starlark value the whole time.

The only concern I have with making such a change is that we are using the int in multiple places around this line of code (`minLen > 0`, `fmt.Sprintf("length greater or equal to %v", minLen)`).

gcheadle-vmware (commented):
Overall, this PR is shaping up really nicely. Really proud of our group effort! @vmunishwar, @pivotaljohn 
I've left my additional thoughts in the comments.

pkg/validations/assert.go
```diff
@@ -59,8 +64,8 @@ func (a *convertAssertAnnsToValidations) Visit(node yamlmeta.Node) error {
 // If any value in the annotation is not well-formed, it returns an error.
 func NewValidationFromValidationAnnotation(annotation template.NodeAnnotation) (*NodeValidation, error) {
 	var rules []rule
-
-	if len(annotation.Args) == 0 {
+	// TODO: what makes a malformed annotation??
```
gcheadle-vmware: Is this a malformed annotation? `#@schema/validation max_len=3, min_len=5`

pkg/validations/assert.go
pivotaljohn: Documenting as an integer, but accepting any number seems legit.

Starlark is dynamically typed; programmers will expect automatic coercion of numbers (maybe not necessarily 1.5 => 1, but 1.0 => 1, for sure).

pkg/validations/assert.go
pivotaljohn: I'm trying to come up with a use-case where an author _intentionally_ uses a keyword argument more than once...

... and the most likely scenario I can come up with is that they did it on accident ... which — if true — I bet they'd appreciate hearing about that.

So, you've got me starting to rethink the current behavior... that we'd rather error out on a duplicate keyword. 🤔 

Thoughts?

pkg/validations/validate.go
pivotaljohn: yeah! wooops!

pkg/validations/validate.go
pivotaljohn: excellent feedback. I see... `nullAllowed` has no apparent connection to the validation...  yeah, while there's a double negative (which is a little harder to reason) it's better than a full-on level of indirection. 👍🏻 

How about this?

```go
func (v validationKwargs) shouldValidate(value starlark.Value, thread *starlark.Thread) (bool, error) {
	_, valueIsNull := value.(starlark.NoneType)
	whenNullSkip := v.whenNullSkip != nil && *v.whenNullSkip
	if !v.notNull && whenNullSkip && valueIsNull {
		return false, nil
	}
```

pkg/validations/validate.go
pivotaljohn: It's a good question.

A big driver for `min` and `max` to become `starlark.Value`s is that many (but not all) types in Starlark can be compared: we needed the wider type, there.
`min_len` and `max_len` _only_ max sense as an integer.

But to your point, we _could_ store these values as `starlark.Int`s... which have a convenient accessor to the underlying int64 value (for the usages you pointed at).

Let's do that!

pkg/validations/validate.go
pivotaljohn: Okay... conversation below about storing these values as `starlark.Int` obviates this comment all together. Let's delete.

pkg/validations/assert.go
pivotaljohn: I'd say it's well-formed because it's syntactically correct.

However, it just won't work in practice. I have an assumption that library authors will test their business... and will discover that there's no set of data values that can satisfy all these rules.

pkg/validations/validate.go
vmunishwar: Done. Good catch, Garrett!

vmunishwar:
To be done later - review tests for duplicate coverage; every test represents maintenance, let's make sure they are doing their work.

pkg/yttlibrary/assert.go
```diff
@@ -8,26 +8,51 @@ import (
...
+	members := starlark.StringDict{
+		"equals": starlark.NewBuiltin("assert.equals", core.ErrWrapper(m.Equals)),
+		"fail":   starlark.NewBuiltin("assert.fail", core.ErrWrapper(m.Fail)),
+		"try_to": starlark.NewBuiltin("assert.try_to", core.ErrWrapper(m.TryTo)),
+	}
+	if experiments.IsValidationsEnabled() {
+		members = starlark.StringDict{
+			"equals":       starlark.NewBuiltin("assert.equals", core.ErrWrapper(m.Equals)),
+			"fail":         starlark.NewBuiltin("assert.fail", core.ErrWrapper(m.Fail)),
+			"try_to":       starlark.NewBuiltin("assert.try_to", core.ErrWrapper(m.TryTo)),
```
cppforlife: instead of overriding members entirely, we should just be able to set new keys (keeping baseline the same)

pkg/validations/assert.go
```diff
@@ -71,45 +76,109 @@ func NewValidationFromValidationAnnotation(annotation template.NodeAnnotation) (
...
 		message, ok := ruleTuple[0].(starlark.String)
 		if !ok {
 			return nil, fmt.Errorf("expected first item in the 2-tuple to be a string describing a valid value, but was %s (at %s)", ruleTuple[0].Type(), annotation.Position.AsCompactString())
 		}
-		lambda, ok := ruleTuple[1].(starlark.Callable)
+
+		assertion, ok := ruleTuple[1].(starlark.Callable)
 		if !ok {
-			return nil, fmt.Errorf("expected second item in the 2-tuple to be an assertion function, but was %s (at %s)", ruleTuple[1].Type(), annotation.Position.AsCompactString())
+			var err error
```
cppforlife: nit: unnecessary decl

pkg/validations/assert.go
```diff
@@ -71,45 +76,109 @@ func NewValidationFromValidationAnnotation(annotation template.NodeAnnotation) (
...
+			case starlark.Bool:
+				if v {
+					processedKwargs.oneNotNull = v
+				} else {
+					return validationKwargs{}, fmt.Errorf("one_not_null= cannot be False")
+				}
+			case starlark.Sequence:
+				processedKwargs.oneNotNull = v
+			default:
+				return validationKwargs{}, fmt.Errorf("expected True or a sequence of keys, but was a \"%s\"", value[1].Type())
```
cppforlife: nit: we've typically used single quotes for user controlled values.

pkg/validations/validate.go
```diff
@@ -125,24 +133,90 @@ func (v *NodeValidation) DefaultNullSkipTrue() {
 // shouldValidate uses validationKwargs and the node's value to run checks on the value. If the value satisfies the checks,
 // then the NodeValidation's rules should execute, otherwise the rules will be skipped.
 func (v validationKwargs) shouldValidate(value starlark.Value, thread *starlark.Thread) (bool, error) {
-	_, isNone := value.(starlark.NoneType)
-	if isNone && (v.whenNullSkip != nil && *v.whenNullSkip) {
+	_, valueIsNull := value.(starlark.NoneType)
+	whenNullSkip := v.whenNullSkip != nil && *v.whenNullSkip
+	if !v.notNull && whenNullSkip && valueIsNull {
```
cppforlife: todo do we really need `!v.notNull` since valueIsNull==true

pkg/validations/validate.go
```diff
@@ -125,24 +133,90 @@ func (v *NodeValidation) DefaultNullSkipTrue() {
...
 		return false, nil
 	}
 
-	if v.when != nil {
-		result, err := starlark.Call(thread, *v.when, starlark.Tuple{value}, []starlark.Tuple{})
+	if !(v.when == nil || reflect.ValueOf(v.when).IsNil()) {
```
cppforlife: nit: more readable with ! + &&

pkg/validations/validate.go
```diff
@@ -125,24 +133,90 @@ func (v *NodeValidation) DefaultNullSkipTrue() {
...
+		result, err := starlark.Call(thread, v.when, starlark.Tuple{value}, []starlark.Tuple{})
 		if err != nil {
 			return false, err
 		}
 
 		isTrue := bool(result.Truth())
-		return isNone || isTrue, nil
+		return valueIsNull || isTrue, nil
```
cppforlife: valueIsNull? are we saying we should always validate for null values.

pkg/validations/validate.go
```diff
@@ -125,24 +133,90 @@ func (v *NodeValidation) DefaultNullSkipTrue() {
...
 	}
 	// if no kwargs then execute rules
 	return true, nil
 }
 
+func (v validationKwargs) convertToRules() []rule {
+	var rules []rule
+
+	if v.minLength != nil {
+		a := yttlibrary.NewAssertMinLen(*v.minLength)
```
cppforlife: nit: inline assertion constuctor

pkg/yttlibrary/assert.go
```diff
@@ -101,3 +129,289 @@ func (b assertModule) TryTo(thread *starlark.Thread, f *starlark.Builtin, args s
 	}
 	return starlark.Tuple{retVal, starlark.None}, nil
 }
+
+// Assertion encapsulates a rule (a predicate) that can be accessed in a Starlark expression (via the "check" attribute)
+// or in Go (via CheckFunc()).
+type Assertion struct {
```
cppforlife, line 129: lets add ConversionHint

pkg/validations/assert.go
pivotaljohn: In fact, the companion variable, `assertion` has a wider scope so this `var` is at least necessary, if not undesirable.

The scope of `err`, here, is only within this `if`, so seems appropriate.

pkg/validations/assert.go
pivotaljohn: There were a couple other places where this slipped in as well; fixed those too.

pkg/validations/validate.go
pivotaljohn: Turns out: we do.

`!v.notNull` should be read, "is null allowed?" ... so we should only entertain skipping validating this value if null is allowed.

introduced a variable to improve the readability, here.

pkg/validations/validate.go
pivotaljohn: ... and remaining `a` expanded to the more selectable, `assertion`. 👍🏻

Issue #682, opened 2022-06-02
Better handling of HTTP-sourced files


pivotaljohn:
**Describe the problem/challenge you have**
As described here: https://github.com/vmware-tanzu/carvel-kapp/issues/521#issuecomment-1145323232 ...

Whenever an HTTP GET is performed, if the response succeeds but is a non-200 result, that response is being read as the contents of the named fail.

**Describe the solution you'd like**
Whenever there is a non-200 response from the HTTP resource, stop and report _that error_ instead of processing the response given (which will never be the desired resource).

**Anything else you would like to add:**
This is the _exact_ same issue fixed by: vmware-tanzu/carvel-kapp#523

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

PR #686, opened 2022-06-13
Return error on non-200 responses

vicmarbev:
Handle http errors as specified in #682. The resolution is the same as in https://github.com/vmware-tanzu/carvel-kapp/pull/523

### DIFF
--- a/pkg/files/sources_test.go
+++ b/pkg/files/sources_test.go
@@ -1,7 +1,7 @@
 // Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-package files
+package files_test
 
 import (
 	"bytes"
@@ -11,6 +11,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"github.com/vmware-tanzu/carvel-ytt/pkg/files"
 )
 
 func TestHTTPFileSources(t *testing.T) {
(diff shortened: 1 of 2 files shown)


vicmarbev:
The linter fails with:
`pkg/files/sources.go:91:1: exported: exported function NewHTTPSource should have comment or be unexported (revive)
func NewHTTPSource(path string) HTTPSource { return HTTPSource{path, &http.Client{}} }`
but this function already existed. Should I add a comment?

cppforlife:
> Should I add a comment?

yeah, linter is a bit too strict.

gcheadle-vmware (approved):
I've read through the referenced issue in kapp, and can confirm that these changes are identical to the change made in the kapp repo. Everything looks good to me 👍

pkg/files/sources_test.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package files
```
praveenrewar: package name should be files_test (should have `_test` suffix)

vicmarbev: Thanks!

praveenrewar (approved):
LGTM!

pivotaljohn:
Thanks, @vicmarbev!

pivotaljohn:
Published in [v0.42.0](https://github.com/vmware-tanzu/carvel-ytt/releases/tag/v0.42.0).

PR #690, opened 2022-06-20
Add math ytt module

### LINKED ISSUE #689: Add support for a math module

We are using YTT to generate the configuration for our robot. Robotics often need to compute rotation, angles, etc... so we often have to use pi, and sometimes cos, sin, etc... However, there are no such functions in YTT.

Add support for basic math functions :
```yaml
#@ load("@ytt:math", "cos", "pi")
cos_45_degrees: #@ cos(45 * pi / 180)
```

starlark-go already defines a ["math" module](https://github.com/google/starlark-go/blob/master/lib/math/math.go). It should be very straight-forward to support.

---

PRs:
- [ ] #690 
- [ ] vmware-tanzu/carvel#465

---
Vote on this request

Hugal31:
Add the math module. Can be used like this:

```yaml
#@ load("@ytt:math", "math")
ceil: #@ math.ceil(42.2)
log: #@ math.log(1024, 2)
```

**WARNING**: The content of the math module is mostly copy-pasted from [Google's last version of starlark-go](https://github.com/google/starlark-go/blob/master/lib/math/math.go) (the upstream of k14s/starlark-go). In result, I let the original copyright notice at the top of the math.go file, which is probably not the right thing to do.

Contributes to #689

### DIFF
--- a/pkg/template/core/starlark_value.go
+++ b/pkg/template/core/starlark_value.go
@@ -57,6 +57,17 @@ func (e StarlarkValue) AsInt64() (int64, error) {
 	return 0, fmt.Errorf("expected starlark.Int")
 }
 
+// AsFloat64 converts a Starlark number (either int or float) to the corresponding Go double-precision float.
+func (e StarlarkValue) AsFloat64() (float64, error) {
+	switch e := e.val.(type) {
+	case starlark.Int:
+		return float64(e.Float()), nil
+	case starlark.Float:
+		return float64(e), nil
+	}
+	return 0, fmt.Errorf("expected float value, but was %T", e.val)
+}
+
 func (e StarlarkValue) asInterface(val starlark.Value) (interface{}, error) {
 	if obj, ok := val.(UnconvertableStarlarkValue); ok {
 		return nil, fmt.Errorf("Unable to convert value: %s", obj.ConversionHint())
--- a/pkg/yttlibrary/all.go
+++ b/pkg/yttlibrary/all.go
@@ -40,6 +40,7 @@ func NewAPI(replaceNodeFunc tplcore.StarlarkFunc, dataMod DataModule,
 
 	std := map[string]starlark.StringDict{
 		"assert": AssertAPI,
+		"math":   MathAPI,
 		"regexp": RegexpAPI,
 
 		// Hashes
--- a/pkg/yttlibrary/math.go
+++ b/pkg/yttlibrary/math.go
@@ -1,0 +1,273 @@
+// Copyright 2022 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Copyright 2021 The Bazel Authors. All rights reserved.
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+// 1. Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the
+//    distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its
+//    contributors may be used to endorse or promote products derived
+//    from this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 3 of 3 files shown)


pkg/yttlibraryext/all.go
```diff
@@ -10,5 +10,6 @@
 package yttlibraryext
 
 import (
+	_ "github.com/vmware-tanzu/carvel-ytt/pkg/yttlibraryext/math" // include math
```
cppforlife: since math does not bring external dependencies, we should probably use "regular" yttlibrary import mechanism: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/yttlibrary/all.go#L41

pkg/yttlibraryext/math/math.go
```diff
@@ -0,0 +1,245 @@
+// Copyright 2021 The Bazel Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license.
```
pivotaljohn:
```suggestion
// Copyright 2022 VMware, Inc.
// SPDX-License-Identifier: Apache-2.0
//
// Copyright 2021 The Bazel Authors. All rights reserved.
// Redistribution and use in source and binary forms, with or without
// modification, are permitted provided that the following conditions are
// met:
//
// 1. Redistributions of source code must retain the above copyright
//    notice, this list of conditions and the following disclaimer.
//
// 2. Redistributions in binary form must reproduce the above copyright
//    notice, this list of conditions and the following disclaimer in the
//    documentation and/or other materials provided with the
//    distribution.
//
// 3. Neither the name of the copyright holder nor the names of its
//    contributors may be used to endorse or promote products derived
//    from this software without specific prior written permission.
//
// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
// HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
```

Rationale: preserve the original copyright notice as-is per the BSD terms. Prepend the VMware copyright with the Apache license indicator to clarify the contribution will be governed under _that_ license.

pkg/yttlibraryext/all.go
pivotaljohn: Agreed. Not well-documented: that "extensions" are _only_ intended for modules that may be opt'ed out of because they bring a potentially unwanted dependency for those using `ytt` as a Go module.

In addition to the point of "registration" @cppforlife noted, above, an example of another "standard library" module that uses only Golang dependencies: https://github.com/vmware-tanzu/carvel-ytt/blob/develop/pkg/yttlibrary/ip.go

pivotaljohn (changes requested):
Thanks for jumping on this, @Hugal31.

Within _this_ review, both a suggestion about what to do about the license notices for this file... _and_ additional details around how to "register" this new module with the "standard library".

pkg/yttlibraryext/all.go
Hugal31: > Not well-documented

It's written just above, I missed it :|

pivotaljohn: > It's written just above, I missed it :|

Given the level of detail that you are addressing, I'd say that signal is not strong enough.

Re-reading that paragraph, it looks like a wall-of-text to me. Even just a little formatting and tightening some word-choice might make a difference.

I'll attempt this in a separate PR.

pkg/template/core/starlark_value.go
```diff
@@ -57,6 +57,17 @@ func (e StarlarkValue) AsInt64() (int64, error) {
 	return 0, fmt.Errorf("expected starlark.Int")
 }
 
+// Accept either int or float starlark values.
```
pivotaljohn:
```suggestion
// AsFloat64 converts a Starlark number (either int or float) to the corresponding Go double-precision float.
```

Our lint config grandfather's in code we wrote prior to our starting to use the linter. This is why existing code is not GoDoc'ed, but the linter is complaining about this function.

It's a little aggressive, but on balance it's leading to a more readable codebase.

pivotaljohn (changes requested):
Just a tweak to the GoDoc and we're lookin' good, here.

pivotaljohn:
Apologies for the delay... We've been wrestling with the impacts to `ytt`'s guarantee to deterministic/consistent results.  The underlying library does not guarantee consistent results across architectures (as is typical of floating point math).

Wrapping that discussion up.

But in the mean time, this is all looking fit.

pivotaljohn (approved):

pivotaljohn:
Published in [v0.42.0](https://github.com/vmware-tanzu/carvel-ytt/releases/tag/v0.42.0). Many thanks, @Hugal31.

Issue #687, opened 2022-06-14
Bad error message and process panic when using exported library functions.


GrahamDumpleton:
**What steps did you take, what happened and what did you expect:**

Create a single level library structure.

```
.
├── default.yaml
├── _ytt_lib
│   └── mylib
│       └── local.star
```

In `default.yaml` add:

```
#@ load("@ytt:library", "library")
#@ load("@ytt:struct", "struct")
#@ load("@ytt:template", "template")
#@ mylib = library.get("mylib")
#@ func = mylib.export("func")
_: #@ template.replace(func(struct.make()))
```

In `_ytt_lib/mylib/local.star` add:

```
def func(obj):
  return {"xxx": 1}
  # return {"xxx": obj.xxx}
  # xxx
end
```

Run `ytt -f .` and you get the output:

```
xxx: 1
```

Now change `_ytt_lib/mylib/local.star` to:

```
def func(obj):
  # return {"xxx": 1}
  # return {"xxx": obj.xxx}
  xxx
end
```

Run `ytt -f .` and you get the output:

```
ytt: Error:
- library.export: Exporting from library 'mylib': Evaluating starlark template:
    in <toplevel>
      default.yaml:5 | #@ func = mylib.export("func")

    reason:
     - undefined: xxx
         4 |   xxx
```

So as expected get nice error message about what the problem is:

Now change `_ytt_lib/mylib/local.star` to:

```
def func(obj):
  # return {"xxx": 1}
  return {"xxx": obj.xxx}
  # xxx
end
```

Run `ytt -f .` and you get the output:

```
panic: Expected to find compiled template: Expected to find '_ytt_lib/mylib/local.star' compiled template

goroutine 1 [running]:
github.com/vmware-tanzu/carvel-ytt/pkg/template.CompiledTemplateMultiError.buildPos({{0x0, 0x0, 0x0}, {0x166a9a8, 0xc000100f60}}, {0xc0001dcd90, 0x0, 0x0})
	github.com/vmware-tanzu/carvel-ytt/pkg/template/compiled_template_error.go:161 +0x39d
github.com/vmware-tanzu/carvel-ytt/pkg/template.CompiledTemplateMultiError.buildEvalErr({{0x0, 0x0, 0x0}, {0x166a9a8, 0xc000100f60}}, 0xc0001db650)
	github.com/vmware-tanzu/carvel-ytt/pkg/template/compiled_template_error.go:144 +0x125
[...]

Issue #156, opened 2020-06-18
[library] panic when exported function contains a syntax error


cppforlife:
https://get-ytt.io/#gist:https://gist.github.com/cppforlife/b9d54ff11278a4e1cfe5d2c8c96f62ca

```
error: panic: Expected to find compiled template: Expected to find '_ytt_lib/app/config.lib.yml' compiled template

goroutine 1 [running]:
github.com/k14s/ytt/pkg/template.CompiledTemplateMultiError.buildPos(0x0, 0x0, 0x0, 0xb11560, 0xc000092a00, 0xc000132940, 0x1a00000006, 0x0, 0x0, 0x0, ...)
	github.com/k14s/ytt@/pkg/template/compiled_template_error.go:157 +0x467
github.com/k14s/ytt/pkg/template.CompiledTemplateMultiError.buildEvalErr(0x0, 0x0, 0x0, 0xb11560, 0xc000092a00, 0xc00013a9f0, 0x0, 0x0, 0x0, 0x0, ...)
	github.com/k14s/ytt@/pkg/template/compiled_template_error.go:140 +0x105
github.com/k14s/ytt/pkg/template.NewCompiledTemplateMultiError(0xb0d4c0, 0xc00013a9f0, 0xb11560, 0xc000092a00, 0x0, 0x0)
	github.com/k14s/ytt@/pkg/template/compiled_template_error.go:53 +0x7e4
github.com/k14s/ytt/pkg/template.(*CompiledTemplate).Eval(0xc000064b40, 0xc000092cd0, 0xb11560, 0xc000092a00, 0xc000092640, 0xc000092cd0, 0xc0000fb740, 0x20, 0x8fcda0)
	github.com/k14s/ytt@/pkg/template/compiled_template.go:120 +0x70e
github.com/k14s/ytt/pkg/workspace.(*TemplateLoader).EvalYAML(0xc000092a00, 0xc000092690, 0xc000092690, 0xc000092640, 0xc00000e1a8, 0x1, 0x1, 0xc00009f650)
	github.com/k14s/ytt@/pkg/workspace/template_loader.go:171 +0x686
github.com/k14s/ytt/pkg/workspace.(*LibraryLoader).eval(0xc00014ba68, 0xc00001f040, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	github.com/k14s/ytt@/pkg/workspace/library_loader.go:140 +0xac7
[...]

PR #693, opened 2022-06-21
Stub unavailable source when reporting Starlark errors

pivotaljohn:
Addresses:
- #156
- #687

### DIFF
--- a/pkg/template/compiled_template_error.go
+++ b/pkg/template/compiled_template_error.go
@@ -5,6 +5,7 @@ package template
 
 import (
 	"fmt"
+	"reflect"
 	"regexp"
 	"strings"
 
@@ -152,16 +153,18 @@ type sourceCode interface {
 	CodeAtLine(pos *filepos.Position) *Line
 }
 
+// unavailableSource is a stub for when the source code line for a filepos.Position cannot be found.
+// This occurs when the position refers to a file that is not part of this Library:
+// - a function from another Library (e.g. via library.get().export())
+// - Starlark sourced from within ytt (e.g. some out-of-the-box validation functions)
 type unavailableSource struct {
 }
 
 func (s unavailableSource) CodeAtLine(pos *filepos.Position) *Line {
 	return &Line{
-		Instruction: Instruction{},
 		SourceLine: &SourceLine{
-			Position:  pos,
-			Content:   "(unavailable)",
-			Selection: nil,
+			Position: pos,
+			Content:  "(unavailable)",
 		},
 	}
 }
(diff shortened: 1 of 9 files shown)


pkg/template/compiled_template_error.go
```diff
@@ -156,32 +174,39 @@ func (e CompiledTemplateMultiError) buildPos(pos syntax.Position) CompiledTempla
 		return CompiledTemplateErrorPosition{}
 	}
 
-	ct, err := e.loader.FindCompiledTemplate(pos.Filename())
+	var sc sourceCode
+	sc, err := e.loader.FindCompiledTemplate(pos.Filename())
```
cppforlife: should we switch away from err to bool (found or not found)?

pkg/template/compiled_template_error.go
```diff
@@ -148,6 +148,24 @@ func (e CompiledTemplateMultiError) buildEvalErr(err *starlark.EvalError) Compil
...
+type sourceCode interface {
+	CodeAtLine(pos *filepos.Position) *Line
+}
+
+type unavailableSource struct {
+}
+
+func (s unavailableSource) CodeAtLine(pos *filepos.Position) *Line {
+	return &Line{
+		Instruction: Instruction{},
```
cppforlife: nit: not necessary since its zero value and not a pointer.

pkg/template/compiled_template_error.go
```diff
@@ -148,6 +148,24 @@ func (e CompiledTemplateMultiError) buildEvalErr(err *starlark.EvalError) Compil
...
+		SourceLine: &SourceLine{
+			Position:  pos,
+			Content:   "(unavailable)",
+			Selection: nil,
```
cppforlife: same as above

pkg/template/compiled_template_error.go
```diff
@@ -156,32 +174,39 @@ func (e CompiledTemplateMultiError) buildPos(pos syntax.Position) CompiledTempla
...
 	if err != nil {
-		panic(fmt.Errorf("Expected to find compiled template: %s", err))
+		// Most likely because the position refers to a file that is not part of this Library:
+		// - a function from another Library
+		// - Starlark sourced from within ytt (e.g. out-of-the-box validation functions)
+		sc = unavailableSource{}
 	}
 
-	line := ct.CodeAtLine(filepos.NewPosition(int(pos.Line)))
+	p := filepos.NewPosition(int(pos.Line))
```
cppforlife: should we add a new constructor that add file and line?

pkg/template/compiled_template_error.go
pivotaljohn: 👍🏻  Avoid using `error`s for flow control. With this PR, it's no longer an error.

pkg/template/compiled_template_error.go
pivotaljohn: 👍🏻  In fact, long overdue.

pkg/template/compiled_template_error.go
pivotaljohn: I decided that `nil` is sufficient in this case (rather than two return values). `nil` is never a valid "found" value.

pkg/template/compiled_template_error.go
```diff
@@ -175,11 +178,9 @@ func (e CompiledTemplateMultiError) buildPos(pos syntax.Position) CompiledTempla
 	}
 
 	var sc sourceCode
-	sc, err := e.loader.FindCompiledTemplate(pos.Filename())
-	if err != nil {
-		// Most likely because the position refers to a file that is not part of this Library:
-		// - a function from another Library
-		// - Starlark sourced from within ytt (e.g. out-of-the-box validation functions)
+	sc = e.loader.FindCompiledTemplate(pos.Filename())
+	if reflect.ValueOf(sc).IsNil() {
```
pivotaljohn, line 182: required idiom as `sc` is an interface, not a pointer.

PR #696, opened 2022-06-27
Add one_not_null to assert module and validation kwarg

gcheadle-vmware:


pivotaljohn:
Cherry-picked these commits into #677

PR #698, opened 2022-06-30
add godoc comments in template package

gcheadle-vmware:


pivotaljohn (approved):
It's hard to imagine a more inclusive parting contribution than to take the time to share your learning in a format and forum that will serve many to come.

These are great adds, @gcheadle-vmware.  🙏🏻

PR #706, opened 2022-07-19
Add `one_of()` / `one_of=` — a value is within an enumeration

### LINKED ISSUE #691: Validate a value is within an enumeration
As a library/package author,
Sometimes I want to limit the range of valid values to a specific and finite set.

To handle cases like this ☝🏻 ,
I want to be able to use out-of-the-box validation rules that generally cover this situation.

👉🏻  In effect, address the [Enumeration](https://github.com/vmware-tanzu/carvel/blob/004-schema-validation/proposals/ytt/004-schema-validation/README.md#use-case-enumeration) use case.

---

As with other OOTB validation rules, there are three APIs:
- to `@ytt:assert` add `one_of()` and `one_of().check()`
- on `@assert/validate`, implement `one_of=`
- to `@schema/validation`, add `one_of=`

Consider:
- Starlark has a [membership test](https://github.com/google/starlark-go/blob/master/doc/spec.md#membership-tests) with the `in` operator.
- Must the values in the enumeration be of the same type?

---
PRs:
- [x] #706 
- [x] vmware-tanzu/carvel#490

pivotaljohn:
Implements #691

### DIFF
--- a/pkg/validations/annotations.go
+++ b/pkg/validations/annotations.go
@@ -24,6 +24,7 @@ const (
 	KwargMax          string = "max"
 	KwargNotNull      string = "not_null"
 	KwargOneNotNull   string = "one_not_null"
+	KwargOneOf        string = "one_of"
 )
 
 // ProcessAssertValidateAnns checks Assert annotations on data values and stores them on a Node as Validations.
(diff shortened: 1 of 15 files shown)


pkg/validations/filetests/one_of=/invalid-when-value-not-in-string
```diff
@@ -0,0 +1,7 @@
+#@assert/validate one_of="debug,info,warning,error,fatal"
+foo: critical
```
cppforlife: this seems a bit suspicious. should we be disallowing string in string check? e.g. bad setup:

```
#@assert/validate one_of="not-valid"
foo: valid
```

pivotaljohn: Yeah, bad.  Cool; will pull back "string in string" support.  Simplifies the impl too. 👍🏻

Issue #724, opened 2022-08-29
Schema Validation v2


pivotaljohn:
_(this is the next feature increment of #561)_

---

**As a** Configuration Author
**I want to** declare what constitutes a _valid_ Data Value right in my schema definition
**In a way that** requires even less custom Starlark code from me ⭐ 
**And** I can write fewer annotations ⭐  because `ytt` can confidently infer for what _have_ written.
*So that**
- I can state those invariants in one central place and know they will be respected throughout the execution
- those who use my `ytt` library and violate one of those variants will receive a helpful error message
- and my overall maintenance effort/costs are less than before. ⭐ 

_(⭐ = a desired improvement on top of the MVP.)_

[[proposal]](https://github.com/vmware-tanzu/carvel/pull/331/files?short_path=a81faf6#diff-a81faf6d33e82e22eee1083ee484a23eda8a928b046116a2cb516e58086984b8)
[[getting started]](https://carvel.dev/blog/ytt-validations-preview/)

---
### Backlog
- _⚠️ This track of work is still being elaborated; some scope is likely missing.⚠️_
- loads of ideas for improvements can be mined from #707's [synthesis](https://miro.com/app/board/uXjVOgHIamo=/)

UX:
- [ ] Use custom violation messages
  - on `@assert/validate`, implement 3-tuple rules where the third tuple is the violation template; to `@schema/validation`, add 3-tuple tules.
  - note: this is critical to improve the DX around unit testing (detailed in [a conversation on #741](https://github.com/vmware-tanzu/carvel-ytt/pull/741#discussion_r969654801))
- [ ] disallow `@schema/validation` in non-schema documents.
- [ ] disallow `@assert/validate` in schema documents.

Fast Follow: other must-have OOTB rules
- [ ] `matches=`

Programmatic Validation of YAML
- [ ] Validate a YAML document that contains `@assert/validate` annotations
  -  expose `assert.is_valid()` as a method in `@ytt:assert` module

Include Validations in Schema Inspect
- [ ] Include validations when exporting schema to OpenAPI v3 (subset: one-for-ones)
  - need: https://kubernetes.slack.com/archives/CH8KCCKA5/p1661186311752459?thread_ts=1661185491.294779&cid=CH8KCCKA5
- [ ] Include validations when exporting schema to OpenAPI v3 (subset: needing translation)

Authoring Conveniences:
- [ ] Set a default validation for descendent Data Values (`@schema/validate-string-default`)
[...]

PR #741, opened 2022-09-10
Allow when= to access value, parent, and root nodes

### LINKED ISSUE #733: Allow access to contained document for conditional validations
**As a** Configuration Author,
**I want** the option to reference the final Data Values within my logic for determining if the validation should be run
**So that** I can succinctly express intra-document dependencies with minimal cognitive load for me, those who read my schema, and those who maintain them.

For example (simplified from original):
```yaml
#@data/values-schema
---
credential:
  useDefaultSecret: true
  #@schema/nullable
  #@schema/validation not_null=True, when=lambda parent: parent["useDefaultSecret"]
  secretContents:
    cloud: ""
backupStorageLocation:
  spec:
    #@schema/nullable
    #@schema/validation not_null=True, when=lambda doc: not doc["credential"]["useDefaultSecret"]
    existingSecret: ""
```

Here:
- `credential.secretContents` is `null`, by default, but must be specified if `credential["useDefaultSecret"]` is `true`
- `backupStorageLocation.spec.existingSecret`, likewise is `null` by default; but must be specified if `credential["useDefaultSecret"]` is `false`.
- if there is also a `when=` condition on a validation, _both_ `when=` and `when_doc=` must execute successfully and produce a `True` result; if either is `False`, do short-circuit running any validation rules.

Suggested tasks:
- [x] implement the `when=` 
- [x] document this keyword in the `@schema/validation` reference
- [x] document the use of this keyword in the "Writing Validations" guide
  - "Conditional Validations" — mention use/trade-offs
  - "Mutually Exclusive Sections" — as an alternative to `one_not_null=` for when removing the discriminator is a breaking change.
- [x] update the ytt Validations Proposal to include this keyword and rationale
  - anywhere `when=` is mentioned is a candidate; for example the [Use Case: Conditional Validations](https://github.com/vmware-tanzu/carvel/blob/004-schema-validation/proposals/ytt/004-schema-validation/README.md#use-case-conditional-validations) section.

## Context

During usability testing (#707), vetting the feature set included in this MVP was determined to be infeasibly difficult to port _existing_ validation code to the `ytt` Data Values Validation mechanism.

In Tanzu Community Edition packages, alone, multiple validation rules are based on values in semantically different branches of the data values schema:
- https://github.com/vmware-tanzu/community-edition/blob/9d91b329f0ef05b0b261ca5b43dba2572c163fd7/addons/packages/velero/1.6.3/bundle/config/values.star#L64-L72
- https://github.com/vmware-tanzu/community-edition/blob/cc7eb24b0e13875a06ac8578cd73f83a217ec4d9/addons/packages/pinniped/0.4.4/bundle/config/values.star#L15-L18
- https://github.com/vmware-tanzu/community-edition/blob/cc7eb24b0e13875a06ac8578cd73f83a217ec4d9/addons/packages/harbor/2.2.3/bundle/config/values.star#L48-L52
- https://github.com/vmware-tanzu/community-edition/blob/cc7eb24b0e13875a06ac8578cd73f83a217ec4d9/addons/packages/pinniped/0.4.4/bundle/config/values.star#L42-L51

All of these cases _could_ be partially address by hoisting the validation rules to the lowest common ancestor node in the YAML document. However, in practice this results in:
- the logic for validation a Data Value is hoisted significantly some distance from that Data Value;
- there is only one option for expressing conditional logic within the annotation (i.e. via the `when=` keyword argument) and so for these cases, multiple rules would need to include situation-specific short-circuiting in order to meet feature parity of what's currently present;
- requiring the Author to write custom rules (instead of being able to rely on the "named" rules), pushing this kind of configuration towards being more complex.

Illustrating with the given example, above; with the current feature set, this is how that logic would be expressed:

```yaml
#@data/values-schema
#@schema/validation ("backupStorageLocation.spec.existingSecret is specified", lambda value: value["backupStorageLocation"]["spec"]["existingSecret"] != None), when=lambda value: not value["credential"]["useDefaultSecret"]
---
#@schema/validation ("secretContents is specified", lambda value: value["secretContents"] != None), when=lambda value: value["useDefaultSecret"]
credential:
  useDefaultSecret: true
  #@schema/nullable
  secretContents:
    cloud: ""
backupStorageLocation:
  spec:
    #@schema/nullable
    existingSecret: ""
```

... all of which represents interface/usability complexity hoisted on the Author.

However, by providing access to the contained document, Authors are able to:
- express these conditions within a validation directly on the Data Value being verified;
- apply generalized rules without modification or adaptation on said Data Values.

... all of which amounts to a much more maintainable schema.

As a trade-off, Authors will now have the ability to hard-code locations of values within the Data Values document; when the document is reorganized, all of those references must be updated as well.
As illustrated above, this explicit dependency will exist regardless of whether the reference occurs at the root of the document or within a sub-node.

pivotaljohn:
Resolves #733


pkg/validations/validate.go
```diff
@@ -163,30 +192,54 @@ func (v validationKwargs) shouldValidate(value starlark.Value, thread *starlark.
...
+			case "v", "val", "value":
+				kwargs = append(kwargs, starlark.Tuple{starlark.String(name), value})
+			case "p", "par", "parent":
+				kwargs = append(kwargs, starlark.Tuple{starlark.String(name), parent})
+			case "r", "root", "d", "doc", "document":
+				// When validation support is extended beyond Data Values and into documents in general, differentiate
+				//   between "root" and "document":
+				//   - since data values validation occurs on a document, "root" and "document" are synonyms, right now.
+				//   - once validations can be defined in templates, then "root" will be a DocSet.
+				kwargs = append(kwargs, starlark.Tuple{starlark.String(name), root})
```
cppforlife: ive been chewing on this a bit more today... care to do a quick zoom with me to talk it through?

pkg/yamltemplate/filetests/ytt-library/assert/max_len/value-fails-assertion.tpltest
```diff
@@ -4,7 +4,7 @@ max_len_of_string: #@ assert.max_len(2).check("abc")
 +++
 
 ERR:
-- fail: length of 3 is more than 2
+- fail: length = 3
```
cppforlife, line 7: i feel like we lost too much of info of how to fix the problem (there might be a good reason for it though... lets talk).

also, i wonder if we can incorporate validation name here? (imagine if `assert.max_len(2)` was turned into a var then even with the source line it would be pretty hard to tell what validation is failing)

pkg/validations/validate.go
pivotaljohn: I would love to! I'll reach out.

pkg/yamltemplate/filetests/ytt-library/assert/max_len/value-fails-assertion.tpltest
pivotaljohn: Yeah, appreciate the attention to this. It's been challenging balancing the primary output (in the form of OOTB/named/keyword versions of the rules) from these more programmatic-based cases.

Definitely appreciate a space to talk through the trade-offs. 👍🏻

pkg/validations/validate.go
pivotaljohn: Summary:
- chief concern is that nowhere else in `ytt` is there an interaction like this: where the name of the argument of a supplied function has meaning outside of what the author wants.
- we landed on
  - one required positional parameter; it will receive the value being validated
  - a second optional positional parameter; when specified will be populated with a `struct` that has two attributes: `parent` and `root`.

pkg/yamltemplate/filetests/ytt-library/assert/max_len/value-fails-assertion.tpltest
pivotaljohn: Key concern here is that the output of the `check(v)` function, here, looks nothing like what one would expect when using this function in an assertion context outside of validations (e.g. in a unit test).

Of significant concern is making the experience of the end-user reading output from a failing validation at readable as possible... but not at the serious detriment of a package author's use of an assertion library for other uses.

Still chewing on this, but the going idea is to include _two_ functions: one that specifically works well with validations, another that is useful for an assertion result.

pivotaljohn: Slept on it.

TL;DR. We'll defer. We have a good quality UX for our primary proto-persona (Consumers). The sub-par experience for our secondary group (Authors) points to a missing feature (which, when implemented ought to resolve the  design flaw) that can be done later without introducing a breaking change.

We deliberately deferred the feature of allowing the author of a rule to customize the "violation template" (i.e. the templated error message, described in the [proposal](https://github.com/vmware-tanzu/carvel/blob/004-schema-validation/proposals/ytt/004-schema-validation/README.md#specification) as the `message` argument). However, in our implementation, we failed to make room for this customization. This comes to a head as we attempt to come up with a violation message that serves _both_ the Configuration Consumer (who should see a message containing only the actuals when the rule fails since the overall message already contains the expected) and the Configuration Author (who would rightly expect the failure message to _include_ the expectation).

Given that:
- we want to release this feature as generally available this week;
- the gap in the DX in the Configuration Author case has a reasonable workaround (i.e. that they include the expectation in their test error messages, themselves);
- the feature that would resolve this tension is already defined;
- that this situation can be resolved without introducing a breaking change...

We'll accept this current implementation, as is; with the commitment to improve the DX for Configuration Authors, [soon](https://github.com/vmware-tanzu/carvel-ytt/issues/724).

PR #745, opened 2022-09-17
Allow template tests to have trailing empty lines

mamachanko:
This makes it easier to author template tests, because editors tend to append a newline when saving files.

Previously, `ERR:` template test could not have trailing empty lines. I am using both _Vim_ and _IntelliJ_. Both make it hard to remove the last line break in the file. I spend a minute to figure it out, and in _Vim_ you can do `:set noendofline binary` for a buffer and it trims trailing whitespace. But that's a bother.

I think it is better testing UX this way. I held back from letting my IDE add a line break to existing `*.tpltest`.

### DIFF
--- a/test/filetests/filetests.go
+++ b/test/filetests/filetests.go
@@ -1,6 +1,10 @@
 // Copyright 2022 VMware, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+/*
+Package filetests houses a test harness for evaluating templates and asserting
+the expected output.
+*/
 package filetests
 
 import (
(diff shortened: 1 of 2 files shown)


mamachanko:
@pivotaljohn thank you so much for taking action and addressing the linter-flagged issues! The failing PR checks were overlooked by me. 🙇🏻 

Furthermore, I forgot to mention that `test/filetests/filetests_test.go` tests both for spaces and tabs, but GitHub doesn't display whitespace characters sufficiently different (which is what one wants most of the time), e.g.

```go
├──┤├──┤├──┤give:·`␊
we␊
␊
··want├──┤␊
├──┤yaml␊
␊
`,␊
├──┤├──┤├──┤want:·`␊
we␊
␊
··want␊
├──┤yaml`,␊
```

pivotaljohn (approved):
With this, you've eliminated one of the last (the last?) vestiges of pain around writing these kinds of tests. That's important because this style of testing _ought_ to be the workhorse for this codebase.

🙏🏻

PR #770, opened 2022-11-29
Bump go version to latest 1.19.3

vmunishwar:


pivotaljohn (approved):
Solid. 👍🏻 

If you want to, might consider squashing into a single commit since the all do go together.

PR #771, opened 2022-11-30
Bump Go to v1.19.3

praveenrewar:


pivotaljohn (changes requested):
The center of the PR looks good.

However, it includes changes that are not relevant to the version bumps. We _aspire_ to create changesets that are about the one change (doesn't always happen, but we aim for it).

Specifically:
- It appears that everything under `./pkg` is reformatted code (and the scope of it suggests that a tool might have done that reformatting, automatically). If you'd like to make this reformatting, let's pull that into a separate PR, please.
- there's an edit to the `go.sum` which also appears to be irrelevant (again, looks like some tool or automation made the change). Let's remove that from the changeset.

praveenrewar:
> However, it includes changes that are not relevant to the version bumps. We aspire to create changesets that are about the one change (doesn't always happen, but we aim for it).

Yeah, these changes are the result of running `hack/build.sh` after bumping Go version in `go.mod`. I should probably move that to a separate PR, the only reason I kept it together was thinking that different version of Go in go.mod and other workflows would probably causes issues (it might still, but I will give it a try).

praveenrewar:
@pivotaljohn  I didn't get around to work on this yet, but I do have a question in my mind regarding it 🙋‍♂️
We are planning to make a separate PR for the changes made by `go fmt`. I am just thinking that then these 2 PRs would be closely related to each other where one would depend on the other. I am totally fine with this approach but I just felt that maybe this is an indication that we should have a single PR for it.

pivotaljohn:
> We are planning to make a separate PR for the changes made by go fmt. I am just thinking that then these 2 PRs would be closely related to each other where one would depend on the other. I am totally fine with this approach but I just felt that maybe this is an indication that we should have a single PR for it.

Ah geez! Not sure why it didn't occur to me that upgrading the Go version (and not GoLand or something reformatting) would be the source of the code changes.
I absolutely agree with you @praveenrewar : in the Go world, the version of the language extends to code formatting... and so these are very related changes... and yeah, I feel the same way you do: they have enough affinity that they belong in the same PR.

Along the lines of making our history readable, I _do_ think that it's valuable to keep even `go fmt` formatting changes in a separate _commit_ from the actual Go version bump.
So, the PR gives the merged view: "hey, version bump implies code formatting changes, sometimes" but the individual commit view makes clear what's the intentional change (i.e. go bump) and what's the knock on affects (i.e. code formatting).

Thanks for bringing all this up. 👍🏻 

So, instead of creating a separate PR... just split the bump and formatting into two different commits in _this_ PR, please!

sethiyash:
> So, instead of creating a separate PR... just split the bump and formatting into two different commits in this PR, please!

Make sense @pivotaljohn. As Praveen is OOO this week, I have made the relevant changes to have them in a separate commit.

pivotaljohn (approved):
LGTM!

(Sorry for the delay.)

PR #773, opened 2022-11-30
0.40.x fix CVE

rohitagg2020:
0.40.x fix CVE


pivotaljohn (approved):
Looks good.

One question: the latest version of the `golangci-lint` is 1.50.1; I don't see an inherent problem with selecting an older version, but what's the driver behind that choice?

rohitagg2020:
I tried using `1.50.1` but was getting one error which is weird because the same is working on `develop` it seems. Hence falled back to this version.

PR #776, opened 2022-12-02
Reformatting comments as per go fmt

rohitagg2020:
Reformatting comments as per go fmt. 

It seems like with go `1.19.3`, go fmt has bit of tweak w.r.t. some of the comments.

This is the PR for comments to be on the similar lines as output of go fmt.

### DIFF
--- a/pkg/schema/annotations.go
+++ b/pkg/schema/annotations.go
@@ -59,7 +59,7 @@ type TitleAnnotation struct {
 	pos   *filepos.Position
 }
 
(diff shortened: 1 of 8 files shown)


pivotaljohn:
LGTM!

BTW, @praveenrewar [made a really valid point in a sister PR](https://github.com/vmware-tanzu/carvel-ytt/pull/771#issuecomment-1337904303) that just didn't occur to me: that while the version bump and formatting changes (for history's readability's sake) should be in separate commits, that they prolly go together in a single PR.

Next time.

For now, this seems perfectly fine. 😁 

Thanks for the help!

PR #815, opened 2023-03-24
Bumping go version to 1.19.6 in 0.43.x line

sethiyash:
- Bumping go version to 1.19.6 in 0.43.x line

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -12,10 +12,10 @@ jobs:
     name: lint
     runs-on: ubuntu-latest
     steps:
-      - name: Install Go 1.19.5
+      - name: Install Go 1.19.6
         uses: actions/setup-go@v3
         with:
-          go-version: "1.19.5"
+          go-version: "1.19.6"
       - uses: actions/checkout@v2
         with:
           fetch-depth: '0'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,10 +17,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Install Go 1.19.5
+      - name: Install Go 1.19.6
         uses: actions/setup-go@v3
         with:
-          go-version: "1.19.5"
+          go-version: "1.19.6"
 
       - name: generate website/generate.go
         if: startsWith(github.ref, 'refs/tags/')
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -10,10 +10,10 @@ jobs:
     name: Test All
     runs-on: ubuntu-latest
     steps:
-    - name: Install Go 1.19.5
+    - name: Install Go 1.19.6
       uses: actions/setup-go@v3
       with:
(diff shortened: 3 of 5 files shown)


kencochrane:
1.19.7 is the latest version; should it be bumped to that one instead?

> go1.19.7 (released 2023-03-07) includes a security fix to the crypto/elliptic package, as well as bug fixes to the linker, the runtime, and the crypto/x509 and syscall packages. See the [Go 1.19.7 milestone](https://github.com/golang/go/issues?q=milestone%3AGo1.19.7+label%3ACherryPickApproved) on our issue tracker for details.

rohitagg2020:
LGTM

pivotaljohn:
> 1.19.7 is the latest version; should it be bumped to that one instead?
> 
> > go1.19.7 (released 2023-03-07) includes a security fix to the crypto/elliptic package, as well as bug fixes to the linker, the runtime, and the crypto/x509 and syscall packages. See the [Go 1.19.7 milestone](https://github.com/golang/go/issues?q=milestone%3AGo1.19.7+label%3ACherryPickApproved) on our issue tracker for details.

Hey @kencochrane. Thanks for calling that out. It's a legitimate question.

There's a number of concerns that get folded into selecting a specific dependency (even the version of Go we use). We have the desire to keep the suite in sync on Go version. It's currently some effort to produce new releases for all the tools and any downstream work that the VMware folks are also responsible for.

So, we'll pick a version and trigger a release.  Right now, sometimes that means not getting the absolute latest.

We're investing in our overall automation and making this effort shorter/cheaper. When we get to where we want to be, yeah, we'll be able to consume dependencies at will — as we'd all hope.

Again, thanks for the review; it's great to know folks are looking out, here. 👍🏻

pivotaljohn:
It's my understanding that we are — whenever possible — making such changes on `develop` and then cherry-picking into version branches.

From what I'm seeing this change is not yet on `develop`.

Do I have my wires crossed? Or are we all still sync'ing up on this intended practice?

sethiyash:
@pivotaljohn Thanks for pointing it out. I have missed checking on develop about the latest go version. 

I'm just closing this PR and raising a new PR to bump on develop first and cherry-picking into this branch.

kencochrane:
@pivotaljohn Thank you for explaining that, and I totally get where you are coming from. I mainly brought it up because there are known [security issues](https://github.com/golang/go/issues/58647) with 1.19.6 that are fixed in 1.19.7, and I thought since it requires work to bump the version anyway, it might be a good idea to use a version of go with no known security vulnerabilities. 

When Go binaries are built with a version of Go that has known security vulnerabilities, it triggers security alerts on containers that contain those binaries. If companies want to stay in compliance, they have to fix those scan alerts within 30 to 90 days (depending on the CVE severity). Bumping to 1.19.7 would help those companies clear those alerts in their SLA window without needing another release.

It's your project, so feel free to do what is best. I just wanted to give a little heads-up in case it would have saved you time without having to do the whole process twice (with 1.19.6 and then again with 1.19.7).

Thanks again!

PR #867, opened 2023-08-18
add cii badge to readme.md

microwavables:


README.md
```diff
@@ -38,3 +38,5 @@ Check out which organizations are using and contributing to Carvel: [Adopter's l
 ### Integrating with ytt
 
 If you want to integrate `ytt` within your own tooling, review our [APIs](examples/integrating-with-ytt/apis.md).
+---
```
praveenrewar: nit: add new line before `---` (otherwise the line above it looks really big (https://github.com/microwavables/carvel-ytt/tree/add-cii-badge#integrating-with-ytt)

(I think the same applies to other tools as well 😄 )

microwavables: @praveenrewar yup! I was in the process of updating them all when I saw the error lol but everything should now be correct! Thank you!

praveenrewar (approved):

PR #878, opened 2023-10-20
Update the module name to start using carvel.dev URLs

joaopapereira:
🚨 Only Merge this PR to develop when after we release any outstanding change that are present in that branch 🚨


pkg/yttlibrary/overlay/insert_annotation.go
```diff
@@ -4,11 +4,11 @@
 package overlay
 
 import (
+	"carvel.dev/ytt/pkg/template"
+	tplcore "carvel.dev/ytt/pkg/template/core"
+	"carvel.dev/ytt/pkg/yamltemplate"
```
praveenrewar, line 11: I think these should be below "fmt", probably adding a new line between "fmt" and "github.com/k14s/starlark-go/starlark" should resolve it.

praveenrewar (approved):
LGTM!

pkg/template/evaluation_ctx.go
```diff
@@ -4,9 +4,9 @@
 package template
 
 import (
+	"carvel.dev/ytt/pkg/template/core"
 	"fmt"
```
vmunishwar, line 7: Just a nit: "fmt" should be above "carvel.dev/ytt/pkg/template/core" and separated by a blank line.
```suggestion
	"fmt"
	
	"carvel.dev/ytt/pkg/template/core"
```

vmunishwar (approved):
LGTM. Just an update to import order suggested at 1 place.

PR #891, opened 2024-01-29
Add an optional boolean to not escape html during json encoding

### LINKED ISSUE #890: Add an optional boolean to not escape html during json encoding
By default the golang encode does escape problematic HTML characters inside JSON quoted strings. Which may leads to undesired behavior.

A typical example is when you are using the [argocd-vault-plugin](https://github.com/argoproj-labs/argocd-vault-plugin) syntax inside a json string, which looks like this:

```yaml
#@data/values
---
credentials:
  username: bob
  password: <path:/to/bob/password>
```
Then if I want to create a secret with

```yaml
#@ load("@ytt:data", "data")
#@ load("@ytt:json", "json")
stringData: #@ json.encode(data.values.credentials, indent=2)
```

Then it compiles to:
```yaml
stringData: |-
  {
    "username": "bob",
    "password": "\u003cpath:/to/bob/password\u003e"
  }
```
Which obviously isn't what we want.

It would be nice if we can call [Encode.SetEscapeHTML](https://pkg.go.dev/encoding/json#Encoder.SetEscapeHTML) somehow to prevent that.

WnP:
Fix #890

### DIFF
--- a/pkg/yttlibrary/json.go
+++ b/pkg/yttlibrary/json.go
@@ -4,6 +4,7 @@
 package yttlibrary
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"strings"
@@ -37,6 +38,7 @@ func (b jsonModule) Encode(thread *starlark.Thread, f *starlark.Builtin, args st
 	}
 	allowedKWArgs := map[string]struct{}{
 		"indent": {},
+		"escape_html": {},
 	}
 	if err := core.CheckArgNames(kwargs, allowedKWArgs); err != nil {
 		return starlark.None, err
@@ -49,26 +51,33 @@ func (b jsonModule) Encode(thread *starlark.Thread, f *starlark.Builtin, args st
 	val = orderedmap.Conversion{yamlmeta.NewGoFromAST(val)}.AsUnorderedStringMaps()
 
 	var valBs []byte
+	buffer := bytes.NewBuffer(valBs)
 	indent, err := core.Int64Arg(kwargs, "indent")
 	if err != nil {
 		return starlark.None, err
 	}
-
 	if indent < 0 || indent > 8 {
 		// mitigate https://cwe.mitre.org/data/definitions/409.html
 		return starlark.None, fmt.Errorf("indent value must be between 0 and 8")
 	}
 
-	if indent > 0 {
-		valBs, err = json.MarshalIndent(val, "", strings.Repeat(" ", int(indent)))
-	} else {
-		valBs, err = json.Marshal(val)
-	}
+	escapeHTML, err := core.BoolArg(kwargs, "escape_html", true)
 	if err != nil {
 		return starlark.None, err
 	}
 
-	return starlark.String(string(valBs)), nil
+	encoder := json.NewEncoder(buffer)
+	if indent > 0 {
+		encoder.SetIndent("", strings.Repeat(" ", int(indent)))
+	}
+	encoder.SetEscapeHTML(escapeHTML)
+
+	if err := encoder.Encode(val); err != nil {
+		return starlark.None, err
+	}
+
+	res := strings.TrimSuffix(buffer.String(), "\n")
+	return starlark.String(res), nil
 }
 
 // Decode is a core.StarlarkFunc that parses the provided input from JSON format into dicts, lists, and scalars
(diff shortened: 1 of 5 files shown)


prembhaskal:
checking.
one comment about boolean flag, can we also explore if we can pass additional parameters to the json.encode() method. (just a thought I am not sure if it is feasible to do).

okhoshi:
> one comment about boolean flag, can we also explore if we can pass additional parameters to the json.encode() method. (just a thought I am not sure if it is feasible to do).

There's no other option on the json Encoder, only `SetIndent` and `SetEscapeHTML`.

WnP:
I've fixed the tests and created one for the new feature.

prembhaskal:
> > one comment about boolean flag, can we also explore if we can pass additional parameters to the json.encode() method. (just a thought I am not sure if it is feasible to do).
> 
> There's no other option on the json Encoder, only `SetIndent` and `SetEscapeHTML`.

ignore my previous comment, I think I had a brain fade moment there.

I feel it is better to name the variable as escape_html with the default value as true, which will be inline with what json encoder is doing.

someone intending to not escape html can set it as escape_html=false

pkg/yttlibrary/json.go
```diff
@@ -37,6 +38,7 @@ func (b jsonModule) Encode(thread *starlark.Thread, f *starlark.Builtin, args st
 	}
 	allowedKWArgs := map[string]struct{}{
 		"indent": {},
+		"no_escape_html": {},
```
prembhaskal: parameter with name "escape_html" will be less confusing.
Its default value can be true, same as in json encoder.

WnP: I would like too, but `core.BoolArg` return `false` by default:

https://github.com/carvel-dev/ytt/blob/b4d0ff06a83291341b06da166dfcb147baee1665/pkg/template/core/args.go#L12-L23

So it's impossible to say if a returned `false` value is explicitly requested by the user or is the default value, negate the value would not help neither.

The only way to implement it would be to either:
- re-wrote `core.BoolArg` to accept a default value.
- re-wrote `core.BoolArg` to return `nil` if the user doesn't pass a value, this would require to return a `*bool` instead of a `bool`.
- write another function `core.BoolArg` with a different signature to accept a default value... which obviously would be `true` because we already have one which return `false` by default, so maybe just provide another function which return `true` by default would be enough.

Which one of those solution do you prefer to be implemented? Or have you another idea on the topic?

Whatever your choice will be I'll be happy to implement it.

WnP: During that time I'm implementing the first solution, which I think is best.

WnP: Feel free to resolve this conversation if the implementation fit your expectations.

pkg/yttlibrary/json.go
```diff
@@ -49,26 +51,33 @@ func (b jsonModule) Encode(thread *starlark.Thread, f *starlark.Builtin, args st
...
 	}
-	if err != nil {
+	encoder.SetEscapeHTML(escapeHTML)
+
+	if err := encoder.Encode(val); err != nil {
 		return starlark.None, err
 	}
 
-	return starlark.String(string(valBs)), nil
+	res := strings.TrimSuffix(buffer.String(), "\n")
```
prembhaskal, line 79: why is the trimsuffix needed now?

pkg/yttlibrary/json.go
prembhaskal: I think changing `core.BoolArg` is fine for me.

prembhaskal:
@cppforlife @100mik please have a look at this when you get time.

prembhaskal:
@WnP also could you please check the lint failure when you get time. 
meanwhile I will try to get our approvers eyes on the review.

pkg/yttlibrary/json.go
WnP: Otherwise the json encoder always add an extra `\n` at the end of the encoded json.

I use `TrimSuffix` for backward compatibility. (and make tests happy)

WnP:
Sorry for the delay @prembhaskal, I was quite busy during the last days.

Feel free to ping me if there's something missing.

prembhaskal:
LGTM,
@cppforlife please review when you get time.

sethiyash:
LGTM!

sethiyash (approved):

joaopapereira (approved):
LGTM

PR #899, opened 2024-03-25
Export named validations to openApi properties

peschmae:
This pull requests, exposes the attributes for the named `#@schema/validations` to the openApi schema, where a direct mapping is possible.

| Type    | ytt validation      | openapi validation | 
| ------- | ------------------- | ------------------ |
| string  | min_len             | minLength          |
| string  | max_len             | maxLength          |
| integer | min                 | minimum            |
| integer | max                 | maximum            |
| any     | one_of              | enum               |

The PR includes tests for all those validations.

Fixes #897

Issue #898, opened 2024-03-25
Add support for jsonschema export similar to openapi-v3


peschmae:
**Describe the problem/challenge you have**
Most text editors use json schema, to autocomplete and validate YAML files. To make it easier to the users, to write YAML files based on a YTT schema, it would be great to have a way to create the json schema from a YTT file, similar to the openapi schema that is currently implemented.

**Describe the solution you'd like**
Support for json schema as a output option for ` ytt -f schema.yaml --data-values-schema-inspect --output`

**Anything else you would like to add:**
During kubecon 2024, the talk [Developers demand UX for K8s](https://kccnceu2024.sched.com/event/1YeNM/developers-demand-ux-for-k8s-mairin-duffy-red-hat-inc-conor-cowman-atlantic-technological-university?iframe=no) highlighted that YAML validation & auto-completion seems to be one of the main struggles developers face when working with Kubernetes.

Using YTT and it's schemas, could help alleviate that issue.

---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help working on this issue.

peschmae:
After diggin in to the openapi schema generation for #897 I think this shouldn't be too hard to implement, as the property attributes are mostly identical, and a lot of the code could be reused.

From a quick glance at https://swagger.io/docs/specification/data-models/keywords/ the biggest difference is that `type` is a list, and `nullable` isn't an attribute, but needs to be added to the `type` list.

@prembhaskal Let me know if I should draft a pull request for this.

peschmae:
I found some time today, and threw a quick and dirty approach together in a feature branch: https://github.com/peschmae/ytt/tree/feat/jsonschema

It would still need a refactoring, to reduce duplicate code shared with the openschema export, but it already converts into a valid schema that is properly interpreted by vscode.

prembhaskal:
@peschmae I am bit stuck with other stuff, but I can review it over weekend. but you can go ahead and put up a PR if you want some early eyes on it.

@100mik @sethiyash PTAL when you get time.

prembhaskal:
I am unable to check it still, so removed assignee.

renuy:
@cppforlife @100mik Please take a look at this

PR #901, opened 2024-03-28
JSON schema export

peschmae:
As described in #898 I'll make a try to integrate JSON schema export into YTT.

It's based on my previous PR #899 since it's touching the same code, and I'd like to have the validation in the json schema.

I can detangle those PRs if required, but I think #899 will merged first, and I plan to rebase and squash anyways, before converting this to a proper PR.

Right now I'm mostly looking for feedback about the structure of the code, and some hints how I could improve it.
Since openApi schema is a superset of json schema, I already tried composition of the structs, but that doesn't work with the recursive nature of `calculateProperties`.
If there is some composition done, I would also argue, that the `OpenAPIDocument` should be based on `JSONSchemaDocument` (and the `openAPIKeys` type renamed to `jsonSchemaKeys` or similar), since that would reflect the `set<->superset` nature of json schema and openApiSchema .


pkg/schema/type.go
peschmae, line 1: Changes in here all relate to #899

pkg/validations/validate.go
peschmae, line 1: Changes in here all relate to #899

cppforlife:
probably worth rebasing on top of https://github.com/carvel-dev/ytt/pull/904 (already merged).

> quick & dirty json schema, needs a lot of refactoring, to make code DRY again

even though openapi and jsonschema is similar (and more recently openapi started using json schema in some parts) i actually dont have a problem with them being two separate code paths.

peschmae:
Rebased onto develop, squashed the commits, and updated the tests, to use the same ytt schemas as the openapi tests.

Having played around with extracting the components from the openApi spec, into a json schema, I'm no longer sure, if it makes sense to have this as it's own export output type, even though I still like the idea to have validation in common text editors through json schema.

Maybe it makes more sense to keep the code in YTT simple, and just refere to another tool/script that can extract the components from the openApi schema into a json-schema, similat to what kubeconform does to get a json schema from a kubernetes CRD using https://github.com/yannh/kubeconform/blob/master/scripts/openapi2jsonschema.py

Stitching together two tools (ytt | openapi2jsonschema ) to get a json schema doesn't sound too bad.

PR #904, opened 2024-04-13
Refactor openapi validation gen to avoid mutation inside GetValueType()

cppforlife:
Also includes:
- do not include named validation in openapi if when= is present
- set min/maxItems and min/maxProperties when generating openapi for min/maxLen annotations

cc @peschmae @joaopapereira


joaopapereira (approved):
LGTM

PR #934, opened 2024-12-04
Issue 928

ernstvonoelsen:
targets resolution of  https://github.com/carvel-dev/ytt/issues/928.

### DIFF
--- a/pkg/files/file.go
+++ b/pkg/files/file.go
@@ -47,6 +47,12 @@ func NewSortedFilesFromPaths(paths []string, opts SymlinkAllowOpts) ([]*File, er
 
 		relativePath := ""
 		pathPieces := strings.Split(path, "=")
+		if strings.HasPrefix(pathPieces[0], "http://") || strings.HasPrefix(pathPieces[0], "https://") {
+			pathPieces = []string{path}
+		} else if len(pathPieces) > 1 && (strings.HasPrefix(pathPieces[1], "http://") || strings.HasPrefix(pathPieces[1], "https://")) {
+			pathPieces[1] = strings.Join(pathPieces[1:], "=")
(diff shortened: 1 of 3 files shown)


pkg/files/file.go
```diff
@@ -47,6 +47,12 @@ func NewSortedFilesFromPaths(paths []string, opts SymlinkAllowOpts) ([]*File, er
 
 		relativePath := ""
 		pathPieces := strings.Split(path, "=")
+		if strings.HasPrefix(pathPieces[0], "http://") || strings.HasPrefix(pathPieces[0], "https://") {
+			pathPieces = []string{path}
+		} else if len(pathPieces) > 1 && (strings.HasPrefix(pathPieces[1], "http://") || strings.HasPrefix(pathPieces[1], "https://")) {
```
joaopapereira, line 52: What is the scenario where this if statement is triggered?

ernstvonoelsen: @joaopapereira this is a safetynet for the use case that the  file's path is overwritten on the command line; cf. the `case 2` part in L60.

Example: want to cover both scenarios of
- `path = "myOverwrittenPath=https://example.com?hello=world"`
- `path = "https://example.com?hello=world"`

joaopapereira (changes requested):
Sorry for dropping this out, do you mind adding some tests for this? just 1 or 2 basic tests I think it would be enough

ernstvonoelsen:
@joaopapereira done :heavy_check_mark: 

for an e2e test, cf. [my comment in the issue](https://github.com/carvel-dev/ytt/issues/928#issuecomment-2519100311)

joaopapereira (approved):
LGTM

PR #948, opened 2025-03-16
[ytt] v0.51.x bump go version and fix linter issues

devacts:


.golangci.yml
```diff
@@ -12,15 +12,21 @@ linters-settings:
...
+      disabled: true
 issues:
   max-issues-per-linter: 0
   max-same-issues: 0
   exclude-use-default: false
-  new-from-rev: bd77aced0a1cd6d6f3ecd3ed04c1832e1c513ab0
-run:
   exclude-dirs:
-    - pkg/yamlmeta/internal
+    - pkg
```
ccoVeille: While I like your approach to enable all revive checkers, I disagree with this line here. You are disabling almost everything golangci-lint may reports.

So either you consider bumping new-from-rev to a recent commit, either you don't enable all revive linters, either you fix the errors reported

ccoVeille (changes requested):

.golangci.yml
devacts: Hi @ccoVeille, Have created a new PR which bumps the new-from-rev to latest commit.
https://github.com/carvel-dev/ytt/pull/949

PR #954, opened 2025-04-08
Convert data module values once in TemplateLoader

### LINKED ISSUE #953: `NewDataModule` performance when loading large amounts of data values
Hello!

We use YTT for some templating and given the size of our templates and the number of times we generate with different inputs, YTT rendering performance is something we pay attention to. I did some quick profiling, and I believe I’ve identified some big performance wins.

I noticed the CPU use for `NewDataModule` was unusually high (close to 90% of all samples in an internal benchmark), mostly spent converting the `yamlmeta.Document` argument into a `starlark.Value`. Since `NewDataModule` is called for each file evaluation, this adds up quickly when many files are evaluated for a template even though the document doesn't change.

I believe it is safe to do this conversion once within the `TemplateLoader` and then reuse the result for each new data module. A quick ~10 line proof of concept of this change passes repo tests and our internal tests.

With `GOGC=off` I see the following stats on my dev machine when running an internal benchmark:
- Mainline `develop`:
`5.26user 0.11system 0:05.49elapsed 97%CPU (0avgtext+0avgdata 353880maxresident)k`
- Convert in `TemplateLoader`:
`0.55user 0.03system 0:00.60elapsed 97%CPU (0avgtext+0avgdata 247556maxresident)k`

I'll plan to open a PR with the proposed change shortly, but wanted to check for any concerns

emh-jump:
Convert the document data values into a `starlark.Value` once in template loader instead of every time a file is evaluated. Resolves #953 

A rough benchmark of ytt called from the command line (that I'm unable to share, though see the issue for a public reproduction) yields a performance improvement as follows:
```
'develop' (mean ± σ): 5.561s ± 0.300s    [User: 5.437s, System: 0.071s]
this PR   (mean ± σ): 0.631s ± 0.078s    [User: 0.585s, System: 0.038s]
```

### DIFF
--- a/pkg/workspace/template_loader.go
+++ b/pkg/workspace/template_loader.go
@@ -52,8 +52,11 @@ func NewTemplateLoader(values *datavalues.Envelope, libraryValuess []*datavalues
 	}
 
 	return &TemplateLoader{
-		ui:                 ui,
-		values:             core.NewGoValueWithOpts(values.Doc.AsInterface(), core.GoValueOpts{MapIsStruct: true}).AsStarlarkValue(),
+		ui: ui,
+		values: core.NewGoValueWithOpts(
+			values.Doc.AsInterface(),
+			core.GoValueOpts{MapIsStruct: true},
+		).AsStarlarkValue(),
 		libraryValuess:     libraryValuess,
 		librarySchemas:     librarySchemas,
 		opts:               opts,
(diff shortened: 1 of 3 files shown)


emh-jump:
@joaopapereira and/or @praveenrewar I see you both have reviewed some PRs recently, do either of you have time to review this or point me to someone who can?

joaopapereira (changes requested):
I think the PR looks good in general, but do you mind fixing the lint issue?

emh-jump:
Thanks, fixed!

joaopapereira (approved):
LGTM

Issue #911, opened 2024-04-28
Trailing comment affecting `yamlfmt.Printer`


prnvbn:
**What steps did you take:**
I have the following Go program and YAML file. The Go file integrates with the ytt API and reads the ytt yaml file and just prints it put to standard output. I would expect the file to be printed out with no semantic differences. However, that is not the case!

- Go program that reads the file andd prints it to stdout
```go
package main

import (
	"os"

	"carvel.dev/ytt/pkg/yamlfmt"
	"carvel.dev/ytt/pkg/yamlmeta"
)

func main() {
	bs, _:= os.ReadFile("example.yaml")
	docset, _:= yamlmeta.NewDocumentSetFromBytes(bs, yamlmeta.DocSetOpts{})
	printer := yamlfmt.NewPrinter(os.Stdout)
	printer.Print(docset)
}
```

- YAML file: I have provided a very minimal YAML file
```yaml
hello: world
# test
```

**What happened:**
I see the following in my standard out.
```yaml
hello: world

# test
---
```

**What did you expect:**
I expected the following in my stdout
```yaml
hello: world

# test
```

**Anything else you would like to add:**
I am pretty sure the trailing comment is causing this issue.

**Environment:**

- ytt version is v0.49.0. I have pasted my go.mod file below

```
module issue

go 1.22

toolchain go1.22.2

require carvel.dev/ytt v0.49.0
```
---
Vote on this request

This is an invitation to the community to vote on issues, to help us prioritize our backlog. Use the "smiley face" up to the right of this comment to vote.

👍 "I would like to see this addressed as soon as possible"
👎 "There are other more important things to focus on right now"

We are also happy to receive and review Pull Requests if you want to help to work on this issue.

praveenrewar:
@prnvbn Apologies for the late reply.
Could you please share some more context around what you are trying to achieve here? 
By default ytt would remove this comment (`ytt -f example.yaml`) as this is not a template file (doesn't have any ytt comments). You can achieve the same result by passing `WithoutComments` as true in `DocSetOpts` if the comment is not required.

renuy:
@prnvbn Thanks for raising this.
We are adding this to our backlog, however if you want to fix this please raise a PR and we will happy to review it on priority.

joaopapereira:
Released version 0.51.0 with the fix to this issue

PR #962, opened 2025-08-07
Fix `ytt fmt` for files with a single document separator

emh-jump:
After #911, there is a regression for files which have only a single document separator (ie my added test case passed before that PR, failed once it was merged, and is fixed by this change)

cc @cppforlife another one for you to review/merge/release when you return, thank you!


joaopapereira (approved):
LGTM

PR #969, opened 2025-12-11
Add "---" to separate functions on lib files

EllySparky:
# What was done?

A yaml separator was added between functions on the example .lib.yml files.

# Why was this done?

I found that the [playground's functions example](https://github.com/carvel-dev/ytt/blob/develop/examples/playground/basics/example-function/def.yml) explains how this can be done to isolate functions when the format conflicts with the rest of the file.

In the case of the .lib.yml files it's my understanding their only purpose is to centralize functions, it's true that a conflict won't always exist, yet I believe it would be a good practice to always isolate functions on this files, by doing this the format of .lib.yml files remains consistent (without sporadic separator here and there) making the file easier to read and in the case of actual conflicts developers who are used to follow this format should (in theory) not even encounter any issue.

### DIFF
--- a/examples/playground/basics/alt-example-k8s-helm-ish2/helpers.lib.yml
+++ b/examples/playground/basics/alt-example-k8s-helm-ish2/helpers.lib.yml
@@ -6,11 +6,15 @@ name: #@ fullname(vars)
 labels: #@ labels(vars)
 #@ end
 
+---
+
 #@ def labels(vars):
 chart: #@ "{}-{}".format(vars.Chart.Name, vars.Chart.Version).replace("+", "_")
 heritage: #@ vars.Release.Service
 _: #@ template.replace(selector(vars))
 #@ end
+
+---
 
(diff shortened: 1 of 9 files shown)


neil-hickey:
Hey @EllySparky firstly let me note that i am no longer a maintainer, this is just my personal opinion.

The use of the separate documents is to overcome a limitation and is generally an exception to the rule (I've seen libraries with 20+ functions which make use of data values as a way to solve this issue you are facing). Separating every function into it's own yaml document seems like it's trying to solve a problem that imo makes the code library harder to read rather than easier (seems like we disagree here!)

I'd love for someone else to chime in here how common is it to actually use a yaml separator in their library files, and perhaps the "more readable" way to go about this all is to use data values in our functions instead of raw maps like in the examples.

joaopapereira:
I agree with @neil-hickey I do not see what benefit we will get from having them separated in the examples.

PR #990, opened 2026-05-27
fix: Fix incorrect markdown in README.md

moijes12:
- Fixed incorrect markdown 
- Updated links for backlog and Zenhub login to their correct positions


joaopapereira (approved):
I completely forgot we still pointed to zenhub, lol it has been so long...
Thanks for the fix

PR #996, opened 2026-07-29
Avoid computing debug template dump when debug is off

armallen:
## What this PR does / why we need it

`TemplateLoader` calls `l.ui.Debugf("### template\n%s", compiledTemplate.DebugCodeAsString())` after compiling every YAML, text, and Starlark template. Since Go evaluates function arguments eagerly, `DebugCodeAsString()` — which serializes every compiled code line via `fmt.Sprintf` and joins them into one large string — runs on every template evaluation, even though `Debugf` immediately discards the result unless `--debug` is set.

This PR adds an `IsDebug()` method to the `ui.UI` interface and guards the three call sites so the debug dump is only computed when debug output is actually enabled. Benchmarking shows a measurable performance improvement on template-heavy workloads, and it also removes the associated allocation/GC pressure.

## Which issue(s) this PR fixes

None filed; found while benchmarking template evaluation.

## Describe testing done for PR

- Added a unit test for `TTY.IsDebug()` (`pkg/cmd/ui/tty_test.go`)
- `go build ./...` and `go test ./pkg/...` pass
- Verified `--debug` output is unchanged

## Does this PR introduce a user-facing change?

```release-note
Improved template evaluation performance by skipping debug string generation when --debug is not set
```

## Additional Notes for your reviewer

`TTY` is the only implementation of `ui.UI` in the codebase, so extending the interface is contained.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/cmd/ui/tty.go
+++ b/pkg/cmd/ui/tty.go
@@ -35,6 +35,12 @@ func (t TTY) Debugf(str string, args ...interface{}) {
 	}
 }
 
+// IsDebug reports whether debug output is enabled. Callers can use it to
+// skip computing expensive Debugf arguments when debugging is off.
+func (t TTY) IsDebug() bool {
+	return t.debug
+}
+
 func (t TTY) DebugWriter() io.Writer {
 	if t.debug {
 		return os.Stderr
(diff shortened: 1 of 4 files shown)


emh-jump:
I'm a big fan of this FWIW, we have run into this internally as well. I would love to see this PR reopened

joaopapereira:
If @armallen what's to reopen it I can review it in s more timely manner if not @emh-jump feel free to open a new one

PR #997, opened 2026-08-03
Deprecate @ytt:md5 module in favor of @ytt:sha256

G-Gobi:
## Summary
- Deprecates the `@ytt:md5` module's `sum()` builtin in favor of `@ytt:sha256.sum()`, which already exists in this codebase.
- `md5.sum()` behavior and output are unchanged (byte-for-byte), so existing templates keep working exactly as before — this is not a breaking change.
- Adds a one-time deprecation warning (printed to stderr via `ui.Warnf`), mirroring the existing convention used by the `@ytt:math` module for its own caveat warning.
- Fixes a latent bug in `test/filetests/filetests.go`: the test harness constructed `yttlibrary.NewAPI(...)` with a literal `nil` for the `ui.UI` parameter. This went unnoticed because no existing `.tpltest` fixture exercised a `ui`-dependent module (like `@ytt:math`) — but `md5.tpltest` does exercise `@ytt:md5`, so adding the warning immediately surfaced a nil-pointer panic. Fixed by passing a real `ui.NewTTY(false)`.

## Why
MD5 is a cryptographically broken hash algorithm. `sha256.sum()` is a safer, already-available alternative. Since `md5.sum()` is a public Starlark API that existing templates may depend on for non-security purposes (e.g. deterministic identifiers), removing it outright would be a breaking change — so this follows the same soft-deprecation pattern already established for `@ytt:math`: keep the function working, warn once per process, and point users to the recommended replacement.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] Full unit test suite passes (`go test ./...`), including the existing `pkg/yamltemplate/filetests/ytt-library/md5.tpltest` fixture, which now also exercises the new warning code path (previously would have nil-pointer-panicked without the `filetests.go` fix)
```
go test ./pkg/yamltemplate/... -run 'TestYAMLTemplate/filetests/ytt-library/md5.tpltest' -v 2>&1


=== RUN   TestYAMLTemplate
=== RUN   TestYAMLTemplate/filetests/ytt-library/md5.tpltest

Warning: @ytt:md5 module is deprecated because MD5 is a cryptographically weak hash algorithm; use @ytt:sha256 instead.
--- PASS: TestYAMLTemplate (0.01s)
    --- PASS: TestYAMLTemplate/filetests/ytt-library/md5.tpltest (0.00s)
PASS
ok      carvel.dev/ytt/pkg/yamltemplate (cached)
```
- [x] Manually verified via `./ytt -f -` that:
  - stdout only contains the rendered template result (unchanged)
  - stderr contains the one-time deprecation warning

### DIFF
--- a/pkg/yttlibrary/md5.go
+++ b/pkg/yttlibrary/md5.go
@@ -6,26 +6,74 @@ package yttlibrary
 import (
 	"crypto/md5"
 	"fmt"
+	"sync/atomic"
 
+	"carvel.dev/ytt/pkg/cmd/ui"
 	"carvel.dev/ytt/pkg/template/core"
 	"github.com/k14s/starlark-go/starlark"
 	"github.com/k14s/starlark-go/starlarkstruct"
 )
 
-var (
-	MD5API = starlark.StringDict{
+// MD5Module contains the definition of the @ytt:md5 module.
+//
+// Deprecated: MD5 is a cryptographically broken hash algorithm. This
+// module is retained only for backward compatibility with existing
+// templates; use the @ytt:sha256 module instead.
+type MD5Module struct {
+	ui ui.UI
+}
+
+// hasWarnedMD5Deprecated indicates whether the deprecation notice for
+// this module has been displayed. This flag ensures we display that
+// warning only once; more than once and its noise. It's an atomic.Bool
+// since Starlark builtins may be invoked concurrently across goroutines
+// when ytt is embedded as a library.
+var hasWarnedMD5Deprecated atomic.Bool
+
+// NewMD5Module constructs a new instance of MD5Module with the
+// configured UI (to enable displaying a warning).
+func NewMD5Module(uiArg ui.UI) MD5Module {
+	return MD5Module{ui: uiArg}
+}
+
+// AsModule produces the corresponding Starlark module definition
+// suitable for use in running a Starlark program.
+func (b MD5Module) AsModule() starlark.StringDict {
+	sumFunc := core.ErrWrapper(b.warnOnCall(b.sum))
+	return starlark.StringDict{
 		"md5": &starlarkstruct.Module{
 			Name: "md5",
 			Members: starlark.StringDict{
-				"sum": starlark.NewBuiltin("md5.sum", core.ErrWrapper(md5Module{}.Sum)),
+				"sum": starlark.NewBuiltin("md5.sum", sumFunc),
 			},
 		},
 	}
-)
+}
 
-type md5Module struct{}
+// warnOnCall ensures that if the wrapped function is called, the
+// user is warned that md5 is deprecated.
+func (b MD5Module) warnOnCall(wrappedFunc core.StarlarkFunc) core.StarlarkFunc {
+	return func(
+		thread *starlark.Thread,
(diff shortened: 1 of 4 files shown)


test/filetests/filetests.go
```diff
@@ -256,7 +257,7 @@ func (l DefaultTemplateLoader) Load(_ *starlark.Thread, module string) (starlark
 		yttlibrary.NewDataModule(
 			core.NewGoValueWithOpts(l.DataValues.AsInterface(),
 				core.GoValueOpts{MapIsStruct: true},
-			).AsStarlarkValue(), nil), nil, nil)
+			).AsStarlarkValue(), nil), nil, ui.NewTTY(false))
```
joaopapereira, line 260: Not 100% sure here. I assume the tests fail without this. I am concerned that this might make a difference to final users.
Also it would be nice if we could check that the warning is written out. How hard would that be?

joaopapereira (commented):
I agree with a premise of the PR my only concern is with users starting to get this warning.

pkg/yttlibrary/md5.go
```diff
@@ -7,25 +7,71 @@ import (
...
+// module is retained only for backward compatibility with existing
+// templates; use the @ytt:sha256 module instead.
+type MD5Module struct {
+	ui ui.UI
+}
+
+// hasWarnedMD5Deprecated indicates whether the deprecation notice for
+// this module has been displayed. This flag ensures we display that
+// warning only once; more than once and its noise.
+var hasWarnedMD5Deprecated bool
```
aroradaman: atomic bool?

test/filetests/filetests.go
G-Gobi: Good catch.

Fixed it by passing a real `ui.NewTTY(false)`.

On checking the warning is actually written out — added `pkg/yttlibrary/md5_test.go`:

```
=== RUN   TestMD5SumEmitsDeprecationWarningOnce
--- PASS: TestMD5SumEmitsDeprecationWarningOnce (0.00s)
PASS
ok  	carvel.dev/ytt/pkg/yttlibrary	0.186s
```

It captures stderr via `ui.NewCustomWriterTTY`, asserts the warning text appears on the first `md5.sum()` call, and asserts it does *not* repeat on a second call.

pkg/yttlibrary/md5.go
G-Gobi: Done — switched to `atomic.Bool` with `CompareAndSwap(false, true)`

G-Gobi:
Manual testing details:
```
# template.yaml
#@ load("@ytt:md5", "md5")
result1: #@ md5.sum("hello")
result2: #@ md5.sum("hello1")

```

```
 ./ytt -f template.yaml 1>out.yaml 2>err.log
```

```
# out.yaml
result1: 5d41402abc4b2a76b9719d911017c592
result2: 203ad5ffa1d7c650ad681fdff3965cd2

```

```
# err.log
Warning: @ytt:md5 module is deprecated because MD5 is a cryptographically weak hash algorithm; use @ytt:sha256 instead.

```

joaopapereira (approved):
Lgtm

PR #1000, opened 2026-08-05
Skip leading UTF-8 BOMs in YAML, Starlark, and text templates

### LINKED ISSUE #999: Files with UTF-8 BOM produce incorrect output
**What steps did you take:**
While processing a file with an [UTF-8 Byte Order Mark](https://en.wikipedia.org/wiki/Byte_order_mark) the processing produces an incorrect result.

**What happened:**
UTF-8 BOM seems not to be processed. Instead the BOM is treated as data.

**Steps to reproduce**
```bash
mkdir withbom
mkdir withoutbom
data='test: #@ None'
printf "$data\n" >withoutbom/data.yaml
printf "\xEF\xBB\xBF$data\n" >withbom/data.yaml
echo "withoutbom"
ytt -f withoutbom
echo "withbom"
ytt -f withbom
```
produces
```
withoutbom
test: null
withbom
"\uFEFF\x74\x65\x73\x74": null
```
**What did you expect:**
Both input files should render to the same output.

**Environment:**
```bash
% ytt --version
ytt version 0.55.1
% cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.5 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.5 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

Thanks and best regards
Sönke

vsolano9:
Fixes #999

## What

A file saved with a UTF-8 BOM was parsed with the BOM as content. In YAML it became a leading U+FEFF on the first key; in Starlark it caused a compile error; and in a text template it leaked into rendered output.

The YAML path also had a second failure: `ParseBytes` decides whether to prepend a document marker by matching `docStartMarkerCheck` against the raw input. A BOM in front of a leading `---` hid that marker, so the parser prepended a second one and the file did not parse.

## Where the fix lives, and why

The shared `files.TrimUTF8BOM` helper removes only a leading UTF-8 BOM. YAML parsing invokes it before the document-marker check. Starlark and text-template evaluation invoke it before compiling or parsing their template input.

The trim deliberately does not happen in `files.File.Bytes()`: that method also feeds the verbatim passthrough path, where a file force-marked `type=text` may be binary and must retain every byte. In `EvalText`, the trim therefore happens only after the plain/non-template passthrough returns.

`NewDocumentSetFromBytes` also trims so the bytes retained for `AsSourceBytes`—which back source lines shown in template errors—match the bytes parsed. The operation is idempotent. A U+FEFF later in a file remains ordinary content.

## Testing

- parser coverage compares BOM-prefixed YAML with the same input without a BOM across a plain mapping, explicit `---`, leading `#!` comment, and multi-document stream, including the hidden-marker failure
- end-to-end template coverage exercises the reported YAML case and confirms a mid-file BOM survives
- follow-through coverage confirms BOM-prefixed `.star` libraries and text templates behave exactly like their no-BOM equivalents

The new cases were confirmed failing before their respective implementation changes.

Fresh local validation on 18 Aug 2026:

- `go test ./... -count=1` — all packages pass, including `test/e2e` and `test/filetests` (with a locally built versioned `ytt` binary as required by the e2e harness)
- focused BOM tests and the affected package suites pass
- `git diff --check` passes; changed Go files are `gofmt` clean

The branch remains mergeable with current `develop` (`a3e1f7b`).

## AI assistance

This patch was prepared with AI assistance. The resulting diff and test output were reviewed locally before submission.

### DIFF
--- a/pkg/cmd/template/utf8_bom_test.go
+++ b/pkg/cmd/template/utf8_bom_test.go
@@ -64,10 +64,12 @@ func TestUTF8BOMStarlarkLibraryLoadsTheSameAsNoBOM(t *testing.T) {
 	renderWithLibrary := func(t *testing.T, libraryData []byte) string {
 		t.Helper()
 
+		dataYML := "#@ load(\"values.star\", \"value\")\nresult: #@ value\n"
 		filesToProcess := []*files.File{
-			files.MustNewFileFromSource(files.NewBytesSource(
-				"data.yml", []byte("#@ load(\"values.star\", \"value\")\nresult: #@ value\n"))),
-			files.MustNewFileFromSource(files.NewBytesSource("values.star", libraryData)),
+			files.MustNewFileFromSource(
+				files.NewBytesSource("data.yml", []byte(dataYML))),
+			files.MustNewFileFromSource(
+				files.NewBytesSource("values.star", libraryData)),
 		}
 
 		out := cmdtpl.NewOptions().RunWithFiles(
--- a/pkg/cmd/template/utf8_bom_test.go
+++ b/pkg/cmd/template/utf8_bom_test.go
@@ -1,0 +1,76 @@
+// Copyright 2024 The Carvel Authors.
+// SPDX-License-Identifier: Apache-2.0
+
+package template_test
+
+import (
+	"testing"
+
+	cmdtpl "carvel.dev/ytt/pkg/cmd/template"
+	"carvel.dev/ytt/pkg/cmd/ui"
+	"carvel.dev/ytt/pkg/files"
+	"github.com/stretchr/testify/require"
+)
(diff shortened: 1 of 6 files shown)


joaopapereira (approved):
Lgtm

PR #1004, opened 2026-08-20
fix: Make rendering safe to use concurrently

thomastaylor312:
I found out when I started using ytt as a library, rather than a CLI, that race tests were failing. I tracked it back down to ytt itself. There were two locations that were using global variables that weren't protected by some sort of syncronization. I made some small changes, along with some tests to check for it as well.

### DIFF
--- a/pkg/template/compiled_template.go
+++ b/pkg/template/compiled_template.go
@@ -15,8 +15,10 @@ import (
 	"github.com/k14s/starlark-go/syntax"
 )
 
-type EvaluationCtxDialectName string
-type EvaluationCtxDialects map[EvaluationCtxDialectName]EvaluationCtxDialect
+type (
+	EvaluationCtxDialectName string
+	EvaluationCtxDialects    map[EvaluationCtxDialectName]EvaluationCtxDialect
+)
 
 type CompiledTemplate struct {
 	name         string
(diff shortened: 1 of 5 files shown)


pkg/template/compiled_template.go
```diff
@@ -101,8 +110,8 @@ func (e *CompiledTemplate) DebugCodeAsString() string {
 // Eval templates a document by executing the compiled code and instructions from a CompiledTemplate.
 // `instructionBindings` maps the compiled code and instructions to functions defined on a CompiledTemplate.
 func (e *CompiledTemplate) Eval(thread *starlark.Thread, loader CompiledTemplateLoader) (
-	starlark.StringDict, interface{}, error) {
-
+	starlark.StringDict, interface{}, error,
```
thomastaylor312: All of these came from the auto formatter running. If they're bothersome, I can revert them

pkg/template/compiled_template.go
```diff
@@ -28,21 +30,28 @@ type CompiledTemplate struct {
...
-	evalDialects EvaluationCtxDialects) *CompiledTemplate {
-
-	// TODO package globals
+// The Starlark templates used here require these resolver options. They live in
+// package-level variables in starlark-go, so they are set once here at init:
+// otherwise this causes a data race two templates compile concurrently, which
+// happens whenever ytt is embedded as a Go module and more than one render is
+// in flight. The values are process-wide by nature and never change, so this is
+// the best way to set it
+func init() {
```
thomastaylor312, line 42: This was the least important of the two races as it just was re-setting the same value on each call, but it was always the same value. So this is mostly to not fail when running `go test -race`

pkg/template/instructions.go
```diff
@@ -20,13 +21,17 @@ type InstructionSet struct {
...
+//
+// This is to protect concurrent usage when consumed as a Go module. In the
+// worse case scenario this could be a race that ends up with two templates
+// having the same ID.
+var globalInsSetID atomic.Int64
 
 func NewInstructionSet() *InstructionSet {
-	globalInsSetID++
-	uniqueID := globalInsSetID
+	uniqueID := globalInsSetID.Add(1)
```
thomastaylor312, line 34: This race was the one that could cause an actual bug. Not sure how it would manifest in the instruction op (since this is my first time in the ytt codebase), but you could have races of this variable and an `InstructionSet` would end up with ops containing different IDs

